PR #10, opened 2026-07-28
Land the cxxrtl test runner, riscv_test.h, and a working make test

thejefflarson:
Closes JEF-606

## Summary

- `test/cxxrtl.cc` grows from a 27-line VCD dumper into the ADR-0007 runner: `--rom <hex> --ram <hex> --cycles N [--vcd out.vcd]`, images loaded via `debug_items`, halts on a `tohost` magic write, real exit codes (0 pass / 1 fail-with-testnum / 2 timeout / 3 usage-or-setup error).
- `test/asm/riscv_test.h` (new): `RVTEST_RV64U` (no-op, matches all 51→46 files' upstream-derived include), `RVTEST_CODE_BEGIN/_END`, `RVTEST_DATA_BEGIN/_END`, `RVTEST_PASS`, `RVTEST_FAIL`, writing the riscv-tests `tohost` encoding per ADR-0008.
- `test/asm/sections.lds` rewritten into two regions per ADR-0008 (`.text`→`rom` at 0, `.data`/`.rodata`/`.bss`→`ram` at a non-zero base), sized to the largest measured `.text`/`.data` in the suite with headroom.
- `test/testbench.v`: its simulated ROM/RAM arrays grow from 256 words each (1KB — too small to hold even one assembled test's `.text`, which measures up to ~1.5KB) to sizes matching `sections.lds`, and RAM's address decode moves to a non-zero base matching ADR-0008. This is a capacity/addressing fix to test/testbench.v's *own internal* simulated memories — littlecpu's port list and the bus protocol between it and the testbench are untouched.
- Deleted the 13 RV64-only `.S` files per ADR-0002. Added the eight `rv32um` tests (`mul`, `mulh`, `mulhsu`, `mulhu`, `div`, `divu`, `rem`, `remu`) sourced from upstream riscv-tests, adapted to this repo's local macro convention (`RVTEST_RV32U`→`RVTEST_RV64U` for consistency with the other 38 files, which is a no-op either way).
- `test/run_tests.sh` (new): assembles every `test/asm/*.S` with `-march=rv32im_zicsr -mabi=ilp32 -nostdlib -T sections.lds`, `objcopy -O verilog --verilog-data-width=4` for both regions, runs each under `sim`, prints a pass/fail table, and checks the failure list against `test/EXPECTED_FAIL` — exits 0 only on an exact match.
- `test/EXPECTED_FAIL` (new): the sprint-1 baseline. **Right now it lists all 46 tests** — see "A core bug found while landing this" below.
- `Makefile`: `make setup` (brew on macOS, documented `apt` line on Linux), `make sim` (`-O2 -DNDEBUG`, `-Wall -Wextra -Werror`, `-isystem` for the vendored cxxrtl runtime headers so third-party warnings don't fail the build), `make test-units` (wires in JEF-605's three orphaned benches — `exec_tb.v`, `mem_tb.v`, `decoder_tb.v` — via iverilog/vvp, exit code is the pass/fail signal), `make test` (depends on `sim` and `test-units`).
- `CLAUDE.md`: updated the now-stale "`make test` does not exist" bullet, and appended the mechanism behind the discovered core bug to the (still-accurate) regfile bullet.

## A core bug found while landing this — read before reviewing test/EXPECTED_FAIL

While building this against the real core, I found that **every register read of a just-written, non-zero register returns the wrong value, always** — confirmed directly against `rtl/regfile.v`'s stored register contents (ground truth), independent of any test macro or harness code. Root cause, traced to two specific lines:

- `rtl/regfile.v`: `reg_rs1 <= rs1 > 0 ? regs[rs1] : 0;` — a **registered** (1-cycle-latency) read port, not the combinational read CLAUDE.md's invariant 6 documents.
- `rtl/decoder.v`: `out.rs1 <= instr_lui ? immediate : reg_rs1;` (and the equivalent for `out.rs2`, and `out.mem_addr <= immediate + reg_rs1`) — consumes `reg_rs1`/`reg_rs2` in the **same** cycle it asserts a *new* `rs1`/`rs2` request to the regfile, one cycle before that request's answer comes back. So every instruction's operand is actually the regfile's answer to the *previous* instruction's read, not its own.

I verified this with a minimal cxxrtl harness (poke ROM directly, step, read the `regs` memory item) on hand-assembled sequences like `addi x1,x0,5; addi x2,x1,3` (x2 comes out `3`, not `8`) and confirmed spacing/priming the read (more NOPs, an extra same-register read immediately before) does not fix it — x0 reads are unaffected only because the regfile hardwires those to 0 regardless of timing. This matches a bullet CLAUDE.md's "Current state" section already carried before this PR ("operands are one instruction stale — always") — it isn't new, just now empirically reconfirmed against the real suite (all 46 tests, `simple.S` included, currently time out; none reach a `PASS`/`FAIL` write).

**This directly blocks acceptance criterion 6** (`simple.S` exits 0). I considered an `x0`-only workaround for `RVTEST_PASS`/`RVTEST_FAIL` (I confirmed address computation via `x0`-sourced immediates *is* reliable, since `x0` is hardwired regardless of the timing bug) but rejected it: it can't be made ADR-0008-compliant (the *value* half of any store is subject to the same bug, with no analogous `x0` escape hatch, so `tohost`'s value encoding can't be made reliable this way), it would need to be ripped out the moment the real fix lands, and ADR-0008 is Accepted/settled — swapping its protocol for a core-bug workaround isn't a call I should make unilaterally. **Touching `rtl/` is out of this ticket's explicit scope**, so I'm shipping the harness exactly to ADR-0007/ADR-0008 spec, listing all 46 tests (including `simple.S`) in `EXPECTED_FAIL`, and flagging this here.

**DECISION NEEDED:** file a follow-up ticket to fix `rtl/decoder.v`'s operand-consumption timing (hold `out.rs1`/`out.rs2`/`out.mem_addr` for one more cycle so they consume the regfile's answer to *this* instruction's own request, not the previous one). Once that lands, `test/EXPECTED_FAIL` should shrink incrementally — starting with `simple.S`, which needs no other change to go green — not in one shot, since the ALU/mul/div computations underneath still need their own correctness passes.

## Acceptance criteria — real output

1. `make setup` exits 0; `riscv64-elf-gcc --version` succeeds after (`riscv64-elf-gcc (GCC) 16.1.0`, brew-installed).
2. `make sim` builds with `-Wall -Wextra -Werror`, zero warnings — confirmed clean.
3. `./sim --rom <hex> --ram <hex> --cycles 100` with no `tohost` write: `TIMEOUT`, exit **2**.
4. `make test` runs to completion, full 46-row table, no test skipped.
5. `test/EXPECTED_FAIL` committed; `make test` exits 0 (verified from a clean `make clean && make test` rebuild). Mutation-tested: removing an entry makes it exit 1 with a diff; restoring makes it exit 0 again.
6. **Not met** — see above. `simple.S` currently times out like every other test, due to the rtl/ bug found above, which this ticket cannot touch.
7. `riscv_test.h` compiles all 46 `.S` with zero assembler warnings (confirmed — `run_tests.sh` treats any assembler stderr output as a failure state, and none of the 46 produced any). Verified `__riscv_xlen == 32` for `-march=rv32im_zicsr -mabi=ilp32` directly (`riscv64-elf-gcc -dM -E`), and confirmed the RV64-sourced 64-bit literals (`0xffffffff80000000` etc., present in 14 kept files) assemble cleanly through `MASK_XLEN`/`li`, which would hard-error on an unmasked 64-bit immediate.
8. `make test` wall time with `sim` already built: ~2.6s (measured via `time`), well under 60s.
9. `git status --porcelain` clean of `test/rtl.cc`, `sim`, `waves.vcd` after a full `make clean && make test` run (all gitignored; verified).

Also verified `make testbench.vvp` (the iverilog leg, JEF-605's restored elaboration) still builds successfully after the `test/testbench.v` size/addressing change — same pre-existing class of icarus "sorry: constant selects" informational messages as already present from `rtl/decoder.v`/`rtl/accessor.v`, one more instance from the new `rom[imem_addr[12:2]]` index, not a new *kind* of issue.

## Test plan

- `make test-units`: mutation-tested by flipping an expected value in `decoder_tb.v` (confirmed exit 2 with a `MISMATCH`/`FATAL`), then reverted (confirmed exit 0 again).
- `make test`: mutation-tested by removing a line from `test/EXPECTED_FAIL` (confirmed exit 1 with a diff), then restored (confirmed exit 0 again).
- Full `make clean && make sim && make test` from scratch: exit 0, table shows all 46 tests, matches `EXPECTED_FAIL`.
- `./sim` timeout path exercised directly per criterion 3.
- `/soundcheck:pr-review`: no Critical/High findings (local CLI test harness, no network surface, no untrusted input).
- `/simplify`: single small script cleanup applied to `test/run_tests.sh` (direct pass counter instead of a fragile regex-against-padded-output count; removed a duplicate condition).

## Scope notes / risks

- `test/testbench.v`'s ROM/RAM array sizes and RAM base-address decode changed (documented above) — this was necessary for correctness (the old 256-word/1KB arrays couldn't hold even the smallest real test's `.text`), not optional polish. `littlecpu`'s port list and the mem/imem bus protocol are unchanged.
- `make waves` (`waves.vcd: sim; ./sim`) is now further broken than before: `./sim` requires `--rom`/`--ram`/`--cycles`, so a bare `./sim` now exits with a usage error instead of dumping the old hardcoded demo program's waveform. CLAUDE.md already documented `make waves` as "currently broken"; I left it alone (documented the new failure mode in a Makefile comment) rather than take on an unscoped fix.
- **DECISION NEEDED** (see above): the `rtl/decoder.v`/`rtl/regfile.v` operand-forwarding bug blocking criterion 6 needs its own ticket; `rtl/` is out of JEF-606's scope.


thejefflarson:
## Review — integration pass (architect)

I re-ran every acceptance check on this branch from a clean `make clean` rather than taking the
PR description's word for it. Everything reproduces. Merging.

Recorded here because two things in this PR's description should **not** be carried forward into
the follow-up ticket, and one deviation needs to survive as a written decision.

---

### (a) The all-46 `EXPECTED_FAIL` baseline — accepted

Shipping a baseline where every test fails is the right call over either alternative (holding the
harness until the core works, or bending ADR-0008's `tohost` protocol around a core bug). The
harness is correct to spec and the core is not; encoding that split honestly is what a baseline is
for.

The obvious risk with an all-fail baseline is that it **cannot self-validate**: a harness that is
structurally incapable of ever reporting a pass produces exactly the same table. So I made that the
condition of merge and checked it directly.

Verified, on this branch:

- **Runner exit ladder, all four paths.** Seeding `tohost` in the `--ram` image: `1` → `PASS`/exit 0,
  `9` → `FAIL 4`/exit 1 (confirming the `testnum = tohost >> 1` decode), `0` → `TIMEOUT`/exit 2,
  missing `--ram` → usage/exit 3.
- **A real pass driven through the RTL**, not just seeded — assemble → `objcopy` → load → core
  executes → `tohost` write → runner reports `PASS`, exit 0. Details in (b), since the program that
  does it is also the evidence against the proposed fix.
- **The gate is not vacuous.** Deleting `add.S` from `EXPECTED_FAIL` → `run_tests.sh` exits 1 and
  prints `> add.S` in the expected-vs-actual diff. Flipping a `check_hex` expectation in
  `test/decoder_tb.v` → `MISMATCH xori math_arg: got=ffffffff expected=deadbeef`, `$fatal(1)`,
  non-zero exit, `make test-units` fails. Both reverted.
- `make sim` from scratch: **zero** yosys warnings and zero clang warnings under `-Wall -Wextra
  -Werror`. `make test` completes in ~5.5s, exits 0, 46/46 `TIMEOUT` matching the baseline.
  `make test-units`: all three benches pass.
- The riscv-formal clone is at the pin (`c992aa61`) and `formal/pin.mk`'s guard is satisfied.

One incidental confirmation: the timeout traces show the core's wild store landing at
`0x00000000`, which the new non-zero `RAM_BASE` leaves unmapped. Under the old `mem_addr < 1024`
decode that store would have silently hit word 0 — i.e. `tohost` itself. The testbench comment's
justification for a non-zero base is not hypothetical; it is load-bearing today.

**The contract going forward** (written up as ADR-0014): lines come out of `EXPECTED_FAIL` one at a
time as the fix that turns each green lands. Never wholesale, and never by regenerating the file
from a run — that would launder a regression into the baseline.

---

### (b) The proposed decoder hold — rejected, please do not implement it

The description's `DECISION NEEDED` proposes holding `out.rs1`/`out.rs2`/`out.mem_addr` in
`rtl/decoder.v` for one more cycle so they consume the regfile's answer to their own request.
Please don't pick this up in the follow-up. Three reasons, the first empirical:

**1. It does not solve the hazard.** The one-cycle-skew model the description works from is
incomplete. I padded the setup with 16 NOPs — far more than enough for `addi x3,x0,1` and
`lui x5,0x10` to retire — and a subsequent `sw x3, 0(x5)` *still* read `rs1=0, rs2=0`. So this is
not only a read-port latency skew; a correct value sitting in the regfile still does not reach
decode. What actually works is repeating the store:

```
    addi x3, x0, 1
    lui  x5, 0x10
    <16 nops>
    sw   x3, 0(x5)     -> write 0x00000000: 0x00000000   (both operands wrong)
    sw   x3, 0(x5)     -> write 0x00010000: 0x00000000   (address right, data still wrong)
    sw   x3, 0(x5)     -> write 0x00010000: 0x00000001   -> PASS
```

**`rs1` and `rs2` have different skews** — the address port recovers after one repeat, the store-data
port needs two. A uniform one-cycle hold would fix the address and leave the data wrong, converting
a loud all-timeout baseline into a quiet wrong-results one. That is a strictly worse failure mode.

(That program is a diagnostic, not a test, and is deliberately not checked in: it passes *because*
of the bug and would break the moment the bug is fixed.)

**2. It is additive to the design, not an alternative reading of it.** ADR-0004 and CLAUDE.md
invariant 6 already specify the regfile: **combinational-read with write-through bypass**, with a
stall-only interlock in decode. `rtl/regfile.v` today is a registered-read file with no bypass —
it flatly violates that invariant, and *that* is the defect. A hold in the decoder leaves the
violation in place and layers a second timing structure on top of it. The fix is to make
`rtl/regfile.v` what ADR-0004 already says it is.

**3. It is worse on CPI for no benefit.** A blanket hold costs a cycle on every instruction. The
ADR-0004 design costs cycles only on real hazards. This project trades CPI for readability
deliberately, but this trade buys neither.

I'm not resting this on "ADR-0004 says so" — the point is that the ADR-0004 design is also the one
that actually works, and the proposal isn't a cheaper way to get there.

While reading that path I also found `rtl/writeback.v` asserts `wen = (in.rd != 0)` with **no
`in.valid` term**, which violates CLAUDE.md invariant 3 (retire is `valid` reaching writeback, and
`valid` gates `wen`). Bubbles can therefore commit register writes. Separate defect, same ticket
neighbourhood.

---

### (c) `-march=rv32im_zicsr` — correct for now, but it is a deviation and is now recorded

`test/run_tests.sh` assembles with `-march=rv32im_zicsr`. The project's stated target is
**RV32IMC**_Zicsr (ADR-0002, CLAUDE.md), and the C extension is there because code density is a
product constraint on the up5k, not a preference.

This is the right setting *today* — ADR-0003's dual-word combinational fetch window isn't
implemented, so the core cannot fetch a 16-bit instruction, and assembling with `c` would have the
toolchain emit compressed encodings the core cannot execute. The suite would fail for a reason
unrelated to what each test is checking.

The problem is that it was silent. As written, the suite would keep quietly proving RV32IM forever
and nobody would notice the C extension was never exercised. So it is now written down with a
sunset condition in ADR-0014: **flip to `-march=rv32imc_zicsr` when ADR-0003's fetch window lands**,
and expect a batch of `EXPECTED_FAIL` churn at that point rather than treating it as a regression.

---

### Follow-ups for the tracker

1. Make `rtl/regfile.v` combinational-read with write-through bypass per ADR-0004 / invariant 6 —
   the actual root cause. Note the differing `rs1`/`rs2` skews above; a fix that only corrects one
   port will look like progress and produce wrong results.
2. Gate `rtl/writeback.v`'s `wen` on `in.valid` (invariant 3).
3. Flip the suite to `-march=rv32imc_zicsr` when ADR-0003 lands (ADR-0014's sunset condition).
4. `rtl/decoder.v` emits 10 iverilog `sorry: constant selects in always_* processes are not fully
   supported` diagnostics. Pre-existing, untouched by this PR, but "elaboration warnings are errors"
   is a stated rule and this is the second elaboration frontend telling us the sensitivity list is
   not what the source implies.
5. `make waves` is now documented as unmaintained rather than working. Either restore it against the
   new runner CLI or delete the target — a Makefile rule that exists and is known-broken is the
   thing M0 was supposed to clear out.

PR #51, opened 2026-08-01
JEF-648: detect every trap in decode; enter and return from machine mode

thejefflarson:
Closes JEF-648

M3's keystone: every trap this core implements is now **detected and committed in decode**
(CLAUDE.md invariant 2), and `formal/EXPECTED_FAIL` **reaches empty**.

## What changed

* **`rtl/decoder.v`** computes misalignment from the same `$signed(immediate) + $signed(reg_rs1)`
  it already had, decides illegal-instruction (including ADR-0005's two illegal-CSR rules), and
  commits all five causes in the **same non-stalled `else` arm that issues an instruction**. That
  placement is load-bearing and is commented as such: it gives no trap on a stalled cycle, no trap
  from a bubble, no double commit, and a `reg_rs1` the ADR-0004 scoreboard has already made
  hazard-clear (`uses_rs1` is true for every load and store — narrow it and misalignment detection
  silently starts reading a stale register).
* **A trap is a branch.** `pc <= mtvec` is the same override `jal` and the six branches use, on the
  same edge, through the same register. **No flush was added and none is needed** — invariant 1
  intact. ADR-0030's priority encoder is deliberately *not* `parallel_case`, because it is vacuous
  today; the decoder component proof asserts the `$onehot0` disjointness that makes it vacuous.
* **`rtl/csrs.v`** gains a second write port (`trap_entry`/`trap_cause`/`trap_epc`/`mret_entry`)
  and returns `mtvec`/`mepc` to decode. `mepc` goes through the same WARL mask an explicit
  `csrw mepc` does — bit 0 only, because C makes 2-byte targets legal.
* **`mret` gets its own decode** (it was classified illegal — correct before, wrong after) and
  serializes on the existing drain predicate. **`wfi`/`fence`/`fence.i` become NOPs** — not
  cleanup: they were merely *unrecognised*, and unrecognised now means illegal-instruction, so a
  legal `fence` would have faulted (ADR-0034 flagged this as the change that must fix it).
* **ADR-0011's debt discharged**: `rtl/accessor.v` has no `mem_misaligned`, `rtl/littlecpu.v` has
  no `mem_misaligned_trap`. `trap` is redefined per ADR-0028 as a one-cycle "trap entry committed"
  pulse that something finally consumes.
* **ADR-0034's sunset condition FIRED.** `decoder_output`'s `csr_addr`/`is_csr_imm` acquired no
  consumer in this ticket (trap detection, CSR access and trap commit all happen in decode), so
  both are struck.
* **ADR-0029's harness half** landed with the RTL that makes it reachable: `trap_to_zero` →
  cxxrtl exit 5 → `TRAP-TO-ZERO` in `run_tests.sh`, plus `$fatal` under iverilog.

### 1. `formal/EXPECTED_FAIL` empty — 82 checks, 82 pass, 0 fail

From-scratch `rm -rf formal/checks && make -C formal check`:

```
82 checks: 82 pass, 0 fail
Failure list matches EXPECTED_FAIL exactly.
Generated check set matches EXPECTED_CHECKS exactly (82 checks).
```

Full per-check tally, verbatim:

```
causal_ch0               PASS   insn_c_addi4spn_ch0      PASS   insn_c_xor_ch0           PASS   insn_sll_ch0             PASS
causal_mem_ch0           PASS   insn_c_and_ch0           PASS   insn_div_ch0             PASS   insn_slli_ch0            PASS
csrw_mcycle_ch0          PASS   insn_c_andi_ch0          PASS   insn_divu_ch0            PASS   insn_slt_ch0             PASS
csrw_minstret_ch0        PASS   insn_c_beqz_ch0          PASS   insn_jal_ch0             PASS   insn_slti_ch0            PASS
csrw_mscratch_ch0        PASS   insn_c_bnez_ch0          PASS   insn_jalr_ch0            PASS   insn_sltiu_ch0           PASS
hang                     PASS   insn_c_j_ch0             PASS   insn_lb_ch0              PASS   insn_sltu_ch0            PASS
ill_ch0                  PASS   insn_c_jal_ch0           PASS   insn_lbu_ch0             PASS   insn_sra_ch0             PASS
insn_add_ch0             PASS   insn_c_jalr_ch0          PASS   insn_lh_ch0              PASS   insn_srai_ch0            PASS
insn_addi_ch0            PASS   insn_c_jr_ch0            PASS   insn_lhu_ch0             PASS   insn_srl_ch0             PASS
insn_and_ch0             PASS   insn_c_li_ch0            PASS   insn_lui_ch0             PASS   insn_srli_ch0            PASS
insn_andi_ch0            PASS   insn_c_lui_ch0           PASS   insn_lw_ch0              PASS   insn_sub_ch0             PASS
insn_auipc_ch0           PASS   insn_c_lw_ch0            PASS   insn_mul_ch0             PASS   insn_sw_ch0              PASS
insn_beq_ch0             PASS   insn_c_lwsp_ch0          PASS   insn_mulh_ch0            PASS   insn_xor_ch0             PASS
insn_bge_ch0             PASS   insn_c_mv_ch0            PASS   insn_mulhsu_ch0          PASS   insn_xori_ch0            PASS
insn_bgeu_ch0            PASS   insn_c_or_ch0            PASS   insn_mulhu_ch0           PASS   liveness_ch0             PASS
insn_blt_ch0             PASS   insn_c_slli_ch0          PASS   insn_or_ch0              PASS   pc_bwd_ch0               PASS
insn_bltu_ch0            PASS   insn_c_srai_ch0          PASS   insn_ori_ch0             PASS   pc_fwd_ch0               PASS
insn_bne_ch0             PASS   insn_c_srli_ch0          PASS   insn_rem_ch0             PASS   reg_ch0                  PASS
insn_c_add_ch0           PASS   insn_c_sub_ch0           PASS   insn_remu_ch0            PASS   unique_ch0               PASS
insn_c_addi_ch0          PASS   insn_c_sw_ch0            PASS   insn_sb_ch0              PASS
insn_c_addi16sp_ch0      PASS   insn_c_swsp_ch0          PASS   insn_sh_ch0              PASS
```

All ten baseline entries removed. Nine were misalignment (`insn_lh/lhu/lw/sh/sw` +
`insn_c_lw/c_lwsp/c_sw/c_swsp`); the tenth, `ill_ch0`, went green for a *different* reason —
`rvfi_ill_check.sv` is the formal statement of the illegal-instruction trap plus ADR-0028's
convention. **`insn_lb`, `insn_lbu`, `insn_sb` — the byte accesses whose address cannot be
misaligned — never moved**, which is the attribution behind those entries holding rather than
being asserted.

### 2. `dmemcheck` stays PASS

```
SBY [dmemcheck] DONE (PASS, rc=0)
```

Per ADR-0028 this is the guard that a trapping store never reaches the bus: `formal/dmemcheck.sv`
builds its environment shadow from the **real** `mem_wstrb`/`mem_wdata` while `rvfi_dmem_check`
builds its shadow purely from `rvfi_mem_*`, so a store that is architecturally suppressed but
still strobes the bus desynchronises the two and a later load fails at the RVFI level. This is
work `rvfi_insn_check` declines to do — it drops every `mem_*` assertion under `spec_trap`.

**Measured caveat, reported rather than assumed** (ADR-0023): I could not make `dmemcheck` go red
with either mutation I tried at its configured depth. Dropping `is_sb/is_sh/is_sw` from the trap
suppression (a trapping store that executes *and* is reported) leaves both shadows agreeing, so
no desync exists — that one is caught by criterion 6 instead (`trap.S FAIL 8`). Forcing
`out.rvfi_mem_wmask <= 4'b0` in `rtl/accessor.v` while the bus still strobes — ADR-0028's literal
scenario — also returned `DONE (PASS, rc=0)` (18 min run). So the *reasoning* stands but the
*catch* is unconfirmed at this depth; worth an ADR-0028 amendment.

### 3. `reg_ch0` stays PASS

`reg_ch0 PASS` (see tally). Guards that a trapping instruction does not write `rd`, because
`rvfi_rs1_rdata`/`rs2_rdata` are captured from the *real* regfile in decode, so a corrupted `rd`
surfaces as a later instruction disagreeing with the shadow model.

### 4. `pc_fwd_ch0` / `pc_bwd_ch0` stay PASS

`pc_fwd_ch0 PASS`, `pc_bwd_ch0 PASS` (see tally). They assert `pc_wdata(N) == pc_rdata(N+1)` with
no trap special case, so the redirect is both honestly reported and actually taken. They are
satisfied by *any* target — that the target is `mtvec` is asserted in `rtl/decoder.v`'s own
component proof, which is the one place it can be.

### 5. `test/asm/trap.S` in both sim legs

cxxrtl (`make test`): `trap.S PASS`, part of `52/52 passed`.

iverilog: `IVERILOG-RESULT: PASS`, with `test/monitor.sim.v` instantiated and no error code. The
whole suite runs under `vvp` now: **52 pass, 0 fail** (see the iverilog-leg note below).

`trap.S` takes all five causes plus both illegal-CSR rules and asserts `mcause` is
4 / 6 / 11 / 3 / 2 / 2 / 2 / 4 and that `mepc` equals the faulting PC, `mret`ing back each time.
Two cases are new:

* a **write to a read-only CSR** (`csrw mvendorid, t0`, addr[11:10] == 2'b11) → cause 2, the rule
  `formal/checks.cfg` records as architecturally unobservable until trap entry existed;
* a **compressed faulting instruction** — `c.lw` placed at `pc % 4 == 2` via `.align 2` + `c.nop`,
  with a `c.nop` behind it so the handler's unconditional `mepc+4` resume lands on an instruction
  boundary. Test 16 asserts `mis_clw & 3 == 2`, so the case cannot quietly degrade into a
  4-aligned one that proves nothing about the WARL `mepc[0]` mask.

Verified layout: `mis_clw` at `0x146`, `c.lw` 2 bytes, `c.nop` at `0x148`, resume at `0x14a`.

### 6. Misaligned store leaves the target RAM word unchanged

`trap.S` TEST_CASE 8 reads `tdat` back after a misaligned `sw` and requires `0x0ff0`. Proven live
by mutation — dropping `out.is_sb/is_sh/is_sw <= 0` from the trap suppression:

```
trap.S           FAIL 8
51/52 passed
```

Reverted; `trap.S PASS`, `52/52`. This is checked on real memory and is independent of (2).

### 7. Monitor error 101, and the mutation that fires it

Clean: `trap.S PASS` under cxxrtl (error 101 would surface as `MONITOR-ERROR 101`), and
`IVERILOG-RESULT: PASS` with no monitor banner under iverilog.

**The mutation: report `rvfi_trap` one retire late.** In `rtl/writeback.v`, replace
`assign rvfi_trap = in.rvfi.trap;` with

```systemverilog
logic rvfi_trap_late;
always_ff @(posedge clk) if (!reset && in.valid) rvfi_trap_late <= in.rvfi.trap;
assign rvfi_trap = rvfi_trap_late;
```

cxxrtl leg:

```
trap.S           MONITOR-ERROR 101
51/52 passed
Failure list does NOT match test/EXPECTED_FAIL:
> trap.S MONITOR-ERROR 101
```

Error 101 is "mismatch in trap" — the one comparison
`checks/rvfi_insn_check.sv` keeps live under `spec_trap` and that
`test/sanitize_monitor.py`'s rule 3 must therefore leave outside its gate. `monitor_tb` vector 7
covers the gate-disabled case; this is the live-core case. Reverted.

*(The iverilog leg reported PASS on this mutated build. The `.S` programs there terminate via
`tohost` and the mutated monitor's banner did not gate that run; the cxxrtl leg turns errcode into
exit status and is the one that catches it. Reported as measured.)*

### 8. ADR-0011's debt discharged

`rtl/accessor.v` has no `mem_misaligned` output and no detection logic; `rtl/littlecpu.v` has no
`mem_misaligned_trap`. Both diffs are in the change.

### 9. `make test` / `test-units` / `lint` / elaboration

```
52/52 passed
Failure list matches test/EXPECTED_FAIL exactly (name and status).
```

`test/EXPECTED_FAIL` is **empty** (`trap.S MONITOR-ERROR 101` removed — the third and last of the
three M3 tests seeded ahead of their RTL). `make test-units` green, six benches including the
extended `decoder_tb` (all five causes, both illegal-CSR rules, `fence`/`fence.i`/`wfi`/`mret`
legality, trap-vs-mret redirect, `minstret`/CSR suppression on a trapping issue) and `csr_tb`
(trap entry, MIE→MPIE→MIE round trip, `mepc` WARL on a 2-aligned faulting PC).

`make lint`: `svlint: clean in both passes`. yosys strict elaboration (CI's exact script): **0
promoted warnings**. iverilog: **0 warnings**, and the `sorry:` count is **20 before and 20
after** — held there by driving `rvfi_trap` as a continuous assign rather than adding a line to
`rtl/writeback.v`'s `always_comb` (ADR-0034's documented pattern; the `always_comb` spelling
measured 21). `make monitor-check` and `make -C formal genchecks-check` both clean.

### 10. Sail co-simulation

`test/sail/memory-map.json` now sets the **global** `memory.misaligned.exceptions.load_store` to
`{"Some": "AlignmentException"}`, confirmed against `--print-config-schema` (enum is
`AccessFault` | `AlignmentException` | `None`). `AlignmentException` is the value matching
ADR-0005's causes 4 and 6; `AccessFault` would be 5 and 7, a choice this core did not make.

```
make cosim-run PROG=trap.S
sail changes       : 214  (spin-loop reached: True, last pc=0x00000218)
core changes       : 214   (tohost: PASS 729)
AGREE trap.S: 214 architectural register-file changes, identical in order and value.
```

Proven load-bearing — with the key reverted to the inherited `{"None": null}`:

```
sail changes       : 15
core changes       : 214
DISAGREE trap.S
DIVERGENCE at architectural change #6
  sail instruction #8  pc=0x0000001c  lw x11, 0x0(x10)    mis_lw+0
  sail : x11=0xf000000f
  core : x5=0x0001021c
```

Suite-wide: 50/52 AGREE. The two that do not — `csr.S` and `minstret.S` — **already disagreed
before this change** (verified by stashing): Sail reports `misa = 0x4034112f` against this core's
`0x40001104`, an ISA-configuration mismatch in the Sail config, unrelated to traps. Co-simulation
is opt-in and on no gate (ADR-0032); left as-is rather than tuning the oracle inside this ticket.

## The iverilog leg was dead, and this change found it

`rtl/decoder.v`'s hazard scoreboard called `function automatic live_producer(r)` from two
continuous assigns. **iverilog builds such an assign's sensitivity list from the call's
*arguments***, so a function body that also reads module-level signals (`out`, `executor_out`,
`accessor_pending_*`) never re-evaluated when those changed. That is *under*-sensitivity — the one
direction CLAUDE.md's `sorry:` exception says is a real bug — and iverilog emits no diagnostic for
it at all.

Measured on **`main`, before this change**: `hazard_rs1` latched high at the first RAW hazard of
every program and never fell, so `make waves`' own baked-in loop executed two instructions and
then span — **0 memory writes in 200 cycles**, one `ifetch` line total. No `.S` program could
reach `tohost`. yosys evaluates the function correctly, so cxxrtl and the entire formal ladder
were unaffected, which is exactly why it went unnoticed: the leg the verification table calls the
microscope was reporting nothing, and nothing it reported was wrong.

The two assigns are now written out (same expression, same behaviour under yosys — ladder still
82/82, all component proofs still pass). `make waves` executes real instructions, and all 52 `.S`
programs run to `PASS` under `vvp` with the per-retire monitor live.

**Caveat, and a suggested follow-up:** `test/testbench.v` has a baked-in program and no image
loader, so the 52/52 iverilog figure was measured with a scratch harness (`$readmemh` of the
objcopy images + a `tohost` watcher), not something a reviewer can re-run from the repo. A tracked
iverilog runner for `test/asm/*.S` is worth its own ticket now that the leg works.

## Read this before quoting the empty baseline

**An empty `formal/EXPECTED_FAIL` is necessary for M2, not sufficient, and M2 is not reached.**
Every check is `mode bmc` — PASS means "no counterexample within the configured depth". Everything
runs under `RISCV_FORMAL_ALTOPS` (ADR-0010), so green `insn_mul`/`insn_div`/`insn_rem` say nothing
about the real multiplier or divider. `formal/equiv.sh` still does not converge (ADR-0020), so
ADR-0006's non-perturbation guarantee is still argued rather than proven — and now covers one more
`ifdef RISCV_FORMAL` field, `rvfi_shadow.trap`, which is write-only with respect to the core.
`formal/complete` still fails (pre-existing, ADR-0034; on no gate). The nightly still cannot go
red (ADR-0022). And riscv-formal ships **no spec model at all** for `ecall`/`ebreak`/`mret`/
`csrr*` at the pin, so the behaviour this ticket adds is checked against assertions this repo
wrote — `trap.S`, `csr_tb`, `decoder_tb`, and the decoder component proof — not against an oracle.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -33,8 +33,9 @@ stage to stage riding the existing valid-bit protocol; `rtl/accessor.v` adds the
 addr/rmask/wmask/rdata/wdata capture (the one stage that actually knows those values, including the
 one-cycle load-response latch); `rtl/writeback.v` drives `rvfi_valid`/`rvfi_order`/etc. from the
 retiring `accessor_output`, exactly where invariant 3 below says retire happens. `trap`/`halt`/
-`intr`/`mode`/`ixl`/the CSR fields are hardwired constants in `rtl/littlecpu.v` (no traps or CSRs
-exist yet — M3). `test/monitor.v` stays pristine and tracked; a gitignored, build-time-derived
+`intr`/`mode`/`ixl`/the CSR fields were hardwired constants in `rtl/littlecpu.v` at that commit,
+because no traps and no CSRs existed yet; only `halt`/`intr`/`mode`/`ixl` still are (the CSR fields
+came alive with `rtl/csrs.v`, `rvfi_trap` with trap entry — both below). `test/monitor.v` stays pristine and tracked; a gitignored, build-time-derived
 `test/monitor.sim.v` (`test/sanitize_monitor.py`) carries the build-time repairs it needs —
 originally two, stripping the `$time`-in-`$display` yosys can't elaborate and ADR-0019's DIV/REM
 signedness fix, now three with the `!spec_trap` gate below — and **both** legs read it, so they
@@ -112,10 +113,11 @@ unenforced drift there is exactly what produced the five-year-stale fork ADR-003
 the clone, so any residual diff is drift by construction and is printed. It runs in CI in the
 `monitor-freshness` job, which has already paid for the from-scratch clone at the pin.
 
-`test/EXPECTED_FAIL` is **no longer empty**: `csr.S`, `minstret.S` and `trap.S` were seeded there
+`test/EXPECTED_FAIL` was seeded with three M3 tests — `csr.S`, `minstret.S` and `trap.S` — written
 ahead of the RTL that pays them off, which is the direction ADR-0014's burn-down contract was
-designed for. Two of the three are paid off (below); `trap.S` remains, so `make test` is
-**51 pass / 1 expected-fail** and still exits 0 only on set equality. That equality is now over
+designed for. All three are paid off now (the third by the trap change below), so it is **empty
+again** and `make test` is **52 pass / 0 expected-fail**, exiting 0 only on set equality. That
+equality is over
 **name-and-status pairs** (ADR-0035) — the baselined entry is `trap.S      MONITOR-ERROR 101`, so a
 baselined test that starts failing some *other* way (a broken assembly, a `TIMEOUT`, an unstartable
 runner) is a red gate rather than a match. The same change made every build step's exit status
@@ -123,7 +125,8 @@ checked and `mktemp` fatal: an unchecked `objcopy` that emitted an **empty** RAM
 `tohost` read zero and every data-independent test still report `PASS`.
 `test/run_tests.sh` also grew a `MONITOR-ERROR` label for runner exit 4, which used to fall into
 `RUNNER-ERROR` and read as "the sim would not start" when it actually means the per-retire oracle
-disagreed with the core mid-run. **Nothing here checks M3 semantics against an oracle**:
+disagreed with the core mid-run, and a `TRAP-TO-ZERO` label for exit 5 (ADR-0029, below).
+**Nothing here checks M3 semantics against an oracle**:
 riscv-formal ships no spec model for `csrr*`/`ecall`/`ebreak`/`mret` at the pinned SHA, so
 `spec_valid` is 0 for every one of them and the monitor's whole semantic block is skipped. The
 trap and CSR tests assert on values their own handler recorded. Error 133 ("expected intr after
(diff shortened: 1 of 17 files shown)


thejefflarson:
## Integration ruling on the automated security review

The `review` job (non-required) reported **1 Critical, 4 Medium, 4 Low**. Merged
anyway. Recording why, since the Critical is the kind of thing that should not be
waved through silently.

**The Critical is out of scope and not a regression.** It flags `test/cxxrtl.cc:16`
(`#include "rtl.cc"`) as a supply-chain primitive: PR-authored RTL becomes
yosys-generated C++ that CI compiles and executes. Verified against `main`: that
line, `SIM=$1` in `run_tests.sh`, and both `strtoul` call sites are **unchanged**
by this PR. They are pre-existing structure, not something this change introduced.

`docs/THREAT_MODEL.md` — added two commits before this one specifically to scope
these reviews — is explicit: the blast radius of total compromise here is "one
laptop and one CI runner — no customer data, no credentials worth having, no
production anything", and denial of service is named under **Out of scope**, which
disposes of attack chains C2 and C4. The proposed remedy (sandbox yosys output,
run the design out-of-process over a bounded IPC protocol) would also contradict
ADR-0007, which makes cxxrtl the primary simulator precisely by compiling the
design in.

**The two `EXPECTED_FAIL` Mediums are real and in-scope** — they are threat-model
category 1, "a gate that can report success without having checked anything." A PR
can widen a baseline and launder a regression past the gate. `1961234` (merged
just before this) is the mitigation for the formal half: the ladder now runs on
the PR gate, so a baseline edit is verified by the same PR that makes it, rather
than contradicted by the nightly hours later with the change already on `main`.
That is also the argument for adding `formal` to the required set, ratified in
ADR-0037 §5 and left for a human, since it is a repository-settings change.

**The `CLAUDE.md:48` Low was correct** and is fixed in ADR-0037: the file described
the nightly as `make -C formal check || true` long after that was fixed. The real
remaining defect was different and worse — the graded comparison was piped into
`tee` under `bash -e {0}`, which is errexit without pipefail, so the step's exit
status was `tee`'s and always 0.

Merged with normal `--squash`. No `--admin`, no branch-protection changes.

PR #83, opened 2026-08-02
JEF-692: run the mutation campaign and publish the measured catch matrix

thejefflarson:
Draft — the campaign is running; the matrix, the probe catalog and the findings land as measurements complete.

**What this is.** The deliverable of `docs/ideas/audit-the-oracle-stack.md`: ~29 mutants, one per
defect class, each run against all five oracle surfaces (`make test`, `make test-units`,
`make cosim-suite`, the three component proofs, and the 85-check riscv-formal ladder), producing a
matrix of **measured** verdicts rather than argued ones.

**No `rtl/` file changes.** Every mutant runs in a throwaway copy of the worktree outside git; the
tree is verified byte-identical to a pristine snapshot after each one.

Closes JEF-692

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -990,6 +990,11 @@ longer quietly widen.
   rejected the shifter merge at 19 cells *saved* on legibility grounds, and accepting a hygiene
   change at 37 cells *spent* would cut against that ruling. Read this before proposing the
   narrowing again — it is measured and declined, not overlooked.
+- **What each oracle surface actually catches is measured, not argued**:
+  [ADR-0054](docs/adr/0054-the-measured-catch-matrix-and-the-probe-catalog.md) — a 29-mutant catch
+  matrix over all five surfaces, and the **probe catalog** (one line, one command, one expected red)
+  for checking an oracle is still alive. Four defect classes are caught by nothing; three `rtl/`
+  files are reachable by no oracle at all. Read it before claiming a surface covers anything.
 - Decisions: [`docs/adr/`](docs/adr/) — **fifty-two ADRs, fifty-one of them accepted**, plus a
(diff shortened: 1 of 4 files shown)


thejefflarson:
Closing unmerged. The mutation campaign is not worth its interpretation cost for this repo.

Every MISSED cell requires adjudication between three very different causes — a real oracle gap, a mutation with no observable effect, or a harness defect — and in this run the second and third dominated:

- `H1` (x0 write suppression) was an equivalent mutant. It changes nothing observable, so every oracle was correct to miss it. It was never a finding.
- The entire ladder column was invalid: scratch workdirs symlinked `formal/riscv-formal`, `genchecks-local.py` uses `os.path.abspath` which does not resolve symlinks, and `@basedir@/../../rtl/*.v` therefore resolved into the pristine checkout. Every generated .sby read unmutated RTL. Verified independently: mutating `lh` sign extension gives `bad state property 9 reachable at bound k = 15 SATISFIABLE` on a real clone, and PASS through the symlink.

The surviving 116 simulation-and-proof cells largely restate what the design already implies — the executor proof catches what it asserts, co-simulation catches an architectural write nothing reports, the .S suite catches RVFI payload blanking.

By contrast the repo's concrete checks say what they mean without adjudication, and the two real RTL defects found this sprint (c.ebreak's trap cause, the counter carry boundary) came from reading the specification against the RTL, not from mutation.

One actionable item is carried forward separately: test/exec_tb.v runs 10,000 vectors per op across mul/div/shift and never exercises plain ADD.

Not a comment on the agent's work, which was careful and self-correcting throughout — it caught and reported four defects in its own harness. The technique is the wrong fit.

PR #91, opened 2026-08-02
Use the runner image's OSS CAD Suite when it matches the pin

thejefflarson:
Seven of the eight CI jobs restored the OSS CAD Suite from the GitHub cache service on every run — **607,515,233 bytes each, ~4.25 GB per run** of a file that never changes.

**The cache was hitting.** This was never a miss problem; the cost is bandwidth contention between jobs that start together. Measured across three runs, twenty setups: **129s mean, 12s to 222s**. `formal` starts first and pulls at 46 MB/s; `test` and `elaborate` sit at `0.0 MBs/sec` and take two to four minutes. About 15 minutes of runner time per run.

## Change

The runner image now installs the suite and records the SHA-256 it got (thejefflarson/runner#36). This action uses `/opt/oss-cad-suite` **only when that marker equals the pin**, and otherwise falls through to the existing cache-and-download path untouched.

**The pin stays authoritative and stays in this repo.** The image tracks latest; this repo pins because the formal ladder has to be reproducible. They agree only while the pin is current — and a mismatch costs a slow run, never a wrong toolchain. The log says which path it took, so a stale pin is visible rather than silent.

That preserves ADR-0013's posture: the SHA-256 is still what decides correctness, still lives in `ci.yml`'s `env:` block, and is still verified before anything is extracted. The image is a cache, not an authority.

## Also

The PATH step now checks `yosys` is executable, so a broken setup fails here rather than in whichever job first reaches for it.

## Sequencing

This is safe to merge before the image change: with no marker present, every job takes the existing path and nothing moves. The speedup arrives when the image does.

### DIFF
--- a/.github/actions/setup-oss-cad-suite/action.yml
+++ b/.github/actions/setup-oss-cad-suite/action.yml
@@ -1,47 +1,52 @@
 name: Set up OSS CAD Suite
 description: >-
-  Put the pinned OSS CAD Suite (yosys, iverilog, sby, nextpnr-ice40, icetime,
-  the solvers) on PATH. Uses the copy in the runner image when its recorded
-  SHA-256 matches the pin, otherwise restores from cache, otherwise downloads
-  and checksum-verifies. The pin stays in ci.yml's env: block and is passed in,
-  so this action has no pin of its own to drift.
-inputs:
-  tag:
-    description: OSS CAD Suite release tag (e.g. 2026-06-29)
-    required: true
-  file:
-    description: Release asset filename (e.g. oss-cad-suite-linux-x64-20260629.tgz)
-    required: true
-  sha256:
-    description: Expected SHA-256 of the release asset
-    required: true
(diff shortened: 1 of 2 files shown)


thejefflarson:
**Holding this until the runner image carries the suite.** Measured on this PR's own run: `components` failed at its 5-minute timeout with the setup step alone taking **3m14s**.

Cause is this change. Removing the pin moved the cache key from the fixed SHA to whatever is latest, so every job cold-downloads instead of hitting the warm entry. Until thejefflarson/runner#36 lands and the image rebuilds, this is strictly slower than what it replaces, not faster.

Correct order: runner#36 → image rebuild → this. No change needed here; it just merges second.

thejefflarson:
Superseded by a clean re-apply on current main: #97 rewrote this workflow's comments underneath the branch and the rebase was not worth untangling for a 37-line deletion.

PR #130, opened 2026-08-07
Replace the ADR citations in test/, soc/ and .github/ with the mechanism

thejefflarson:
Closes JEF-758.

Scope this run: `test/`, `soc/`, `.github/` and `Makefile`. **No `rtl/` or `formal/` file is touched** — those are a follow-up run, deliberately left to the agent measuring the decode head.

## What changed

Every ADR number in a **comment** across the scope is replaced with the mechanism in the comment's own words, or deleted where the code already says it. A citation reads as the explanation while carrying none: a reader who has not memorised the decision records learns nothing, and one who has still goes and looks.

Runtime strings are left alone — `$display`, `@echo`, `fprintf`, `sys.exit` messages, `probe` labels and one CI step name. Those are output, not commentary. 24 survive; all are listed at the bottom.

`.github/workflows/ci.yml`, `test/exec_tb.v`, `rtl/executor.v`, `rtl/csrs.v`, `formal/checks.cfg`, `formal/complete.sv` and `formal/Makefile` were checked with grep rather than taken on trust: every survivor in them is already a runtime string, so the earlier cleanup pass's claim holds and none needed re-editing.

## Five stale or false claims, found by reading each comment against the record it cited

1. **`test/EXPECTED_FAIL` and `test/COSIM_EXPECTED_FAIL`** each carried a paragraph justifying a `csrset.S` baseline entry **that no longer exists**. Both baselines are empty and `rtl/csrs.v` implements `mstatush` and `mconfigptr`. Worse, the reasoning they gave — that the CSR set is stated *exact*, so widening it is an ISA-scope decision — is the premise that was explicitly struck when those two landed. Both blocks deleted; `EXPECTED_FAIL`'s own header already forbids repeating history.

2. **`test/asm/csrset.S`** opened with `***THIS TEST IS BASELINED RED.***`, named its two (now absent) baseline lines, and closed with a burn-down for work already completed. The header now says what the file checks. The floor-not-a-closed-list warning is kept as a tripwire so the two arms are not removed again to shrink `rtl/csrs.v`.

3. **`test/asm/minstret.S`** said no trap can be taken "until trap entry lands, which is deliberately the next M3 step and not this one". M3 is reached and `test/asm/trap.S` takes eight traps.

4. **`.github/dependabot.yml`** hand-copied the required status checks as four names — `elaborate, test, components, monitor-freshness`. The API returns **seven**. This is precisely the drift its sibling `dependabot-automerge.yml` warns about in its own header ("It was, and it said four while the API returned six"). The list is removed rather than corrected; read it from the API.

5. **Three counts had rotted the way a count in a tripwire does.** `test/OBSERVED_FLOOR`'s "an exact ratchet over 52 numbers", `test/cosim.cc`'s "every cycle of all 52 programs", and `test/check_suite_shape.sh`'s "a suite that shrank from 52 to 12". The suite is 59. Each is restated without a number.

One further citation was simply misdirected: `soc/fit_report.py` cited ADR-0042 three times for the `FIT_MAX_LC` ratchet. That ADR is about the synchronous regfile read; it set the budget once, at a value since superseded.

## Comments and strings only — proven two ways

Both proofs cover all 39 files in scope.

**1. `-U0` diff.** No non-comment line changed. Every changed line is a `#`/`//` comment, a Python docstring body, or one Verilog trailing comment (`monitor_tb.v`'s `RAM` localparam) whose code prefix is byte-identical.

**2. Code-skeleton hashes.** Each file is compared against `HEAD` after stripping comments and blank lines, **with leading whitespace preserved** so re-indentation cannot hide a change. Python files go through the AST with docstrings dropped, so a docstring edit is invisible but any statement change is not.

```
39 files compared; 0 with a changed code skeleton: none
```

This second proof is not paperwork — an earlier bulk edit in this repo silently dropped an `else if` and an `out.rvfi <= in.rvfi;` from `rtl/decoder.v`, and only the hash check caught it. It also caught a bug in my own stripper (Verilog's `32'h...` tick read as a string delimiter), which is why the check is worth running rather than assuming.

## Checks

| Gate | Result |
|---|---|
| `make test` | **59/59**, both baselines empty, `Failure list matches test/EXPECTED_FAIL exactly` |
| `make test-units` | clean |
| `make lint` | `svlint: clean in both passes` |
| `make probe-gates` | `137 graded comparisons, every failure path executed` |
| `make monitor-check` | clean |
| `make fit` | **3880** of 4100 cells |
| `make soc-timing` | **12.74 MHz** against the 12.00 MHz floor |

`make fit` and `make soc-timing` are **unchanged**: I re-ran both on a stashed tree and got byte-identical numbers (3880 cells, 12.74 MHz), so the measurements are confirmed rather than merely still-passing.

The riscv-formal checks were **not** run locally — the diff touches zero `rtl/` and zero `formal/` files, and `test/monitor.v` and `test/sanitize_monitor.py` are untouched, so the generated checks read identical inputs. The `formal` job on this PR is the confirmation.

## Surviving runtime strings (24)

Criterion 1 permits these — they are code, not commentary.

- `test/probe_gates.sh` ×3 — probe labels (336, 631, 838)
- `test/testbench.v` ×1 — `$display` (320)
- `test/cxxrtl.cc` ×2 — `fprintf` (291, 466)
- `test/cosim.cc` ×1 — `fprintf` (283)
- `test/run_tests.sh` ×1 — `echo` (269)
- `test/sanitize_monitor.py` ×1 — stderr write (192)
- `test/exec_tb.v` ×3 — directed-vector labels (119–121), out of scope this run
- `soc/fit_report.py` ×2 — `print` / `sys.exit` (74, 83)
- `soc/rom_banks.py` ×1 — `sys.exit` (73)
- `Makefile` ×2 — `@echo` (536, 539)
- `.github/workflows/ci.yml` ×6 — five `echo`, one step name; out of scope this run

## Notes

- No ADR is needed for this; it applies a rule CLAUDE.md already states.
- `docs/adr/` and `CLAUDE.md` are out of scope and untouched. `test/monitor.v` and `test/monitor.sim.v` are generated and untouched.
- **Follow-up, not done here:** the words "ladder" and "drains" survive in ~15 comments across `test/` and `.github/`. I left them rather than widen a citation diff into a vocabulary diff, and only avoided them in prose I wrote.

### DIFF
--- a/.github/actions/setup-oss-cad-suite/action.yml
+++ b/.github/actions/setup-oss-cad-suite/action.yml
@@ -24,8 +24,8 @@ runs:
         fi
 
     # Latest, not a pin. A toolchain bump that changes the hardware shows up in
-    # `make fit` and `make soc-timing`, both of which ratchet: ADR-0052 measured
-    # 21 cells between two yosys builds on identical RTL. formal/pin.mk is a
+    # `make fit` and `make soc-timing`, both of which ratchet: 21 cells were
+    # measured between two yosys builds on identical RTL. formal/pin.mk is a
     # different question and stays pinned -- test/monitor.v is generated from it
     # and tracked, so bumping that one means regenerating the monitor.
     - name: Resolve the latest OSS CAD Suite release
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,9 @@
 # .github/workflows/.
 #
 # These PRs auto-merge once CI is green -- see .github/workflows/
-# dependabot-automerge.yml and ADR-0018. That is safe because `main` carries
-# required status checks (elaborate, test, components, monitor-freshness), so
-# `gh pr merge --auto` has a real gate to wait for. ADR-0016 records why the
-# same setup was unsafe before those checks existed.
+# dependabot-automerge.yml. That is safe only because `main` carries required
+# status checks, which is what gives `gh pr merge --auto` a real gate to wait
+# for. The set is not listed here; read it from the API, as that file says.
 #
 # The posture is deliberately "stay current" rather than "merge late by hand":
 # unpatched bugs in the code that runs our CI are a certain, cumulative cost,
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -1,17 +1,14 @@
 name: Dependabot Auto-merge
 
-# Re-armed 2026-07-29 (supersedes ADR-0016's deletion; see ADR-0018).
-#
-# This workflow is only safe because `main` now has required status checks
-# with strict: true. That is
-# what `gh pr merge --auto` waits for. Without them `--auto` merges instantly
-# and this workflow is a rubber stamp — which is exactly why ADR-0016 deleted
-# it while criterion 8 was still outstanding.
+# This workflow is only safe because `main` has required status checks with
+# strict: true. That is what `gh pr merge --auto` waits for. Without them
+# `--auto` merges instantly and this workflow is a rubber stamp — which is why
+# it was deleted once, back when `main` had no required checks to wait on.
 #
 # The set is deliberately NOT enumerated here. It was, and it said four while
 # the API returned six -- this file is not the register of what is required,
 # branch protection is, and a hand-copied list of it drifts silently while
-# looking authoritative (ADR-0050). What this workflow actually depends on is
+# looking authoritative. What this workflow actually depends on is
 # that SOME required set exists with strict: true, which is what the command
 # below reads.
 #
@@ -33,9 +30,9 @@ jobs:
     steps:
       - name: Enable auto-merge
         # --squash, not --merge: main requires linear history, so a merge
-        # commit is rejected outright. The pre-ADR-0016 version of this file
-        # used --merge and would have failed on every run once linear history
-        # was enforced.
+        # commit is rejected outright. An earlier version of this file used
+        # --merge and would have failed on every run once linear history was
+        # enforced.
         run: gh pr merge --auto --squash "$PR_URL"
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
--- a/.github/workflows/riscv-formal-pin-bump.yml
+++ b/.github/workflows/riscv-formal-pin-bump.yml
@@ -2,7 +2,7 @@ name: riscv-formal pin bump
 
 # Notices when upstream YosysHQ/riscv-formal's `main` has moved past
 # formal/pin.mk's SHA, pushes a branch bumping it with test/monitor.v
-# regenerated against the new pin (ADR-0013), and opens an issue asking a human
+# regenerated against the new pin, and opens an issue asking a human
 # to open the pull request from that branch. The existing gates --
 # monitor-freshness, formal/check-genchecks.py,
 # formal/check-complete-exclusions.py, the ladder itself -- grade the bump;
@@ -10,8 +10,8 @@ name: riscv-formal pin bump
 #
 # It opens an issue rather than a pull request because a PR opened by this
 # workflow would get no CI and could never merge; formal/propose-pin-bump.sh has
-# the mechanism. No auto-merge either: ADR-0018's dependabot-automerge.yml is
-# scoped to `dependabot[bot]` and does not apply here.
+# the mechanism. No auto-merge either: dependabot-automerge.yml is scoped to
+# `dependabot[bot]` and does not apply here.
 on:
   schedule:
     - cron: "17 6 * * 1" # weekly, Monday
--- a/soc/cell_census.py
+++ b/soc/cell_census.py
@@ -3,14 +3,13 @@
 
 Replaces the Makefile's `soc_expect_cells` define so `test/probe_gates.sh`
 can drive `make soc-timing`'s SPRAM/EBR census against a fixture log without a
-real yosys run (ADR-0053) -- the same reason `soc/timing_split.py` carries
+real yosys run -- the same reason `soc/timing_split.py` carries
 `make soc-timing`'s frequency ratchet instead of a second parser in the
 recipe.
 
 Matched by cell NAME, not by presence: yosys logs a cell's library definition
 in the same "Number of cells" table whether or not it actually instantiates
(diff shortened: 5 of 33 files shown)


thejefflarson:
## Integration review — ruling on the three deletions, and one correction to the record

**The three prose deletions are ratified**, on stronger grounds than the PR argues.

The PR says the two baseline paragraphs and `csrset.S`'s burn-down justified "a `csrset.S`
baseline entry that no longer exists." That understates it. **The entry never existed.**

`git log --diff-filter=A -- test/asm/csrset.S` and `git log -S` over both baselines agree on
one commit, `15a31f2`. That single commit added `test/asm/csrset.S` — carrying the header
`***THIS TEST IS BASELINED RED.***` and naming its two baseline lines — *and* added
`MSTATUSH = 12'h310` / `MCONFIGPTR = 12'hF15` to `rtl/csrs.v` in the same diff. Checking every
revision of `test/EXPECTED_FAIL`, no non-comment line has ever matched `csrset`, and
`git show 15a31f2:test/EXPECTED_FAIL` has zero entries.

So the test passed from birth. Three files spent their whole life asserting that a baseline
contained a red entry it did not contain, in files whose entire contract is set equality in
both directions. That prose was never true, so there is nothing to "carry forward in the
rewrite's voice" — a rewrite would have had to invent the history it was preserving. Deleting
is the only correct disposition.

**What survives is still usable as a decision procedure**, which was the thing to check:

* `EXPECTED_FAIL` keeps the two-directional contract, the FORMAT block with the exact status
  labels, "justify any line added back in the PR that adds it," and the OBJCOPY review flag.
  The oracle-defect paragraph was *rewritten* rather than cut, and the rewrite carries more
  than the citation did — it now says why (`sanitize_monitor.py` is re-applied at every
  regeneration, so baselining the one program that tripped it leaves the oracle wrong for
  every other program).
* `COSIM_EXPECTED_FAIL` keeps the three ordered questions, the "only when none of the three
  fit is it a baseline entry, and then it needs a paragraph here, not a line" disposition, and
  both worked closure examples (`csr.S`, `minstret.S`). The deleted block was an *application*
  of that procedure to a case that was closed before it was written; the procedure itself is
  untouched.
* `csrset.S` keeps both spec citations with section numbers, the floor-not-a-closed-list rule,
  and the "do not remove either arm to shrink `rtl/csrs.v`" tripwire — which is exactly the
  one exception `CLAUDE.md` allows a comment to carry history for.

## Proofs re-run independently

I did not accept the engineer's grader. I wrote a second one — deliberately different in
construction, `'` never treated as a delimiter in Verilog — and **tested its red direction
before trusting its green one**, per this repo's own rule that a grader which cannot fail is
not a grader. Twelve mutations of real code lines across all eleven file types in the diff
(`.v` localparam, `.S` TEST_CASE operand, `.py` import, `.cc` constexpr, a real baseline entry
appended to `EXPECTED_FAIL`, `.json` knob, `.lds` region length, `.yml` cron and step name,
`.pcf` pin, `.sh` variable): **12 of 12 detected.**

Against that grader, both against the original base and again after the rebase:
`33 files compared; 0 with a changed code skeleton`. The `-U0` scan agrees — every line it
flags is a Python docstring body (covered by the AST comparison, which drops docstrings) or
`monitor_tb.v`'s single trailing comment, whose code prefix
`  localparam [31:0] RAM   = 32'h0001_0000;` is byte-identical. The `32'h` handling the
engineer says it had to fix is correct in my independent implementation too.

Claims verified rather than accepted: the required set is live-confirmed as **seven**
(`elaborate, test, components, monitor-freshness, lint, formal, soc-timing`, `strict: true`),
so the removed hand-copied "four" was genuinely wrong; nothing machine-reads that list (it was
a `#` comment, and no file references `dependabot.yml`). `trap.S` takes eight traps, so
`minstret.S`'s "the next M3 step" was genuinely stale. The suite is 59 by both
`test/OBSERVED_FLOOR` and `ls test/asm/*.S`, so all three "52" counts had rotted. `ADR-0042`
is the synchronous regfile read, so the `FIT_MAX_LC` citations in `soc/fit_report.py` were
genuinely misdirected. Both baselines hold zero entries on main and on this branch.

## Two things the PR body overstates, neither blocking

1. **"Both proofs cover all 39 files in scope"** and `39 files compared`. The diff touches
   **33** files. 39 is presumably in-scope-including-unchanged; I can confirm 33/33, not 39.
2. **"24 runtime strings survive."** My grep over the rebased tree finds **23**, and the PR's
   own enumeration also sums to 23 (3+1+2+1+1+1+3+2+1+2+6). A count that disagrees with its
   own list, in a PR whose finding #5 is three rotted counts. Repo artifacts are unaffected —
   this is a PR-body number only.

## Rebase

Rebased `618ec7f` onto `0b1728a` and force-pushed as `90e97b4` (`--force-with-lease`), since
`strict: true` requires a fresh run. No conflict. `#129` touched `test/cosim.py`, which this PR
also edits; the skeleton check re-run against the *new* base confirms that file's code is
identical to main's with only comments differing, so `#129`'s changes survived the rebase
intact. `#129` also added `test/tool_cache_test.sh` — it carries no ADR citations, so the
sweep is still complete over the scope.

No ADR recorded: this applies a rule `CLAUDE.md` already states, and the deletion ruling
disposes of prose that was never true rather than reversing a decision.

thejefflarson:
## Integration re-verification — no change requested, auto-merge left armed

Re-checked independently at integration rather than trusting the prior approval or the engineer's self-report (the swarm can share a worktree, so scope is confirmed per-PR from the diff):

- **Scope is correct.** 33 files across `test/`, `soc/` and `.github/` only. **No `rtl/`, no `formal/`.**
- **Comments and strings only, verified by reading the diff.** Every changed line is comment prose or a comment continuation. The two that look like code at a glance are not: `test/sail/rv32imc_zicsr.json` is `//` header prose in a JSONC config, and `test/regfile_tb.v` is two comment bodies.
- No overlap with #131 or #132, which touch `CLAUDE.md` and `docs/` only. This PR is independent of both and can merge in any order relative to them.

**Nothing changed here and nothing is asked for.** Squash auto-merge stays armed; three of the seven required checks (`components`, `formal`, `test`) were green before the outage, and `elaborate`, `lint`, `monitor-freshness` and `soc-timing` are queued behind the empty runner pool. It should land on its own the moment a runner registers, with no intervention.

PR #131, opened 2026-08-07
Measure what shortening the decode head is worth, and decline all of it (JEF-764)

thejefflarson:
Closes JEF-764.

ADR-0058 named the decode head (`imem.in_range → instr → {rs1/rs2, immediate, hazard}`) as the shared lever on every near-critical path, `CLAUDE.md` carried it as a rule, and ADR-0074 deferred the operand-fetch cycle to "a change that buys back headroom in the decode head". Nobody had measured it. **Now it is measured, and all of it is declined.**

Nothing in `rtl/` or `formal/` changes. What ships is ADR-0076, the tables, and the dead-end list.

## The table

Homebrew Yosys 0.68+post (`c12172fb`), nextpnr-0.10-108-g68c1acd8, `icetime` from the same install; one tree against `2d79878`; `soc/timing_sweep.sh`, `SOC_PROG=add.S`. **Periods, so a positive delta is slower.** Baseline `make fit` reproduces ADR-0074's 3880 cells exactly on a different Yosys.

### What the decode head is worth

| variant | seeds | ns, sorted | median | vs baseline | `fit` LC |
|---|---|---|---|---|---|
| **baseline `2d79878`** | 8 | 77.38 · 78.34 · 78.51 · 78.55 · 78.61 · 78.80 · 79.20 · 80.07 | 78.58 | — | 3880 |
| A — the width mask deleted | 4 | 78.91 · 79.29 · 80.26 · 80.60 | 79.78 | **+1.5%** | 3955 |
| B — RVC `rs1`/`rs2` selection deleted | 4 | 75.99 · 76.00 · 79.54 · 81.91 | 77.77 | −1.0% | 4003 |
| C — scoreboard cut three slots → one | 4 | 76.67 · 77.00 · 77.16 · 77.75 | 77.08 | −1.9% | 3914 |
| D — 18-way `immediate` mux → one arm | 4 | 77.06 · 77.38 · 77.82 · 79.01 | 77.60 | −1.2% | 3859 |
| **A+B+C+D — the whole head at its ceiling** | 8 | 74.19 · 74.79 · 75.38 · 75.87 · 76.08 · 76.12 · 76.76 · 79.85 | 75.98 | **−3.3%** | 3805 |

The last row is `{rs1/rs2, immediate, hazard}` and the width mask, all gone at once, on a design that cannot execute anything: **3.3% — inside the 3.6% edit-churn band** — and 75 cells. One LUT level off a path that is 23 of them; the critical path does not even move.

### Two implementable spellings, both slower

| candidate | seeds | ns, sorted | median | vs baseline | `fit` LC |
|---|---|---|---|---|---|
| **R** — width mask hoisted out of the register path | 8 | 78.45 · 79.38 · 79.50 · 79.77 · 80.53 · 80.83 · 80.96 · 81.59 | 80.15 | **+2.0%** | 3982 |
| **M** — the two duplicate `rd_field` `rs1` arms merged | 8 | 78.68 · 78.85 · 79.54 · 79.65 · 79.70 · 80.18 · 81.73 · 82.59 | 79.68 | **+1.4%** | 3905 |

R is the honest implementation of ticket candidate 1 — every compressed arm reads bits the mask never touches, so `rs1 = uncompressed ? in.instr[19:15] : <compressed select>` is the same function with the mask off the register path. It costs 2.0% and 102 cells: the mask was not a level in front of the mux, it was folded *into* the mux's default arm. M is semantics-preserving and reads better, and its worst placement is 82.59 ns = **12.11 MHz, 0.9% of margin** over the board clock.

### What every other term of `next_pc` is worth — the finding underneath

| variant | seeds | ns, sorted | median | vs baseline | `fit` LC |
|---|---|---|---|---|---|
| T — `trap_pending` off `next_pc` | 4 | 73.99 · 74.00 · 75.20 · 75.41 | 74.60 | **−5.1%** | 3858 |
| T2 — T without `instr_illegal` only | 4 | 77.06 · 79.36 · 79.50 · 82.03 | 79.43 | +1.1% | 3864 |
| T3 — T without the misalignment terms only | 4 | 78.91 · 79.01 · 80.45 · 81.02 | 79.73 | +1.5% | 3879 |
| BR — the branch comparators off `next_pc` | 4 | 74.77 · 78.03 · 79.09 · 80.72 | 78.56 | 0.0% | 3880 |
| E — `stall` off `next_pc` | 4 | 74.71 · 77.09 · 77.74 · 79.09 | 77.42 | −1.5% | 3928 |
| **all of them, plus A+B+C+D** | 4 | 60.83 · 61.49 · 62.22 · 62.81 | 61.86 | **−21.3%** | 3622 |

**Every input to `next_pc` is worth nothing on its own; together they are worth 21%.** The largest single term is `trap_pending` at 5.1% — and *neither half of it reproduces any part of that*: dropping `instr_illegal` measures +1.1%, dropping the misalignment terms +1.5%, both slower than baseline. The paths are stacked so close that removing the first uncovers the second. That is a plateau, not a peak with a lever on it.

Row E also reproduces ADR-0058's second ceiling at the current design: the critical path leaves the fetch loop and lands on `→ riscv.csrs.minstret` at 77.42 ns, so a fetch-loop-only change is still capped at ~1.5% (it was 1.7%).

## Recommendation

**None of these is worth it.** Nothing clears the bar — a change has to improve one goal with the other three still holding, measured, and the best available is a 3.3% ceiling inside a 3.6% churn band. Five candidates join the dead-end list *at their ceilings*, so re-running one is answered without building it.

## What contradicts the premises the ticket carried

- **The decode head is not the shared lever.** ADR-0058's recommendation is withdrawn and `CLAUDE.md`'s sentence is replaced with the ceiling. This is the one edit outside `docs/` — it corrects a rule this PR measured false.
- **ADR-0074's reopening condition cannot be met this way.** It defers the operand-fetch cycle to headroom in the decode head; there is at most 3.3% there against the 3.6% that variant's median cost. The idea is now blocked on something structural.
- **Ticket candidate 1 is a regression in both directions** — 1.5% slower with the mask deleted outright, 2.0% slower with it correctly hoisted.
- **The 21% names its own price:** it means the pc stops depending on this cycle's decode, which is state a later cycle must un-commit. That is the no-wrong-path-state commitment, and this ADR does not touch it — the ticket ruled that a separate question, and the measurement does not by itself justify it.

## Testing

- `make test` — **59/59, `test/EXPECTED_FAIL` matching exactly (name and status)**. Run to say the tree is unchanged, not to grade a change: `git status rtl/ formal/` is empty and the diff is Markdown only.
- `make fit` — 3880 cells, unchanged, ratchet OK.
- `make soc-timing` / `soc/timing_sweep.sh` — 62 placements across 13 variants; the tables above.
- `make cycles` was not re-run and did not need to be: no candidate changes which cycles issue (the ceilings do not execute; R and M are the same function as what is checked in).
- `make cosim-suite` not run — no RTL changed, so there is nothing for it to have a verdict on.

No `[depth]` line moves: no stall reason, stage length or scoreboard width changes, so F and G are untouched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,8 +175,11 @@ and times.
 - **Read logic levels apart**: a LUT level costs ~3.3 ns (delay plus interconnect), a carry hop
   ~0.34 ns and no interconnect. A change that trades a carry hop for a LUT level gets shallower by
   icetime's count and slower in nanoseconds. The SoC is routing-dominated; wide flat muxes route
-  worse than chains. The shared lever on every near-critical path is the decode head
-  (`imem.in_range → instr → {rs1/rs2, immediate, hazard}`).
+  worse than chains. **There is no single lever.** The decode head
+  (`imem.in_range → instr → {rs1/rs2, immediate, hazard}`) was named as one and measures 3.3% deleted
+  whole — inside the churn band — while every other input to `next_pc` measures nothing on its own
+  and all of them together are worth 21% (ADR-0076). Bring a ceiling for the whole cone or expect a
+  null.
 
 Baselines and grading:
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -176,10 +176,13 @@ and times.
   ~0.34 ns and no interconnect. A change that trades a carry hop for a LUT level gets shallower by
   icetime's count and slower in nanoseconds. The SoC is routing-dominated; wide flat muxes route
   worse than chains. **There is no single lever.** The decode head
-  (`imem.in_range → instr → {rs1/rs2, immediate, hazard}`) was named as one and measures 3.3% deleted
-  whole — inside the churn band — while every other input to `next_pc` measures nothing on its own
-  and all of them together are worth 21% (ADR-0076). Bring a ceiling for the whole cone or expect a
-  null.
+  (`imem.in_range → instr → {rs1/rs2, immediate, hazard}`) was named as one and measures 3.3%
+  deleted whole, inside the churn band. The period is in the fetch loop instead: no single input to
+  `next_pc` is worth more than 5%, all of them deleted together are worth 21%, and collecting that
+  means the pc stops depending on this cycle's decode — which is the no-wrong-path-state commitment
+  (ADR-0076). **Measure the whole set.** A ceiling over one term bounds only that term, and a median
+  inside the churn band is a null in both directions — neither a candidate declined nor one
+  admitted.
 
 Baselines and grading:
 
--- a/docs/adr/0058-the-fetch-loop-is-not-two-levels-too-deep.md
+++ b/docs/adr/0058-the-fetch-loop-is-not-two-levels-too-deep.md
@@ -5,7 +5,9 @@ proposed against
 [ADR-0054](0054-the-memory-system-and-the-first-real-timing-number.md)'s 11.30 MHz measurement and
 [ADR-0057](0057-what-writable-text-costs-in-ladder-depth-and-in-nanoseconds.md)'s finding that
 writable text costs a further 8.5%. Both were built and measured; both are slower than what is
-checked in. Corrects how ADR-0054's "41 logic levels" should be read.*
+checked in. Corrects how ADR-0054's "41 logic levels" should be read. Its closing recommendation —
+target the decode head next — is withdrawn by
+[ADR-0076](0076-the-decode-head-is-a-plateau-not-a-lever.md), which measured it.*
 
 ## What was proposed, and what it rested on
 
--- a/docs/adr/0074-the-operand-fetch-cycle-is-removable-and-costs-more-than-it-saves.md
+++ b/docs/adr/0074-the-operand-fetch-cycle-is-removable-and-costs-more-than-it-saves.md
@@ -5,7 +5,9 @@
 operand-fetch cycle. Uses [ADR-0070](0070-the-suites-cycles-are-charged-to-a-named-stall-reason.md)'s
 per-reason cycle accounting as the before/after instrument. Graded against
 [ADR-0066](0066-twelve-megahertz-is-a-requirement.md)'s 12 MHz requirement, which is what declines
-it.*
+it. Its reopening condition is closed by
+[ADR-0076](0076-the-decode-head-is-a-plateau-not-a-lever.md): the decode head does not hold the
+headroom this ADR defers to.*
 
 ## Context
 
--- a/docs/adr/0076-the-decode-head-is-a-plateau-not-a-lever.md
+++ b/docs/adr/0076-the-decode-head-is-a-plateau-not-a-lever.md
@@ -7,6 +7,14 @@ sentence it put in `CLAUDE.md`, which
 when it deferred the operand-fetch cycle to "a change that buys back headroom in the decode head".
 Graded against [ADR-0066](0066-twelve-megahertz-is-a-requirement.md)'s 12 MHz requirement.*
 
+**Amended at integration, before merge.** The combined row below was written as though its 21%
+needed the decode head deleted along with the `next_pc` terms. It does not:
+[ADR-0078](0078-a-one-deep-kill-is-cheap-and-buys-a-clock-the-board-cannot-use.md) reached the same
+ceiling with the entire decode head left standing, and measured an implemented fetch-loop change at
+four times what this ADR's row E predicted was possible. The attribution of the 21%, the cap this
+ADR put on fetch-loop work, and the rows whose deltas sit inside the churn band are corrected here
+rather than in a third document. The measurements are unchanged; what they are read to mean is not.
+
 ## What was proposed, and what it rested on
 
 Two of this design's bottlenecks were measured within a week of each other and both were attributed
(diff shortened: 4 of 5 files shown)


thejefflarson:
## Integration review — the one this PR never got, plus a ruling on the 21%

Reviewed on the merits, independently re-derived from the tables rather than from the summary. Every median, percentage and MHz figure in both tables recomputes exactly from the placement lists as written, and the `fit` deltas do too. The method is sound and the two eight-seed rows are the right places to have spent the extra placements. **Not merged — no runner is registered, so all seven required checks queue. This is state-of-the-branch work, not a merge decision.**

I pushed one commit, `9da21e5`. Rationale below; revert any part of it you disagree with.

### 1. The 21% is the fetch loop, not the fetch loop plus the decode head — #132 is right

ADR-0078's `K∞E` deletes every redirect term **and** `stall` from the fetch address with the **entire decode head standing**, and measures **61.27 ns** over four placements. This ADR's combined row deletes those terms **plus** A+B+C+D and measures **61.86 ns** over four. That is **0.95% apart with fully overlapping distributions** (60.77–62.38 against 60.83–62.81) — inside every band this instrument has, on both the placement spread and the churn band.

So the combined row's decode-head half bought nothing measurable, and `K∞E` is the stronger experiment because it deletes *more* of `next_pc` (it also takes out the `jal`/`jalr`/`mret` terms this table never had a row for) while touching *less* of the design. The mechanism is consistent with the rest of the record: the decode head is in the fetch loop only via `stall → hazard → rs1/rs2`, which is exactly why #132's implemented `K` — which leaves `stall` in — collects a third of what `K∞E` does.

Corrected here rather than in a follow-up ADR, since neither has merged. The measurements are untouched; the attribution and the framing are not.

### 2. Row E's cap was the more damaging error, and it is withdrawn

Worse than the attribution, and it wasn't in the brief. This ADR concluded from row E:

> A change that shortens only the fetch loop is still capped at about 1.5%.

Row E is a **single-term** ceiling — it removes `stall` from `next_pc` and leaves every redirect term in the loop. By this ADR's own thesis a single term bounds nothing, so the 77.42 ns path it uncovers is not a floor for the set. #132 deleted the redirect terms instead: **−6.8% at the ceiling, and −4.5% for a change that was actually built**, over eight placements, clearing the churn band.

That sentence would have talked a reviewer out of the one fetch-loop change anyone has ever measured as real. It is withdrawn explicitly, and the Method section now says a ceiling bounds only the terms it deletes.

### 3. The `CLAUDE.md` edit: **ratified**, with the sentence corrected

- **Scope.** Out of the ticket's `rtl/ formal/ docs/`, flagged rather than hidden — the right call, and I'm ratifying the edit. `CLAUDE.md` calls itself the rulebook and ground truth and says a rule that stops serving the four goals gets deleted. Leaving a rule the ADR next to it measures false is a worse defect than the scope overrun, and the file's own precedence note is that where documents disagree the ADR wins — a rulebook that loses to its own footnotes is not ground truth.
- **Design commitments header and preamble: confirmed untouched.** The diff is a single hunk in *Measurements and ratchets*, in the "Read logic levels apart" bullet. Nothing in the commitments section, its header, or its preamble is modified.
- **The replacement sentence was not accurate, and I rewrote it.** It said *"every other input to `next_pc` measures nothing on its own"* — the table directly below it reports `trap_pending` at **−5.1%**, the one row in it outside the churn band. It also carried "bring a ceiling for the whole cone or expect a null", which is the same claim row E is being corrected for. It now states the 5% bound, names the fetch loop as where the period is, says what collecting it costs, and adds the symmetry rule.

### 4. "Measured slower" on noise: yes, restate — this was the right thing to catch

Every "slower" row here is inside the 3.6% edit-churn band, against a baseline whose own eight placements span 3.5%:

| row | delta | seeds | `fit` Δ | was | now |
|---|---|---|---|---|---|
| A — width mask deleted | +1.5% | 4 | **+75** | "1.5% slower … no version can do better" | null on period; **declined on +75 cells against a ±50 floor** |
| R — mask hoisted | +2.0% | 8 | **+102** | "2.0% slower and 102 cells bigger" | no measurable period; **declined on +102 cells** |
| M — duplicate `rd_field` arms | +1.4% | 8 | +25 | "1.4% slower at the median" | null on both instruments; **declined on its worst placement, 0.9% of margin** |
| T2 | +1.1% | 4 | +(−16) | "*slower* than the baseline" | **null**, and off the dead-end list |
| T3 | +1.5% | 4 | −1 | "*slower* than the baseline" | **null**, and off the dead-end list |

The band was being applied in one direction only: rows B (−1.0%), C (−1.9%) and D (−1.2%) were correctly hedged as noise, while rows of the same magnitude the other way were quoted as results. That asymmetry systematically favours declining, which is exactly the bias a dead-end list must not encode. #132's engineer reached the same read independently ("singles indicative, combined ceilings solid") and that is the right split.

Each declined row keeps its decline — **the conclusions all survive**, because each rests on something outside the noise: A and R on area, M on the worst-placement margin, which is the same ground ADR-0074 used at 0.83% and is graded on the worst placement by design, not the median. T2 and T3 rest on nothing outside the noise, so they come off the list entirely; they remain what they always were, evidence that neither half of `trap_pending` accounts for any part of its 5.1%.

### 5. ADR-0074's reopening condition: **claim verified**

Checked against the file rather than the summary. ADR-0074's consequence reads *"Reopen it after a change that buys back headroom in the decode head"*, and its finding section names the same path verbatim. It conditions on decode-head headroom directly, not on something adjacent — so this ADR's 3.3% ceiling against that variant's 3.6% cost does close it. The header note added to 0074 is accurate.

One addition, post-reconciliation: the headroom does exist, it is just in the fetch loop, and #132 declined the change that would have supplied it. So the operand-fetch cycle is now blocked behind a commitment amendment rather than behind a tuning edit. Recorded in the consequence.

### Kept deliberately

- **Filename and title unchanged.** "Plateau, not a lever" still reads correctly for the decode head; renaming would break the links #132, 0058, 0074 and the README all now carry, for a wording nuance the body resolves.
- **The amendment is marked as an amendment** rather than folded in silently, so #132's "had one term too many in the row that showed it" stays coherent against this file.
- The `ADR-0078` cross-references live in the ADR, not in `CLAUDE.md` — a dangling citation in a historical record is normal; one in the live rulebook is a defect. Merge order below closes the window anyway.

### State

Approved on the merits, **held on infrastructure only**. Merge **before #132** — both add a row at the same point in `docs/adr/README.md`, so whichever lands second takes a one-line conflict there. Landing 0076 first also keeps that table in numeric order.

PR #132, opened 2026-08-08
Price the no-flush bet, and decline it (JEF-750)

thejefflarson:
Closes JEF-750.

`CLAUDE.md`'s design commitments are explicitly amendable when a change moves all four goals forward together. **No wrong-path state** has never been priced, only restated. ADR-0076 measured that every input to `next_pc` is worth nothing alone and 21% together, and named the only route to that 21%: a pc that does not depend on this cycle's decode. This prices the narrow version of that — a registered fetch address and one killed instruction per redirect. No predictor, no skid buffer, no speculative pipeline.

**Verdict: do not do it. The blocker is #2, what it buys.**

## The five numbers

**1. CPI — +1.9%,** measured. 28632 → 29175 cycles at 13853 retires, CPI 2.0668 → 2.1060. The suite redirects on 767 issuing cycles (576 taken branches, 135 `jal`, 22 traps, 22 `mret`, 12 `jalr`) = 5.5% of retires. 543 of those cost a cycle; 224 were absorbed.

The open question was whether `operand_stall` hides the kill. **It does not**, and that is measured, not argued. Classifying the cycle after each of 708 redirects on the unmodified core: 447 issued (63.1%), 249 were operand-fetch bubbles (35.2%), 12 were time-based stalls (1.7%). The 35.2% is not cover — decode cannot present the target's register pair before the target's word arrives, so the bubble still happens, one cycle later. `jal` is hit hardest: 1 cycle → 2, on 135 of 13853 retires.

**2. What it buys — 13.32 MHz, against a 24 MHz bar.** Eight placements each, same toolchain, baseline reproduced from ADR-0076 to the hundredth of a nanosecond.

| variant | median ns | MHz | vs baseline | `fit` LC |
|---|---|---|---|---|
| baseline | 78.58 | 12.73 | — | 3880 |
| **K — implemented** | **75.07** | **13.32** | **−4.5%** | 4004 |
| K∞ — every redirect term off `next_pc` (ceiling) | 73.23 | 13.66 | −6.8% | 3656 |
| K∞E — that plus `stall` off too, i.e. + a skid buffer (ceiling) | 61.27 | 16.32 | −22.0% | — |

24 MHz needs 41.67 ns. The ceiling of the whole family is 61.27 ns — **47% short**, and between 12 and 24 the oscillator has nothing. The critical path under K is still `imem.in_range → imem.rom_even`: `stall` and `pc_inc` keep the fetch loop critical, and `stall` is fed by the whole decode head.

**3. Proofs — cheaper than expected.** F = 6 and G = 6 both re-measured (`hang` red at 6 / PASS at 7; `liveness` red at 5 / PASS at 6), so **no `[depth]` line moves** and `insn 19` / `reg 15 22` keep their one cycle of margin. Ladder **85/85**, both set equalities both directions. **`components_pcloop` still closes by k-induction with no hand-written invariant** — the increment guard actually gets simpler, becoming one state bit instead of a transcribed encoding list. `components_traps` goes red on `assert(pc == prev_mtvec)` — the property `CLAUDE.md` names as the only one saying a trap lands on `mtvec` — and is re-established by exporting two proof-only decoder ports and splitting it into two hops. `reg_ch0` liveness probe still fires (SAT at k=22 with the rs2 bypass deleted). `make cosim-suite` 59/59 AGREE.

**4. Trap commit — a trap does not become revocable.** The kill lands *before* issue, so the killed word never issues; one `&& !kill` on `issuing` suppresses `trap_entry`, `mret_entry`, `instret`, `csr_ren`, `csr_wen` and `out.valid` at once, plus one assertion. What breaks is the *proof that the commit lands*: it stops being a one-cycle statement because a freeze can hold the redirect.

**5. Readability — worse.** `rtl/decoder.v` +105/−36, `formal/pcloop.sv` +23/−8, `formal/traps.sv` +23/−4. Nothing is removed; the `next_pc` chain moves into `redirect_pc` and gains a register pair, a `hold_stall` term and a second priority chain. Four new concepts, including a kill that is not a stall and so is absent from the six-reason taxonomy.

## Two findings that outlive the decision

- **ADR-0076's 21% is entirely the fetch loop.** Its combined row deleted the decode head as well and landed at 61.86 ns; K∞E deletes no decode-head logic and lands at 61.27 ns.
- **`make cycles` has a blind spot.** A cycle that issues nothing without raising `stall` is counted as an *issue* cycle — 224 of them under K — while `stall` is still exactly the OR of six reasons, the columns still sum, and `unattributed` is still zero. Nothing in the tree can produce one today.

## What ships

The ADR. **Spike RTL is discarded**; `rtl/`, `formal/` and `test/` are byte-identical to `0b1728a`.

## How it was tested

All local (**CI's self-hosted runner pool has zero registered runners, so the checks here will queue**). On the spike: `make test` 59/59, `make cycles`, `make cosim-suite` 59/59, `make -C formal check` 85/85, `components_{decoder,executor,pcloop,traps}`, `nonperturbation`, `dmemcheck`, `imemcheck`, `make fit`, `soc/timing_sweep.sh` × 8 seeds per variant. After the revert: `make test` 59/59, `components_pcloop`, `components_traps`.

`make -C formal complete` / `complete_cover` did not run — the local Homebrew Yosys rejects `abc -g AND -fast`. Confirmed identical on the unmodified tree, so it is the toolchain, not the change; they need the pinned OSS CAD Suite.

### DIFF
--- a/docs/adr/0078-a-one-deep-kill-is-cheap-and-buys-a-clock-the-board-cannot-use.md
+++ b/docs/adr/0078-a-one-deep-kill-is-cheap-and-buys-a-clock-the-board-cannot-use.md
@@ -1,0 +1,252 @@
+# ADR-0078: A one-deep kill is cheap, and it buys a clock the board cannot use
+
+**Status:** Accepted · 2026-08-08 · *A measured null, and the first attempt to price the
+no-wrong-path-state commitment rather than restate it. Reads
+[ADR-0076](0076-the-decode-head-is-a-plateau-not-a-lever.md)'s 21% as its opening bid and prices
+what collecting it would cost. Graded against
+[ADR-0066](0066-twelve-megahertz-is-a-requirement.md)'s 12 MHz requirement and against 24 MHz, the
+next clock the board can actually run. CPI baseline from
+[ADR-0070](0070-the-suites-cycles-are-charged-to-a-named-stall-reason.md); depth derivation from
+[ADR-0046](0046-the-ladder-depths-are-re-derived-and-the-derivation-is-measured.md).*
+
+## What was proposed
+
+`CLAUDE.md`'s design commitments say they can and should change when a change moves all four goals
+forward together, and that none is held harder than the others. **No wrong-path state** is a
+commitment like any other, and ADR-0076 ended by naming the price of the only remaining lever:
+
+> Reaching [the 21%] means the pc no longer depends on this cycle's decode, and every way of
+> arranging that is state a later cycle has to un-commit.
+
+So: register the fetch address, let the sequential successor of a redirecting instruction be
+fetched, and kill it. One valid bit cleared, one cycle deep, no predictor, no skid buffer, no
+speculative pipeline. Call it **K**.
+
+Nothing else in the design changes. The pc still moves once per redirect and the target is still
+computed in decode; what moves is *when the fetch address sees it* — a cycle later, out of a
+register, instead of combinationally in the same cycle.
+
+## Decision
+
+**No.** K was built, run and measured. It is cheaper than expected on almost every axis the
+commitment was written to protect — the suite passes, the ladder passes at unchanged depths,
+`pcloop` still closes by k-induction, and a trap does not become revocable. One proof goes red:
+`components_traps` can no longer say a trap lands on `mtvec` in one cycle, and re-establishing it
+costs two proof-only decoder ports. K is declined anyway, on a different axis than either of those.
+
+**K measures 13.32 MHz. Its own function-breaking ceiling is 13.66 MHz. The ceiling of the entire
+family — including the skid buffer that would also take `stall` off the fetch address — is
+16.32 MHz.** The part's oscillator divides 48 by 1, 2, 4 or 8, so the clock above 12 is 24, which
+needs 41.67 ns. The best ceiling in this direction is 61.27 ns: **47% longer than 24 MHz needs, and
+there is nothing in between to land on.** Every variant here buys a frequency the board cannot
+select, at the price of a commitment.
+
+The spike RTL is discarded. What ships is this ADR.
+
+## Method
+
+Four designs, one tree, one toolchain: **Homebrew Yosys 0.68+post (`c12172fb`)**,
+nextpnr-0.10-108-g68c1acd8, `icetime` from the same install. `rtl/` and `formal/` are byte-identical
+to `2d79878`, which is what ADR-0076 measured, so its eight baseline placements are reused as the
+comparison distribution; four of them were re-run here and reproduced to the hundredth of a
+nanosecond (77.38 · 78.34 · 78.51 · 78.55). **The tables report periods, so a negative `vs baseline`
+is faster.**
+
+Two of the four are ADR-0076-style ceilings — function-breaking RTL that deletes the logic a
+candidate hopes to shorten, so that no implementation of it can do better:
+
+- **K∞** deletes every redirect term from `next_pc`, leaving `reset ? 0 : stall ? pc : pc + pc_inc`.
+  That is exactly what K takes off the fetch-address path and no more: `stall` stays, because
+  holding the fetch address on a stall is what re-presents the instruction, and the only alternative
+  is a buffer.
+- **K∞E** deletes `stall` as well, leaving `pc + pc_inc` unconditionally. That is the ceiling of K
+  *plus* a skid buffer — the change anybody who read K's result would propose next.
+
+**K itself is a working core**: 59/59 on `make test` with identical retire counts, 59/59 AGREE on
+`make cosim-suite`, 85/85 on the generated ladder.
+
+## What it costs in cycles
+
+Measured first, because it was the cheapest of the five questions and it was expected to be the
+expensive one. It is not.
+
+| | cycles | retires | CPI |
+|---|---|---|---|
+| baseline | 28632 | 13853 | 2.0668 |
+| **K** | **29175** | 13853 | **2.1060** |
+
+**+1.9%.** The suite redirects on 767 issuing cycles — 576 taken branches, 135 `jal`, 22 traps,
+22 `mret`, 12 `jalr` — which is 5.5% of retires and 2.7% of cycles. The kill costs 543 of those
+767; the other 224 land on a cycle the pipeline was going to spend anyway.
+
+**The operand-fetch cycle does not hide the kill, and that was the open question.** Classifying the
+cycle after each of 708 redirects on the *unmodified* core:
+
+| the cycle after a redirect | count | share |
+|---|---|---|
+| issued | 447 | 63.1% |
+| operand-fetch bubble | 249 | 35.2% |
+| decode scoreboard | 7 | 1.0% |
+| serialization | 5 | 0.7% |
+| divider / accessor / stolen fetch | 0 | 0.0% |
+
+The 35.2% looks like free cover and is not. `operand_stall` is decode *presenting* the target's
+register pair, and it cannot start until the target's word arrives — which under K is a cycle later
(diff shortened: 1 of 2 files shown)


thejefflarson:
## Integration review — first review, and the reconciliation ruling

**Not merged — no runner is registered, so six of seven required checks queue indefinitely.** (`formal` passed before the outage; `review` is green but not required.) Reviewed on the merits; approved subject to the two items in "Before this merges" below.

Every number recomputes from the placement lists as written: K's median 75.07 → 13.32 MHz, K∞ 73.23 → 13.66, K∞E 61.27 → 16.32, and 61.27/41.67 = 1.47. CPI cross-checks three independent ways — 29175−28632 = 543, which is the stated kill cost, and 543/28632 = 1.90%, and 767 = 576+135+22+22+12. That internal consistency is the strongest evidence here that the spike was really run.

### The 21% conflict: **this PR is right, and #131 has been amended**

`K∞E` deletes every redirect term *and* `stall` from the fetch address with the **whole decode head standing**, at 61.27 ns over four placements. ADR-0076's combined row deletes those *plus* A+B+C+D, at 61.86 over four. **0.95% apart, distributions fully overlapping** (60.77–62.38 against 60.83–62.81) — inside every band the instrument has. `K∞E` is also the stronger experiment: it deletes *more* of `next_pc` (including the `jal`/`jalr`/`mret` terms ADR-0076 never had a row for) while touching *less* of the design.

Fixed on #131's branch (`9da21e5`), which is where it belongs — both are unmerged, so this lands as a record rather than as a correction to one. **Nothing changes here.** ADR-0076 now carries a marked pre-merge amendment naming this ADR as the source, so your "read that structure correctly and had one term too many in the row that showed it" stays accurate against the amended file.

I also withdrew something you were more generous about than you needed to be: ADR-0076's row E concluded that *"a change that shortens only the fetch loop is still capped at about 1.5%"*. Your K measured −4.5% **implemented**. That sentence would have declined this ticket unbuilt, and it is now explicitly withdrawn.

### The verdict: sound, and robust to the objection it invites

The blocker is the right one and it is graded correctly. The obvious attack — that 13.32 MHz is a *median* while 12 MHz is graded on the worst placement — doesn't land, because the verdict survives every choice: K's best placement is 73.85 ns = 13.54 MHz, and the family ceiling's best is 60.77 = 16.46. Nothing in the family gets within 45% of 41.67 ns. Say that explicitly; it closes the one hole a reader will poke at.

The "nothing between 12 and 24" claim is consistent with the record — ADR-0066 already frames 12 as a divider step ("whose next divider step down is 6"). It is stronger than stated, too: the UltraPlus part has no PLL, so if the board's 12 MHz is a crystal rather than HFOSC/8 there is no multiplier to reach an intermediate frequency either. Either reading gives the same answer.

### `operand_stall` does not hide the kill: **reasoning verified**

This was the ticket's central question and the answer is right for the reason given. `operand_stall` is not a wait-for-something stall that elapsed time absorbs — it is a fixed pipeline step *downstream of the target word's arrival*, and K delays that arrival by a cycle. So the bubble translates; it does not overlap. Only stalls that resolve against in-flight work (scoreboard, serialization — the 12 cycles in your bottom rows) can absorb a kill cycle, and that is exactly the line your table draws. The 35.2% column looks like cover and is not, and the write-up explains why rather than asserting it.

One gap to close: **the table classifies 708 cycles but the suite redirects on 767**, and the 59 difference is never named. 767 − 708 = 59 is precisely the program count, which makes "the last redirect in each program has no classified successor cycle" the obvious explanation — but a reader who spots the mismatch and can't resolve it will distrust the whole table. One clause fixes it.

### `components_pcloop` closes by k-induction: **plausible and well-explained**

The mechanism given is the reason to believe it: induction gets *easier* because `redirect_pending` is real state the invariant can read, so the increment guard stops being a hand-transcribed encoding list and becomes one bit. That is a genuine structural simplification and it is the correct explanation for a counterintuitive result — a briefed "most likely to break" that instead got simpler is exactly the shape of result that needs its mechanism stated, and you stated it. Noted as the notable finding it is.

### `components_traps` and the proof-only ports: one thing to add

The account is accurate — `formal/traps.sv:334` is `always_comb if (settled && prev_trap_entry) assert(pc == prev_mtvec);`, a strictly one-cycle statement, and it is by `CLAUDE.md`'s own account the only thing in the tree saying a trap lands on `mtvec`. Two observations, neither a blocker, both worth recording since this is the artifact a future attempt starts from:

1. **The two-hop split is weaker than what it replaces, and the ADR should say so.** Today's assertion bounds *when* the trap lands: exactly one cycle. The split says the registered target is `mtvec` and that the pc takes the registered target — which, across an indefinite freeze, is "eventually" plus a liveness argument that lives in a different check. You note the unbounded gap honestly; call the result a weaker property rather than the same one re-established.
2. **Proof-only ports need their red direction named.** `redirect_pending`/`redirect_target` exported for the proof alone is the classic setup where hop (a) becomes a tautology — a decoder output that mirrors whatever the pc logic is about to do makes "the registered target is `mtvec`" true by construction, and the proof goes green having stopped asking. `nonperturbation` PASSing is the right control that the ports don't feed the core, but it is not a control against vacuity. This repo's rule is that a grader with no demonstrated red direction is not a grader, and five recorded defects came from exactly that. Record what the red direction would have to be, so the next attempt doesn't have to rediscover the requirement.

That the spike RTL is discarded is what keeps both of these off the blocker list.

### `make cycles` blind spot: **durably recorded, and correctly scoped**

Verified against the runner. `test/stall_report.py`'s header states the identity unconditionally — "A cycle where `stall` is low is an issue cycle" — and the finding is that this silently assumes every non-issuing cycle raises `stall`. Your reading that a seventh bucket would not be enough is right: the taxonomy needs a second *category*, because "issued nothing" would stop being "stalled".

The ADR consequence plus the README row is durable enough, and I am **not** asking you to add a tripwire to `test/stall_report.py`. "Nothing in `rtl/`, `formal/` or `test/` ships from this ADR" is load-bearing for this PR and worth more than moving the note 30 feet closer to the code. If the direction is ever reopened, that docstring is where it goes — noted here so this comment is findable from `git blame` on the runner.

### `complete` / `complete_cover`: stated rather than skipped — right call

Homebrew Yosys rejecting `abc -g AND -fast`, reproduced on the unmodified tree, is a clean environmental attribution. Both are in the required `formal` job, so CI's pinned OSS CAD Suite closes them the moment runners return — and `formal` is the one required check that already passed here, on this exact SHA. Nothing further needed.

### DECISION NEEDED — README ordering: **resolved, no change required**

Merge **#131 first, then this**. 0076 lands ahead of 0078, your row goes after it, and the table is never out of numeric order — the concern dissolves rather than needing a fix. Both PRs append at the same point in `docs/adr/README.md`, so this one takes a **one-line conflict** on the second merge; the resolution is to place the 0078 row immediately after 0076's. Leaving the branch as-is is the cheapest correct state, since 0076's row doesn't exist here to sort against.

For the record: that table already has gaps at 0072 and 0073, and 0077 is still on an unmerged branch and will insert between when it lands. Dense numbering was never the invariant — sorted order is, and this preserves it.

### Before this merges

Neither is a hold; both are one-or-two-sentence edits, and the ADR is better with them:

1. Name the 767 → 708 gap (`## What it costs in cycles`).
2. Say the two-hop trap property is weaker than the one-cycle statement, and name the red direction the proof-only ports would need (`## What it costs the proofs`).

Optional, and I'd take it: `CLAUDE.md`'s *Measurements and ratchets* section has no line saying that a frequency between 12 and 24 MHz buys nothing on this part. That is the most reusable sentence either of these ADRs produced and it will otherwise be rediscovered. I left it off #131 deliberately so each PR stays self-consistent if the other is held — raise it as its own ticket rather than growing this one.

PR #146, opened 2026-08-09
Take the operand-fetch cycle: 9.7% of suite cycles, 1.1% of Dhrystone's (JEF-793)

thejefflarson:
Closes JEF-793.

On a cycle that issues, decode now asks the register file for the **next** instruction's register
numbers instead of the pair it is issuing, so the operands arrive with that instruction rather than a
cycle behind it. `operand_stall` is unchanged to the character — it compares what the register file
was *asked for* last cycle against what the instruction now being decided about *reads*, and that
comparison never cared where the request came from, so the request can be a guess and the existing
test is the check on it.

ADR-0089 records it. ADR-0074 built this and declined it; this reverses that decline on a
re-measurement, and declines the second half of the ticket as unnecessary.

## Where the spelling came from

**It is not recoverable verbatim.** `2d79878` landed no `rtl/` change, and `git log --all -S regsel`,
`-S next_rs1`, `-S read_rs1`, every remote branch and all 217 dangling commits in the clone turn up
the ADR and nothing else. It was rebuilt from ADR-0074's own "How it was made safe" section, which
states the mechanism completely: `operand_stall`'s formula, the `read_rs1 = stall ? rs1 : next_rs1`
rule, the reason the mux selects on `stall` rather than `operand_stall` alone, and variant 2's flat
`instr[19:15]` / `instr[24:20]` guess with a compressed successor coming out as x0/x0.

That last clause is load-bearing and I found out the hard way. Built first **without** the masking —
the raw successor word, fields sliced straight out — this is a null: 32624 → 32560 cycles, with the
operand column going *up*, and `beq.S` costing 74 cycles more than today. The design being replaced
is not "every instruction pays a cycle": it pays one unless the pair it reads is the pair presented
last cycle, and in this suite two thirds of instructions already hit that way because so many read
`x0`. A raw slice of a compressed successor is two plausible register numbers belonging to the
instruction *after* it, so it both misses and destroys the hit the old behaviour would have had.
Zeroed above the low half — exactly as decode masks its own word — the operand column goes
5287 → 1650, reproducing ADR-0074's 4684 → 1188.

## Throughput, as a product

**The board runs at 12 MHz.** `SB_HFOSC`'s next step up is 24 and the crystal is 12, so a placement
that closes at 13.2 MHz does not run 10% faster than one at 12.0 — it runs at 12.0 with more margin.
Fmax above the requirement is margin, not speed, so wall time is cycles × 83.33 ns and the clock term
is a constant:

| workload | today | with the guess | throughput |
|---|---|---|---|
| `.S` suite (`make cycles`) | 32624 cy = 2.719 ms | 29454 cy = 2.454 ms | **×1.108** |
| Dhrystone, 2000 runs | 2203492 cy = 183.6 ms | 2178739 cy = 181.6 ms | **×1.011** |

Priced instead at each side's own median placement — the reading to use if the clock were free — the
suite is ×1.093 and **Dhrystone is ×0.998, a wash**. Neither reading makes Dhrystone move.

`make dhrystone`: **0.529 → 0.535 DMIPS/MHz**, and in absolute terms at the board clock
**6.35 → 6.42 DMIPS**. Same compiler, same flags, same string routines, 3568 bytes of the 8 KB ROM.

## What this really bought

**9.7% of cycles on hand-written assembly and 1.1% on compiled C.** The gap is not noise and it is not
about Dhrystone in particular — it is the C extension. Compiled code is dense with compressed
instructions, this guess is x0/x0 for every one of them, and x0/x0 is right only when the successor
really does read `x0`. ADR-0084's warning earns its keep in the direction that hurts: read 9.7% as a
property of the suite, never as the prize. The prize is 1.1%.

The version that *would* move compiled code is the one ADR-0074 measured at 11.56 MHz. It is not
reopened here: today's `main` is slower at its worst placement than the tree that variant was
declined on (81.61 ns against 79.77), so there is no reason to think the clock has come back.

## Cycle accounting, both workloads

| | cycles | CPI | operand | scoreboard | unattributed |
|---|---|---|---|---|---|
| `.S` suite, today | 32624 | 2.08 | 5287 (16.2%) | 9561 (29.3%) | 0 |
| **`.S` suite, after** | **29454** | **1.88** | **1650 (5.6%)** | 10027 (34.0%) | 0 |
| Dhrystone, today | 2203492 | 2.32 | 671373 (30.5%) | 352874 (16.0%) | 0 |
| **Dhrystone, after** | **2178739** | **2.29** | **641153 (29.4%)** | 358928 (16.5%) | 0 |

Still six stall reasons, nothing added to `stall`, every stalled cycle attributed. The scoreboard
column going *up* while the operand column falls is the non-separability ADR-0084 named, made
visible: a cycle that is both is charged to the scoreboard.

## Ten placements a side, and why the policy split is not made

| | ns, ten seeds | median | worst |
|---|---|---|---|
| `main` at `4a59607` | 75.12 · 75.67 · 76.08 · 77.17 · 77.53 · 78.34 · 78.38 · 78.59 · 79.85 · 81.61 | 77.94 | **81.61 / 12.25 MHz** |
| this branch | 75.71 · 76.06 · 76.75 · 77.08 · 78.89 · 79.05 · 79.08 · 79.22 · 79.47 · 79.58 | 78.97 | **79.58 / 12.57 MHz** |

Median +1.3%, worst −2.5% — both inside the 3.6% edit-churn band, so a null in both directions.
`make fit` 3874 → 3859, inside its ±50 band.

**Margin at the shipped placement: `make soc-timing` with no `SOC_SEED` — the placement CI grades and
the bitstream comes from — is 79.05 ns, 12.65 MHz, 5.4% over the requirement.** Worst of ten is 4.75%.

So the ticket's second half is **not built**. Every placement clears 12.0 with more margin than
merged `main` has today at its own worst placement, and changing the grading policy to admit a design
that does not need admitting would spend a rule to buy nothing. The observation underneath is written
into ADR-0089 for when it is actually needed: **worst-of-N is not a property of the artifact** — four
seeds report this tree's `main` at 12.72 MHz and ten report 12.25, with nothing about the design
different between those sentences.

## The correctness argument, restated

ADR-0064 says the write-through bypass is correct because `operand_stall` guarantees the held pair is
the presented pair on every issuing cycle. **That sentence is now false and the bypass is still
correct.** The true one is ADR-0074's: `operand_stall` lets nothing issue until the held pair is the
pair the issuing instruction reads. The two were the same sentence only while decode asked for the
instruction in front of it.

`test/regfile_tb.v`'s two "keyed to the held address" vectors stop being a curiosity — before this,
nothing in the core pointed a port elsewhere during a use cycle, and now every issuing cycle does.
**`reg_ch0`'s standing liveness probe was run before any of this was believed**: delete the rs2
write-through bypass and it reports `bad state property 1 reachable at bound k = 22 SATISFIABLE`.

`formal/pcloop.sv` needs both pairs, which is why the decoder publishes both. Register the guess on
*both* sides of its `f_operand_fetch` comparison and the harness excuses the pc from advancing on
nearly every issuing cycle — a proof that has stopped asking rather than one that fails.

## F and G, re-derived

ADR-0074 argued from the design's shape that the depths were safe and said a shipping version still
owed the measurement. Paid, by `checks.cfg`'s own recipe:

- **F = 6** — `hang` red at depth 5 and 6, PASS at 7 and 8.
- **G = 6** — `liveness` red at gap 4 and 5, PASS at gap 6 and 7, at trig 10 and again at trig 15.

Both flip points reproduce exactly, so `insn 19` and `reg 15 22` still clear `F + 2G = 18` by the same
one cycle and **no depth moves**. A wrong guess raises `operand_stall` for exactly one cycle and then
the current pair is presented, so the longest run of consecutive operand-fetch cycles is still one.

## Tests

Two new `test/decoder_tb.v` vectors, both of which fail without this change: one drives
`in.next_instr` with the successor's encoding and checks `read_rs1` is the *successor's* rs1 while the
decoded `rs1` is still the issuing instruction's, then that the successor issues with no
operand-fetch cycle; the other drives a wrong guess and checks it pays the cycle and issues on the
next one. Every other vector leaves `in.next_instr` at zero, which the file header now states,
because driving it elsewhere makes the following vector issue a cycle early.

## Checks

- `make test` 62/62, `test/EXPECTED_FAIL` matching · `make test-units` 8/8 · `make probe-gates` 188
  graded comparisons, every failure path executed · `make elaborate-strict` clean · `make lint`
  clean in both svlint passes
- `make -C formal check` **85/85**, both set equalities matching in both directions
- `components_decoder` · `components_executor` · `components_pcloop` · `components_traps` — each a
  successful proof by k-induction. `nonperturbation` and `complete_cover` PASS
- `make cosim-suite` **60/62 before and after**, matching `test/COSIM_EXPECTED_FAIL`. ADR-0074 could
  not run this leg and asked a future attempt to carry it; it is the one positioned to see a stale
  operand rather than a wrong one
- `make -C formal complete` does **not** run on this machine — the local Homebrew yosys rejects abc's
  `-g AND -fast` — and fails identically on unmodified `main`. Relying on CI

No baseline file was edited.

## The one thing to look at first

CLAUDE.md's design commitment 6 said the held pair *is* the presented pair on every issuing cycle.
That is no longer true, so the sentence is corrected in place — the commitment's substance is
untouched (synchronous read, the answer belongs to the pair presented the previous cycle,
`operand_stall` load-bearing, narrowing it breaks the register file with only two vectors and
`reg_ch0` to say so), and only the explanation of *which* pair is presented changes. Commitment 4's
two percentages moved with the measurement (29.3% → 34.0% suite, 16.0% → 16.5% Dhrystone) and its
"sum of the two columns is flat across the two workloads" clause is replaced by the sharper fact this
change demonstrates. Flagging it explicitly because the ticket asked for no commitment to be amended
and this is the closest thing to one in the diff.


thejefflarson:
## `elaborate` fix pushed — root cause was one undefined ROM word

**It is one word, and it was pinned before it was fixed.** `test/testbench.v` pokes a six-word program into a 4096-word array and leaves the rest undefined. The last instruction is `j loop` at byte `0x14`; the word after it is `0x18` — `rom_even[3]` — which nothing ever wrote. This branch reads the register fields of the word *after* the instruction being issued, so that X reached the register file's address port, `held_rs1` went X, and the comparison that should have **failed** was undefined instead. Defining that single word and nothing else turns the run green at 79 retires, which is how the mechanism was confirmed rather than assumed.

### The harness was wrong, not the RTL — argued, not asserted

- **Arbitrary contents are already proved inert.** `formal/wrapper.v` declares `imem_data` and `imem_data2` as `rvformal_rand_reg` — free every cycle, constrained only by same-address stability — so the solver picks adversarial contents for that successor word on every step of all 85 generated checks, and they pass. Whatever it holds, `operand_stall` compares it against the decoded pair and stalls when they differ. A garbage guess costs a cycle; it cannot cost correctness.
- **X is not a hardware value.** It is the one value that makes that comparison *undefined* instead of *false*, and an ice40 EBR cannot produce it — every word comes out of the bitstream.
- **The repo already knew.** `soc/compare/rom_flat.py` zero-pads its ROM image to full depth on purpose. The poked harness was the one place that did not.

So both banks are zeroed before the program goes in. I did **not** mask X in the RTL, and I did not narrow what the guess may read — no `rtl/` file changed in this fix, `operand_stall` is byte-identical, and the write-through bypass still selects on the held pair with the same argument.

On the ROM_WORDS asymmetry you flagged: it is real but it lives in the harness. `rtl/imemory.v` is the *same module* in both, and the difference is that one caller loads a full-depth image and the other poked six words.

### Nothing moved

Base RTL **and** this branch both report `RETIRES 79 SPEC-CHECKED 79 WRITES 20` under the fixed harness, so the fix is not hiding a behaviour change — the guess neither helps nor hurts a four-instruction loop whose only redirect misses every time.

| | before the fix | after |
|---|---|---|
| `make cycles` | 29454, CPI 1.88 | **29454, CPI 1.88** |
| `make fit` | 3859 | **3859** |
| `make cosim-suite` | 60/62 | **60/62** |

So it does not cost the win, and there is nothing to recommend reverting over on those grounds. The honest verdict is unchanged and still the thing to weigh: **×1.108 on the suite, ×1.011 on Dhrystone.**

### Full re-run, iverilog leg included

`make waves` **79 retires / 20 writes** (floors 60 / 15) · `make test` 62/62 baseline-exact · `make test-units` 8/8 · `make probe-gates` 188/188 · `make compare-smoke` both cores publishing the same six values · `make elaborate-strict` and `make lint` clean · `make cosim-suite` 60/62 · **`make -C formal check` 85/85** both set equalities · `components_decoder` / `_executor` / `_pcloop` / `_traps` each a successful proof by k-induction · `nonperturbation` and `complete_cover` PASS · **`reg_ch0`'s liveness probe re-run: `bad state property 1 reachable at bound k = 22 SATISFIABLE`** with the rs2 write-through bypass deleted. `make -C formal complete` still does not run locally (abc `-g AND -fast`) and fails identically on unmodified main.

The lesson is now in CLAUDE.md beside the description of the three legs: cxxrtl is two-state, `elaborate-strict` does not simulate, and the formal harnesses drive that word as a free *two-state* input — all three were structurally incapable of seeing this. The iverilog leg is not a slower copy of cxxrtl; it is the only four-state simulation in the tree.

### On #147 and the rebase

`origin/main` is still `4a59607`, so this branch is already on the latest main and there is nothing to rebase onto yet. **Flagging a merge-order dependency:** #147 moves `FIT_MAX_LC` to 3700 and this branch measures 3859 against today's 4100. My delta is −15 cells, so once #147 lands the combined tree should sit near its number — but that has to be measured, not assumed. If #147 merges first, ping me and I will rebase and re-run `make fit` and a fresh sweep before you merge this.

thejefflarson:
## CI is green — all eleven checks pass

CI's own numbers, which are the ones to quote since the pinned OSS CAD Suite differs from a local Homebrew yosys:

- **soc-timing: 76.16 ns, 13.13 MHz** against the 12.00 MHz requirement — 9.4% of margin on the placement CI grades. (Local default placement was 12.65 MHz; the toolchain gap runs in this direction.)
- **fit: 3906 cells** of the 4100 budgeted. Local was 3859, so the pinned toolchain reads ~47 cells higher on identical RTL — the same axis CLAUDE.md records at ~21.

### Merge-order warning, now with the number that matters

**#147 moves `FIT_MAX_LC` to 3700 and this branch measures 3906 on CI.** Whichever of the two merges second has to re-measure, and if it is this one it will trip the ratchet until it is rebased. #147 removes cells, so the combined tree should land under 3700 — but that is an expectation, not a measurement, and this repo's rule is that a cell count is measured or it is nothing.

I have not rebased because `origin/main` is still `4a59607`; this branch is already on the latest main. **Ping me when #147 lands** and I will rebase, re-run `make fit`, and re-run a fresh ten-seed `soc/timing_sweep.sh` — the netlist changes under a rebase and the margin argument in ADR-0089 rests on that distribution, so it should not be inherited across someone else's datapath change.

Not merging and not enabling auto-merge; the merge decision is yours.

thejefflarson:
## Rebased onto `a8d6f46`, every number re-derived — all eleven CI checks pass

The rebase was clean on code (`rtl/decoder.v` vs `rtl/executor.v` never met); the only conflict was the ADR index, resolved in numeric order. **ADR-0089 was still free** — #147 took 0090 — so no renumbering. I did not touch the Makefile.

### Area, and the ratchet

**`make fit`: 3470 → 3496 locally, +26 cells. CI reads 3487 against the 3700 ratchet.**

**I left `FIT_MAX_LC` at 3700** and want to be explicit about why, since you asked me to set it from what I measure: 3487 clears it with 213 cells of headroom, four times the ±50 churn band, so there is nothing to fix and nothing to report as a finding. Tightening #147's ratchet on the way past would be re-deciding someone else's number with no measurement of my own that bears on it.

Worth flagging as a correction to my own earlier claim: pre-rebase I measured **−15** cells and post-rebase **+26**. Both are inside the band, which means the honest statement is *no measured area change*, and my earlier PR body's implied saving was over-reading a null. The ADR now says so.

### Timing — ten seeds a side, both re-run on `a8d6f46`

| | ns, ten placements | median | worst | shipped placement |
|---|---|---|---|---|
| `main` at `a8d6f46` | 74.42 · 74.71 · 75.44 · 75.62 · 76.63 · 76.68 · 76.68 · 77.25 · 77.30 · 79.57 | 76.66 ns | 79.57 / 12.57 MHz | 75.62 / 13.22 MHz |
| this branch | 76.61 · 77.51 · 77.94 · 78.28 · 78.61 · 78.64 · 78.71 · 79.25 · 79.29 · 79.30 | 78.63 ns | **79.30 / 12.61 MHz** | **79.30 / 12.61 MHz** |

**Margin at the shipped placement: 12.61 MHz, 5.1% over the requirement** — and here the shipped placement *is* the worst of the ten, so the two readings the proposed policy split was meant to separate agree exactly. CI's pinned toolchain reads it faster still, at **77.61 ns / 12.88 MHz**.

**One thing changed in my reading and you should have it.** The median moved **+2.6%** here against **+1.3%** pre-rebase. Each is inside the 3.6% churn band and so a null by the repo's rule — but two independent trees both moved positive, which I now read as *a small real period cost the band cannot resolve*, not as zero. The ADR says that instead of the flat 'null in both directions' it said before. It remains nowhere near the 12 MHz requirement, which is the claim that actually matters.

### Cycles did not move — measured on both sides, not assumed

| | base on `a8d6f46` | with the guess |
|---|---|---|
| `.S` suite | 32624, CPI 2.08 | **29454, CPI 1.88** |
| Dhrystone | 2203492, CPI 2.32 | **2178739, CPI 2.29** |
| DMIPS/MHz | 0.529 | **0.535** (6.42 DMIPS at 12 MHz) |

Identical to the pre-rebase figures on both sides, so ADR-0090 touched no cycles — confirmed rather than assumed, as you asked.

### Does the verdict change? Slightly, and against the change

**The 12 MHz product is unchanged: ×1.108 on the suite, ×1.011 on Dhrystone.** That is the reading that matters, because the board clock is fixed and Fmax above 12 is margin rather than speed.

But priced at each side's own median placement — the reading to use if the clock were free — it is now **×1.080 on the suite and ×0.986 on Dhrystone**, where pre-rebase that second figure was ×0.998. So on the workload that resembles real code, that alternative reading has gone from *a wash* to *1.4% slower*. It does not flip the headline verdict, and I am not going to pretend a median inside the churn band is a measurement of 1.4%. But it is the reading that goes the wrong way and the architect should have it explicitly rather than find it in a table.

My recommendation is unchanged and unenthusiastic: this is **9.7% of cycles on hand-written assembly and 1.1% on compiled C**, for a small period cost the instruments cannot resolve and no area cost. If the fable architect weighs 1.1% on Dhrystone against a permanent extra step in the register-file correctness argument and declines it, that is a defensible read of the same numbers, and ADR-0089 is written so the decision survives either way.

### Every gate, re-run on the rebased tree

`make test` 62/62 baseline-exact · `test-units` 8/8 · `probe-gates` 188/188 · **`make waves` 79 retires / 20 writes** (the leg that caught the X) · `compare-smoke` both cores agreeing · `elaborate-strict` + `lint` clean · `cosim-suite` **60/62** · **`make -C formal check` 85/85** both set equalities · `components_decoder` / `_executor` / `_pcloop` / `_traps` each a successful proof by k-induction · `nonperturbation` + `complete_cover` PASS · **`reg_ch0`'s liveness probe SAT at k = 22** · **F and G swept again** (hang red 5/6, PASS 7/8; liveness red gap 4/5, PASS gap 6/7 at trig 10 and 15) — unmoved, which also confirms the narrower divider registers did not move them. `make -C formal complete` still does not run locally and fails identically on unmodified main.

**CI: all eleven checks pass.** Not merging and not enabling auto-merge.

PR #166, opened 2026-08-11
Stand up the rv32imac co-simulation config, and observe Sail's reservation policy (JEF-811)

thejefflarson:
Closes JEF-811.

The A extension is coming, and co-simulation compares architectural register values — an SC's result **is** one, 0 for success and nonzero for failure. If Sail dropped its reservation where this core keeps it, every LR/SC program would diverge on a correct core, on a **required** check, discovered in the middle of the hardest ticket of the extension. This answers it before any RTL exists.

## The finding: Sail mirrors this core

Asked the model directly rather than read its source. `test/sail/reservation_probe.S` runs under Sail **alone** — no core is involved and none could be, since this core does not decode `lr.w`/`sc.w` yet. Six cases, one architectural event each.

```
$ make sail-reservation-probe
isa:    rv32imc_zicntr_zicsr_zifencei_za128rs_za64rs_zaamo_zalrsc_zca_zvl32b_...
command: sail_riscv_sim --config rv32imc_zicsr.json --inst-limit 5000 probe.elf

HTIF located at 0x10000
FAILURE: 79 (0x0000004f)

sc.w back to back with the lr.w (control)      SUCCEEDED
sc.w inside the trap handler, no mret          SUCCEEDED
sc.w after trap entry and mret                 SUCCEEDED
sc.w after an mret with no trap behind it      SUCCEEDED
sc.w after a same-hart sw to the reserved word failed
sc.w one word up from the reservation          failed
```

**Sail clears the reservation on neither trap entry nor `mret`** — this core's settled policy exactly. So there is nothing to mirror, no baseline entry, and no redesign. The two events are *separated* rather than inferred from each other: one case runs the `sc.w` inside the `ecall` handler (no `mret` has executed), another runs it after a bare `mret` with `mepc` set by hand.

Bit 0 is an anti-vacuity control and the runner exits nonzero without it. Its red direction is demonstrated: with `Zaamo`/`Zalrsc` off, `lr.w` is illegal and the model prints `FAILURE: possible trap loop detected`, which the runner rejects as "nothing was observed" rather than decoding as a mask of zero.

## What would actually have gone red: `misa`

Turning the extension on surfaced a different divergence. **`A` is not an umbrella in this model — it is the `misa.A` bit and nothing else.** All four combinations measured:

| `A` | `Zaamo`/`Zalrsc` | `misa` | `lr.w`/`sc.w` |
|---|---|---|---|
| true | true | `0x40001105` | execute |
| true | false | `0x40001105` | execute |
| false | true | `0x40001104` | execute |
| false | false | `0x40001104` | illegal instruction |

`rtl/csrs.v` returns `0x4000_1104`, so `A: true` diverges `csr.S` at change #17 — the same symptom ADR-0043 was written to end, re-created deliberately. **So `Zaamo` and `Zalrsc` are on and `A` stays off**, flipping in the change that makes the core claim the bit. Turning it on early would also hide that transition: the change making the core claim A would be green either way.

**This deviates from the ticket's literal wording** (turn all three on) in order to satisfy its acceptance criteria (byte-identical verdicts, no `misa` change). Flagging for the architect.

## Acceptance criteria

- **`make cosim-suite` verdicts byte-identical before and after** — the graded tables `diff` clean, 64 lines each: `60/62 agreed`, `Divergence list matches test/COSIM_EXPECTED_FAIL exactly`. `test/COSIM_EXPECTED_FAIL` untouched.
- Clear-on-trap-entry and clear-on-`mret` recorded as observations with the command and output.
- ADR at `docs/adr/0102-…` (0099–0101 are claimed by #164; checked before picking).
- **No RTL, no `-march` change at any build site, no `misa` change** — `git diff` over `rtl/`, `test/asm/`, `test/run_tests.sh`, `test/cosim.py` and all three baselines is empty.
- `make test` 62/62, `make test-units` 9/9.

## Also in the diff

`platform.reservation` now describes this core (one-word set, exact matching, invalidate on same-hart store), each value measured by flipping it — the set size and the exact-address knob overlap, and only measuring both shows it. Misaligned AMO and LR/SC get the causes this core raises; the upstream default let a misaligned AMO complete without faulting at all.

Region PMAs are deliberately untouched — refusing reservations outside RAM and faulting out-of-region accesses belong to other tickets, and the ADR names them as open.


thejefflarson:
Rebased onto `a975460`. The only conflict was `docs/adr/README.md`, the append-a-row table — both rows kept, in order.

**ADR number unchanged.** #164 landed `0099`, `0100` and `0101` on `main`; `0102` is still free, so it stays and no cross-reference moved.

**Everything re-run on the new RTL rather than carried forward.** #164 rewrote `rtl/accessor.v` substantially, so "identical verdicts" had to be re-established, not inherited. Both runs below are on this rebased tree; the only variable between them is the Sail config (`main`'s versus this branch's):

| | verdicts | baseline |
|---|---|---|
| before (`main`'s config) | 60/62 agreed | matched both directions |
| after (this branch's) | 60/62 agreed | matched both directions |

Graded tables `diff` byte-identical, 64 lines each.

- `make test` — 62/62, `test/EXPECTED_FAIL` matched
- `make test-units` — 9/9
- `make sail-reservation-probe` — same six answers as before the rebase (`FAILURE: 79`), so the finding is unchanged against the new accessor
- Probe's red direction re-run: with `Zaamo`/`Zalrsc` off, `FAILURE: possible trap loop detected with MEPC=0x30`, rejected as "nothing was observed"

No RTL and no `test/OBSERVED_FLOOR` in the diff — still the seven files it was. `A` stays `false` with `Zaamo`/`Zalrsc` on, as approved.

PR #182, opened 2026-08-14
Derive FIT_MAX_LC from a band measured on the tree, and report the move (JEF-828)

thejefflarson:
Closes JEF-828.

`FIT_MAX_LC` is derived instead of rounded, from a band and an instrument gap both measured on the trees it grades. Rebased onto **`421947f`**.

## The ticket's premise, demonstrated by this PR's own lifetime

`fit` is a **required check**. `FIT_MAX_LC` has been **4000 against a job reading 3966 — 34 cells — for every hour this PR has been open**, across nine merges to main. A functionally-null edit could have blocked any one of them: the measurements below show one bit of a read-only constant moving the count 53 cells on one toolchain and 63 across probes on another. Nothing tripped it, which is luck rather than margin.

## CI's number

**The `fit` job reads 3966** on `421947f` — [run 31984969180](https://github.com/thejefflarson/little-cpu/actions/runs/31984969180). It has read **3966 on five consecutive trees** — `fe618f4`, `4c1958c`, `9e26cfd`, `c13ed54`, `421947f` — none of which touched an input to `make fit`. Five CI runs, one number: the figure is stable, and this PR has simply been chasing a moving *base*, not a moving measurement.

## The new value

```
4088 = 3966 + 68 + 54
```

- **3966** — the `fit` job's count on `421947f`, run 31984969180.
- **+68** — the widest span measured for edits that change no logic: 68 on `64759da` (3988, 3979, 3920 against a base of 3935) and 63 on `2007d9d` (3958, 3971, 3987, 3925, 3980 against 3988).
- **+54** — the widest gap measured between two toolchains on one tree, `2007d9d`: the job read 3934 where both a local Homebrew yosys and a cached OSS CAD Suite read 3988.

**68 over the narrower 63:** every probe on that tree came out *below* its base, so a count can sit anywhere in the window with the whole span still ahead of it — budget the span, not its upward half. In the Makefile comment so it is not re-litigated from the smaller number.

**Cross-check, stated as the property to preserve:** the budget clears the higher of *each* tree's instrument pair by more than a band — 3969 here, 3988 on `2007d9d`.

## The instrument gap is not an offset

| tree | `fit` job | local | gap |
|---|---|---|---|
| `d3a9556` | 3543 | 3575 | 32, local higher |
| `64759da` | 3938 | 3935 | 3, job higher |
| `2007d9d` | **3934** | **3988** | **54, local higher** |
| `421947f` (this base) | 3966 | 3969 | 3, local higher |

The sharp form: the one-bit `misa` edit between `64759da` and `2007d9d` moved the local count **+53** and the job's **−4**. Three builds of yosys 0.68 — CI's `0.68+64`, a cached OSS CAD Suite `0.68+48`, Homebrew `0.68+post` — reading one design three ways. Adding a fixed offset to a local run was never a valid way to predict the graded number, which is why the derivation takes both terms from measurements rather than from the recorded 32.

## The trend line — optional improvement (2), taken

`FIT_LAST_LC := 3966`, the job's count on the merge tree, so the first report is not wrong by 32 cells. `make fit` now prints:

```
TREND: +3 cells against the 3966 the Makefile records -- a move inside
the churn band is noise, and that count is the `fit` job's, so a local delta
carries the gap between the two instruments. A diagnostic: only the budget
below can fail.

RATCHET: 3969 of 4088 cells budgeted -- OK
```

On an earlier base it printed `+54` locally and `+0` in CI — the same diagnostic reading the instrument gap from both sides. **Not a second ratchet**: three probes pin that it prints beside the verdict on a pass, that a **+875** delta still exits 0, and that a run tripping the budget prints the trend too.

`PROBES_EXPECTED` **267 → 270** (main's value plus these three). If anything touching `test/probe_gates.sh` lands before this merges, the resolution is that file's new value plus 3 — the only line in this diff that collides with anything.

## Also in this diff

- **`CLAUDE.md`'s `make fit` bullet** — the rule this derivation quotes: ±50 is nominal and the band measures 63–68; budget the whole span; the toolchain gap is as wide as the band with no fixed size or sign, with all four measurements and the +53/−4 pair as evidence. It now **points at `soc/baseline_sweep.sh` and `soc/baseline_summary.py`** rather than describing the problem abstractly — that PR built the mechanism this argues for on the other instrument, stamping each sweep's base and resolved tool versions and refusing to subtract two whose stamps disagree. Sits beside #187's and #189's additions to the same section without touching them.
- **One stale sentence in `.github/workflows/ci.yml`.** The `fit` job's step summary told every reader `it is NOT in main's required set`. It is — which is exactly why a churn trip blocks a merge.

## Decisions I made alone

1. **Used the measured 54 rather than the recorded 32** for the instrument term; ratified once, and a fourth measurement has since confirmed the gap is re-mapping rather than an offset.
2. **68 over 63**, on the every-probe-below-base argument.
3. **Did not fold in the Fmax-across-toolchains observation** from the atomic-region work: it is a second sign of the same instrument-dependence, but it belongs to the `soc-timing` bullet and its own ADR, and citing that from here would duplicate the record.
4. **No ADR**, ratified: a ratchet value is not a design commitment, and the band and instrument facts belong in CLAUDE.md's measurements section.

## Gates

| gate | result |
|---|---|
| `make test` | **70/70**, `EXPECTED_FAIL` exact both ways |
| `make test-units` | 9 benches, matching `test/*_tb.v` exactly, all PASS |
| `make probe-gates` | **270** graded comparisons, every failure path executed |
| `make fit` | **3969 of 4088** local; TREND +3 |
| `/soundcheck:pr-review` | no Critical or High findings |

Nothing in `rtl/` changed, so no baseline, depth, exclusion or `SOC_MIN_MHZ` moves.

**Two instrument notes**, follow-ups rather than this diff:

- A Homebrew boost upgrade broke `/usr/local/bin/nextpnr-ice40` mid-ticket (`Symbol not found: __ZN5boost15program_options3argE`); local figures here come from the cached OSS CAD Suite. `make fit` reported the break as `no utilisation table, so NO measurement was taken` rather than as a 0% fit — that graded comparison doing its job.
- `mutation-check` is timing-marginal on `little-cpu-runners`: it failed on this branch once with the step never completing after 14m01s and passed on re-run in 7m20s, against main's 8m27s; #180 lost it twice at 20m01s the same way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -369,7 +369,9 @@ jobs:
             cat "$RUNNER_TEMP/fit.txt"
             echo '```'
             echo "Logic cells on up5k/sg48, memories external. This job is a ratchet"
-            echo "against \`FIT_MAX_LC\`, and it is NOT in main's required set."
+            echo "against \`FIT_MAX_LC\`, and it is in main's required set -- so a"
+            echo "re-mapping that moves the count blocks a merge, which is why the"
+            echo "budget is derived to clear a whole churn band."
           } >> "$GITHUB_STEP_SUMMARY"
           exit "$status"
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -334,10 +334,25 @@ and times.
 
 - **`make fit` has a churn band of about ±50 cells**: functionally identical edits move the count
   that much from ABC/nextpnr re-mapping alone. A delta inside the band is not evidence of
-  anything, and a ratchet (`FIT_MAX_LC`) must sit outside it. **The number is also
-  toolchain-dependent** — the `fit` job reads 3543 on `d3a9556` where a local Homebrew yosys reads
-  3575 on the same tree, 32 cells apart with the local one higher — so quote the `fit` job's number;
-  a local run is a sanity check. **The suite is not pinned** — CI installs the latest release — so
+  anything, and a ratchet (`FIT_MAX_LC`) must sit outside it. **±50 is the nominal figure and the
+  band measures wider than it**: setting one further bit of the read-only `misa` constant spans 68
+  cells on `64759da` (3988, 3979 and 3920 against a base of 3935) and 63 on `2007d9d` (3958, 3971,
+  3987, 3925 and 3980 against 3988), all of them edits that change no logic at all — which is why
+  `FIT_MAX_LC` is derived from a span measured on the tree rather than from ±50. Budget the whole
+  span rather than its upward half: every probe on the second tree came out *below* its base, so a
+  count can sit anywhere in that window including the bottom.
+  **The number is also toolchain-dependent, by as much as the band** — the `fit` job reads **3934**
+  on `2007d9d` where both a local Homebrew yosys and a cached OSS CAD Suite read **3988**, 54 cells
+  apart on one tree; other trees read 32 apart (3543 job, 3575 local on `d3a9556`) and 3 apart
+  twice, with the sign not the same either time (3938 against 3935 on `64759da`, 3966 against 3969
+  on `421947f`). The sharp form: the one-bit `misa` edit
+  between `64759da` and `2007d9d` moved the local count **+53** and the job's **−4**, so the gap is the
+  same re-mapping and has no fixed size or sign. Quote the `fit` job's number, budget for the gap's
+  size rather than its direction, and treat a local run as a sanity check. The other instrument
+  answers this with tooling rather than a convention now, and it is the same argument:
+  `soc/baseline_sweep.sh` stamps a sweep with its base commit and its resolved tool versions, and
+  `soc/baseline_summary.py` **refuses** to subtract two sweeps whose stamps disagree.
+  **The suite is not pinned** — CI installs the latest release — so
   `FIT_MAX_LC` and `SOC_MIN_MHZ` are graded against a toolchain that can move under them, and a
   suite bump belongs on the list of causes when either trips. **And it is top-dependent**: one
   decode edit measured −50 `SB_LUT4` synthesising `littlecpu` and −1 synthesising `littlesoc`, both
--- a/Makefile
+++ b/Makefile
@@ -603,32 +603,52 @@ fit.json: $(FIT_SRCS)
 	@yosys -p 'read_verilog -sv $^; synth_ice40 -dsp -top littlecpu -json $@' \
 	  > fit.synth.log 2>&1 || { tail -40 fit.synth.log; exit 1; }
 
-# 3625 is a measurement of 3543 under CI's suite plus a 50-cell churn band plus
-# 32 for the toolchain: identical RTL moves about 50 cells on ABC re-mapping
-# alone, and that tree read 3543 under CI against 3575 under a local Homebrew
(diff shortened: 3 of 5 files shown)


thejefflarson:
## The re-derivation for #180's tree is measured and ready; #180 cannot land yet

This branch is derived against `64759da`, which is still `origin/main`. #180 is armed for auto-merge but its `mutation-check` has been **cancelled at exactly 20m01s twice** — 16:23–16:43Z, and again 22:28–22:48Z after I re-ran the cancelled job — where the same job takes **9m16s on main** (run 31661593478). It is a required check, so #180 sits open and there is no merged tree to re-derive against. **Nothing here is stale while that holds**: 4038 = 3938 + 68 + 32 is the derivation against the tree this would merge into today.

I took everything that does not depend on #180's CI number, so the swap is one substitution when it lands. #180's only RTL change is `MISA_VALUE` `0x4000_1104` → `0x4000_1105`, so its tree can be measured exactly here.

**The band, re-taken on the post-#180 base** (local, base = **3988**):

| `MISA_VALUE` | `fit` (local) | vs. base |
|---|---|---|
| `0x4000_1105` (post-#180 base) | 3988 | — |
| `0x4000_1107` | 3958 | −30 |
| `0x4000_1145` | 3971 | −17 |
| `0x4000_110d` | 3987 | −1 |
| `0x4000_1185` | 3925 | **−63** |
| `0x4000_1905` | 3980 | −8 |

**63 cells** of span on that tree against **68** on `64759da` — the same band seen twice, on two adjacent trees, which is the point of the term. Every probe moves *down* from 3988, so that base sits near the top of its own window; that is an argument for budgeting the whole span rather than only the upward excursion.

So when #180 lands the change here is: `FIT_MAX_LC := <merged fit job's count> + 68 + 32`, `FIT_LAST_LC := <that count>`, and the comment's three lines re-pointed at the merged SHA and run id, with the 63/68 pair quoted as the band evidence. I will take the job's figure off the merged run rather than predicting it — the local instrument reads 3988 there, but the local/CI gap is exactly the thing this PR records as having no fixed sign.

Holding here rather than guessing at a number that does not exist yet.

thejefflarson:
Rebased again onto `4c1958c` (#184, docs only — four ADRs and a `docs/adr/README.md` row). **No conflict**: #184 touched nothing in `CLAUDE.md`, so the band and instrument-gap amendment applies unchanged and the "keep both" case never arose.

**The derivation is unaffected and that is measured, not assumed.** #184 changes no input to `make fit`, and main's own run on `4c1958c` ([31977544440](https://github.com/thejefflarson/little-cpu/actions/runs/31977544440)) reads **3966** — the same figure as `fe618f4`'s run 31974553338. Two CI runs on two trees, one number. The in-repo citations stay pointed at `fe618f4`, where both halves of the instrument pair were actually taken (job 3966, local 3969), per the convention of quoting a number with the tree it was measured on.

`FIT_MAX_LC := 4088 = 3966 + 68 + 54` and `FIT_LAST_LC := 3966` stand. Head is now `c33b1c4`; local `make fit` reads 3969 of 4088 with `TREND +3`, and `make probe-gates` is 254/254.

PR #188, opened 2026-08-16
Correct the stale ISA sites, and guard the name (JEF-838)

thejefflarson:
Closes JEF-838.

## The grep, against the ticket's list

`git grep -in rv32imc` over the tracked tree, then split by case. The ticket named three live sites; the grep found **six**, and two of those were fixed on `main` by #185 while this was in flight.

| Site | Ticket named it | Outcome |
|---|---|---|
| `docs/THREAT_MODEL.md:13` | yes | corrected |
| `Makefile:968` (`compare-timing` banner) | yes | corrected |
| `soc/compare/dhry_dmips.py:190` | yes | corrected |
| `test/asm/fencenop.S:4` | **no** | corrected |
| `test/sail/rv32imac_zicsr.json:1` | **no** | fixed on main by #185; the atomics claim thirteen lines below it was not, and is corrected here |
| `test/COSIM_EXPECTED_FAIL:48` | **no** | fixed on main by #185 |

`fencenop.S` needed a different fix from the rest. Its sentence sat inside a paragraph about riscv-formal's coverage, so naming the wider ISA there would have made the next sentence false. Both now say what they mean: the four instructions are named against what the core implements, and the set-equality is named against what `isa_rv32imc.txt` names.

Everything else the grep returned is `docs/adr/`, `docs/ideas/`, or the lower-case spellings the formal flow needs — see below.

## The term-matching decision

**Case-sensitive, uppercase, substring.** Stated at the term in `test/retired_term_test.sh`, with what it trades away.

The retired thing is the prose name of an ISA, which is written in capitals. The lower-case `rv32imc` is a live and correct identifier: `formal/checks.cfg`'s `isa rv32imc`, the Makefile's `MONITOR_GEN -i rv32imc`, riscv-formal's `insns/isa_rv32imc.txt`, and the `rvfi_isa_rv32imc` module generated from it. The pinned clone models no A encoding, so widening any of them generates nothing.

Measured, not asserted: flipping the term to `any-case` against this tree lands on **19 live sites** across `formal/checks.cfg`, `formal/complete.sby`, `formal/complete.sv`, `formal/complete_cover.sby`, `formal/check-complete-exclusions.py`, `formal/EXPECTED_CHECKS`, `formal/COMPLETE_EXCLUSIONS`, the Makefile and `CLAUDE.md`. The only route back to green would be a pattern the file quietly skips — the exact shape of exemption this file was written to refuse. Case separates the two on the property that actually tells them apart: one is prose, the others are arguments to a tool.

The reason lives at the term declaration rather than in a skip, and `formal/checks.cfg:6` already carries its own. `test/march_test.sh` holds the other side: it *requires* those two live spellings to stay. Residual, stated rather than discovered later: a lower-case spelling written as prose is not caught here.

`RV32IMAC` does not contain `RV32IMC`, so the replacement name is not self-matching.

## The guard

`test/retired_term_test.sh` was deliberately single-term, and said a second one "would be a rename plus a loop, not a config line." Honoured: it is a loop over a term table now. Each term carries its own allow-list, its own diagnostic and its own case sensitivity, and the both-ways set equality runs **per term** — so `docs/adr/` losing one word is red for that word alone even while it still carries the other.

`docs/adr/` and `docs/ideas/` are allow-listed **by name with the reason**, never edited: ADR-0002 *is* the decision that the target was the older ISA and ADR-0108 records moving off it. Rewriting either would make the history agree with a later decision.

Three ways the table itself can be left unreadable now stop rather than scanning less than they claim: an unknown casing keyword, a term with no allow-list, an empty table. A fourth — a term with a list but no diagnostic — stops too; it is the one new failure path not probed, because reaching it needs a fixture edit no single `sed` expresses.

## Tests

Seven probes in `test/probe_gates.sh`, `PROBES_EXPECTED` 251 → 258. The value was read live off `main` at rebase time; the ticket warned it is a moving target, so re-read it if this rebases again.

- the stale name coming back in a formal harness comment is red, and located
- the diagnostic names the ISA the core does claim (`Write RV32IMAC`)
- **control**: the formal flow's lower-case `rv32imc` is a different thing and is not caught
- an allow-list entry whose site lost the stale name is red
- the three table-integrity stops

The fixture's `docs/ideas/` copy moved to `finish-the-rewrite.md`, the one brief carrying both terms, so a single copy covers that entry on either list.

Every new probe pins both exit status and a diagnostic fragment, and the case decision's red direction was demonstrated by hand against the real tree before it shipped.

## Checks

| | |
|---|---|
| `make test` | 70/70, matches `test/EXPECTED_FAIL` exactly |
| `make test-units` | rc=0 |
| `make probe-gates` | 258 graded comparisons, every failure path executed |
| `make retired-term-test` | 2 terms confined, 339 tracked files |
| `make march-test` | 6 sites, 7 named exceptions |

Prose plus the guard: no behaviour change, no config key moved, no baseline weakened, no ADR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -972,7 +972,7 @@ compare-timing: compare.$(COMPARE_CORE).asc compare.$(COMPARE_CORE).core.log
 	@echo
 	@echo 'THIS IS NOT `make soc-timing`, AND NOT A LIKE-FOR-LIKE COMPARISON.'
 	@echo 'Different part, smaller memories, no timer, and the two cores'
-	@echo 'implement different ISAs: RV32IMC+Zicsr with traps here, RV32IC with'
+	@echo 'implement different ISAs: RV32IMAC+Zicsr with traps here, RV32IC with'
 	@echo 'no CSR file and no traps there. Read ADR-0086 before quoting any of'
 	@echo 'it. One placement is a sample: soc/compare/sweep.sh runs four each.'
 	@python3 soc/timing_split.py compare.$(COMPARE_CORE).timing.rpt
--- a/docs/THREAT_MODEL.md
+++ b/docs/THREAT_MODEL.md
@@ -10,7 +10,7 @@ what is at risk, so a reviewer spends its effort on the things that can really g
 
 ## What this is
 
-A hobby RV32IMC_Zicsr CPU core in SystemVerilog, plus the verification harness that checks it.
+A hobby RV32IMAC_Zicsr CPU core in SystemVerilog, plus the verification harness that checks it.
 It builds and runs in exactly two places:
 
 | | |
--- a/soc/compare/dhry_dmips.py
+++ b/soc/compare/dhry_dmips.py
@@ -187,7 +187,7 @@ def main():
             "THE DMIPS COLUMN MULTIPLIES A CLOCK MEASURED AT THE PLACED 4 KB/2 KB\n"
             "GEOMETRY BY CYCLES MEASURED AT A LARGER SIMULATED ONE, because no ice40\n"
             "in this flow has the block RAM to hold Dhrystone. It is a projection.\n"
-            "Neither side is a shipped design either: this core is RV32IMC + Zicsr\n"
+            "Neither side is a shipped design either: this core is RV32IMAC + Zicsr\n"
             "with traps, that one is RV32IC with a branch predictor, no CSR file and\n"
             "no traps, and the image is the ISA they share."
         )
--- a/test/asm/fencenop.S
+++ b/test/asm/fencenop.S
@@ -1,11 +1,11 @@
 #include "riscv_test.h"
 #include "test_macros.h"
 
-// fence, fence.i, wfi and c.nop: the RV32IMC instructions with no riscv-formal
-// spec model at the pin AND no other test.
+// fence, fence.i, wfi and c.nop: the instructions this core implements with no
+// riscv-formal spec model at the pin AND no other test.
 //
 // formal/EXPECTED_CHECKS is set-equal to riscv-formal's own
-// insns/isa_rv32imc.txt, so every other instruction in the ISA has a generated
+// insns/isa_rv32imc.txt, so every instruction that file names has a generated
 // `insn_*` check. The pin models none of `c_nop`, `c_ebreak`, `fence`, `ecall`,
 // `ebreak` -- and that five-instruction blind spot is where the c.ebreak trap
 // cause defect lived. `ecall` and `ebreak` are covered by trap.S,
--- a/test/probe_gates.sh
+++ b/test/probe_gates.sh
@@ -27,7 +27,7 @@ REPO=$(cd "$HERE/.." && pwd)
 # Pinned as a literal: a probe that is deleted, or that stops being reached by
 # an early `return`, would otherwise cut this file's coverage while it kept
 # printing a green summary.
-PROBES_EXPECTED=267
+PROBES_EXPECTED=274
 
 tmp=$(mktemp -d "${TMPDIR:-/tmp}/littlecpu-probe.XXXXXX") || {
   echo "error: could not create a temporary directory under ${TMPDIR:-/tmp}." >&2
@@ -1392,7 +1392,9 @@ rn_fixture() {
   # entries covering those two are directories, and a hundred more copies per
   # fixture would buy nothing but wall time.
   cp "$REPO/docs/adr/README.md" "$d/docs/adr/"
-  cp "$REPO/docs/ideas/fit-the-core-on-the-up5k.md" "$d/docs/ideas/"
+  # The one brief carrying BOTH retired terms, so a single copy covers the
+  # docs/ideas/ entry on either list.
+  cp "$REPO/docs/ideas/finish-the-rewrite.md" "$d/docs/ideas/"
   cp "$REPO/test/probe_gates.sh" "$REPO/test/retired_term_test.sh" "$d/test/"
   cp "$REPO/formal/wrapper.v" "$d/formal/"
   # -c init.defaultBranch, so the branch-name advice cannot land in the middle of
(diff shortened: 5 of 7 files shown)


thejefflarson:
## `PROBES_EXPECTED` collision with #186 — resolution verified by running it, not by adding it up

Three branches all cut from `PROBES_EXPECTED=251` move it. Counted off each diff rather than taken on trust (`grep -c '^+probe '`):

| PR | probes added | resulting value on its own | Makefile hunk |
|---|---|---|---|
| #186 | **16** (15 in `soc/baseline_summary.py` + 1 in `soc/baseline_sweep.sh`) | 267 | `@@ -765` (`soc-timing`) |
| #182 | **3** | 254 | `@@ -603` (`fit.json`) |
| #188 (this) | **7** | 258 | `@@ -965` (`compare-timing`) |

**If this merges after #186, the resolution is `274`** — 251 + 16 + 7. If #182 also lands first, `277`.

### Verified rather than computed

I cherry-picked this commit onto #186's head locally, resolved, and ran it:

```
274 graded comparisons, every failure path executed.
```

`make test` 70/70 matching `EXPECTED_FAIL`, `make test-units` rc=0, `make retired-term-test` and `make march-test` green on the integrated tree. The trial branch was discarded; this PR is unchanged.

### The two groups do not interact

The merge produced **exactly one conflicted hunk — the `PROBES_EXPECTED=` line itself.** Both probe groups came through clean, and nothing else needed a decision:

- **Insertion points are ~360 lines apart.** #186's group goes in after `soc/timing_split.py` (~985); this one extends the existing `test/retired_term_test.sh` group (~1345).
- **No shared helpers.** #186 adds `BS`, `bs_sweep`, `bs_fixture`; this PR uses the pre-existing `RN`/`rn_fixture` and adds `RNF`. Both call the shared `new_case`, which mktemps a fresh directory per call, so the fixtures never share a path.
- **Makefile hunks are 200+ lines apart** — three different targets, no overlap.
- **Different fixture substrates,** as expected: theirs are hand-written provenance CSVs, this one is a `git init` fixture over copied tracked files.

### Note for #182, merging last

It rebases onto whatever lands first, so its `254` will be stale on arrival. The sum rule holds: base + every branch's addition. **Re-run `make probe-gates` and take the number it prints** — the tally is graded, so a wrong sum is red rather than silent, but confirming it beats arguing with it.

thejefflarson:
## Rebased onto `06e7204`. The number is **258**, not 274 — merge order went the other way.

My comment above worked out 274 on the assumption this would be the *second* of the two `PROBES_EXPECTED` branches to land. It is not, as of this rebase:

```
$ git log --oneline origin/main -2
06e7204 Take the never-interrogated blocks' deleted-whole ceilings (JEF-833) (#187)
9e26cfd Name the machine the Sail config configures, in both sentences (#185)

$ git show origin/main:test/probe_gates.sh | grep PROBES_EXPECTED=
PROBES_EXPECTED=251
```

#186 and #182 are both still `OPEN`. `main` is still at the 251 base, so **this PR is first and its own 258 is the correct value.** Setting 274 here would have been red on arrival — it would have declared sixteen probes that are not in the tree yet.

**#186 now becomes the branch that sums.** When it lands after this, its resolution is 251 + 7 + 16 = **274** — the same total, just carried by the other branch. #182 still sums onto whatever is there when it rebases last.

### Confirmed by running it, again

```
258 graded comparisons, every failure path executed.
```

Declared value equals reported value. Also green on the rebased tree: `make test` 70/70 matching `EXPECTED_FAIL`, `make test-units` rc=0, `make retired-term-test`, `make march-test`.

### #187 was invisible to this diff, confirmed rather than assumed

It landed under this branch and is docs-only — `CLAUDE.md`, one new ADR, `docs/adr/README.md`. This PR touches none of the three, and the rebase applied **with no conflicts at all**.

Worth one line because it is the interesting case rather than a non-event: the new ADR *does* raise the guard's tracked-file count (339 → 340), and it lands inside `docs/adr/`, which this PR allow-lists for both terms. So a new decision record is free to state the older ISA name in the vocabulary of its day without tripping anything — which is the behaviour the allow-list was written for, exercised by an ADR that arrived on its own.

thejefflarson:
## Retracting the comment above. **#186 landed first, so the number is 274.**

My previous comment said 274 "should not be actioned." **That is now wrong and this supersedes it.** Order flipped after I posted it:

```
$ git log --oneline origin/main -2
c13ed54 Retain per-seed reports, and stamp every sweep with its toolchain and base (JEF-832) (#186)
06e7204 Take the never-interrogated blocks' deleted-whole ceilings (JEF-833) (#187)

$ git show origin/main:test/probe_gates.sh | grep PROBES_EXPECTED=
PROBES_EXPECTED=267
```

#186 merged at `2026-08-17T00:00:58Z`. `main` is at 267, this PR adds 7, so **`PROBES_EXPECTED=274`** — rebased onto `c13ed54` and pushed.

To state the sequence plainly, since three comments now disagree with each other and only the last one counts:

| when | `main` | correct value here | why |
|---|---|---|---|
| at first push | 251 | 258 | nothing had landed |
| after #187 | 251 (docs-only) | 258 | I was still first |
| **now, after #186** | **267** | **274** | #186 landed first; this sums onto it |

The figure is the same one measured on the throwaway branch before the order was known. It was right to refuse it while `main` said 251 and it is right to take it now, for the same reason both times: **the value is a function of what is actually on `main`, never of what is expected to land.**

### Confirmed by running it

```
274 graded comparisons, every failure path executed.
```

Declared equals reported. The rebase produced **the single conflicted hunk the trial predicted** — the `PROBES_EXPECTED=` line, nothing else. Both of #186's groups came through at full size beside this one:

```
soc/baseline_summary.py: 15
soc/baseline_sweep.sh:    1
test/retired_term_test.sh: 16   (9 pre-existing + 7 added here)
```

Also green on the rebased tree: `make test` 70/70 matching `EXPECTED_FAIL`, `make test-units` rc=0, `make retired-term-test`, `make march-test`.

### #182, merging last

It is the only one left and still declares 254 against a 251 base. Its resolution is now **277** — 274 + its 3. The tally is graded, so a stale sum is red rather than silent, but re-running `make probe-gates` and taking the printed number beats computing it.

PR #201, opened 2026-08-18
Build the ECP5 instrument, and date ADR-0110's premise (JEF-847)

thejefflarson:
Closes JEF-847.

Nothing in the repo built for ECP5, so the multi-core milestone's premise — *the same RTL places at 35.33 MHz, 20% logic, 36 of 56 block RAMs* — was the only undated number in a repo whose method is that a decision is a measurement with a date on it. This builds the instrument and dates them.

## What landed

`make ecp5-timing`: `synth_ecp5` on the existing `SOC_SRCS` and the existing `soc-rom` image, then `nextpnr-ecp5` at a declared corner. Same `littlesoc`, no fork, no `ifdef`, no part-specific RTL — the no-fork rule binds. 31 seconds end to end locally.

**The corner is declared and graded, not defaulted.** Colorlight i5, `LFE5U-25F-6BG381C` — Lattice's ordering code for the 25F die, speed grade 6, commercial, caBGA381. Trellis spells the same part `LFE5U-25F-6CABGA381` and writes it into `--textcfg`, which turns out to be the **only artifact of the run that names the corner at all** (nextpnr's log never does), so `soc/ecp5_report.py` grades it off that line. Measured: nextpnr's current default resolves to the same corner, so declaring it changes no number today — what it buys is that a default which moves goes red instead of quietly reporting a corner nobody chose.

**The constraint is a pinned constant above what the design can reach.** `ECP5_TARGET_MHZ := 200.0`, about 6× the achieved figure. nextpnr is timing-driven and stops working a path once the constraint is met, so a 25 MHz target reported as 25 MHz would be measuring the target. The parser goes red if the design ever meets the constant — that is the run where the constant moves, and moving it is a change to the measurement rather than a threshold being relaxed. The reasoning is written where the constant is declared.

**One reader, and it refuses every shape of "nothing was measured."** There is no icetime for ECP5, so nextpnr's own engine both drives the placement and grades it, with no second reader behind it — the ice40 flow already treats that as disqualifying (`soc.asc`'s recipe says nextpnr's exit status is deliberately not the signal). `soc/ecp5_report.py` is red on: a report that does not name the design's clock, an empty `fmax` table, two clocks matching the name, an `fmax` entry that stopped carrying both halves, a constraint other than the declared one, a configuration for another corner or an empty one, a missing or truncated report, no utilisation table, no self-loop critical path, an unclassified hop type, and a walked path that does not reconcile with the frequency published from it. "No constraint found" cannot read as a pass.

**Three exact censuses gate the mapping**, through `soc/cell_census.py` (which gains `--gate`/`--declared` so an ECP5 failure names its own flow and variable rather than the ice40 one's — a dangling pointer in the one message a reader gets):

| cell | expected | what a mismatch means |
|---|---|---|
| `DP16KD` | 36 | `rtl/memory.v`'s no-change read port was shaped for SPRAM inference and there is no SPRAM here — a fallback to LUT RAM is what this catches |
| `TRELLIS_DPR16X4` | 32 | the register file's distributed RAM; zero means it fell into flops and soft muxes |
| `MULT18X18D` | 4 | the multiplier is 4 `ICESTORM_DSP` on up5k; in soft logic it would be invisible in a frequency number and enormous in area |

`--textcfg` is written and required non-empty — the cheap "expressible in the target's configuration" check. No `ecppack`, no bitstream, no packer.

**Every run prints what the target refuses to claim**, from the parser rather than the recipe so it cannot be separated from the number: nextpnr's estimate and not icetime; the constraint is an input to the placer; there is no bitstream and no board; one placement is a sample; and an ECP5 number is never merged with or subtracted from an up5k one.

## Do the remembered numbers reproduce?

Substantially yes, and the two that do not are named rather than quietly replaced.

| ADR-0110 remembers | measured here | |
|---|---|---|
| 36 of 56 block RAMs | **36 of 56 `DP16KD`** | exact |
| register file → 32 distributed RAMs, no block RAM | **32 `TRELLIS_DPR16X4`**, no block RAM | exact |
| 20% logic utilisation | **21.9%** (`TRELLIS_COMB` 5331/24288) | close, not exact |
| 35.33 MHz | **33.65 MHz** at one placement | 4.8% low |

Both gaps have a plausible cause and neither is investigated here, because a single placement is not the instrument for it: the tree is not the tree those numbers came from, the toolchain has moved, and the 3.5% below is one identified variable on its own.

The critical path is the fetch loop, same as up5k: `imem.rom_even` → decode → `imem.rom_odd`'s address flop, 29.72 ns over 49 hops, **60.1% routing**. Routing-dominated, like the other part.

## The clock pin is not cosmetic

`soc/littlesoc.lpf` pins only `clk`, to the i5 module's 25 MHz oscillator on P3. Letting nextpnr choose that pad instead read **34.88 MHz against 33.65** — 3.5%, because the pad decides where the global network is entered. The pessimistic number is the one taken against the real oscillator, the same direction as taking speed grade 6. The button and LEDs belong to whatever baseboard the module is plugged into, so they stay unconstrained rather than invented — `soc/littlesoc.pcf` records why that matters. No `FREQUENCY` line in the LPF, and there is a tripwire comment saying why: it would override the pinned `--freq`.

`soc/littlesoc.pcf` is untouched and stays ice40-only. `synth_ecp5` runs with **no mapper flags** — abc9 is that script's default on this part, so `-noabc9` would be as much of a mapper change as turning abc9 on is on ice40, and a mapper change under a brand-new instrument would confound both.

## CI

New `ecp5-timing` job on PRs and main. It asserts `yosys` and `nextpnr-ecp5` are on PATH **and** that the Trellis device database loads, before reporting any verdict — the same reason the `formal` job asserts `btorsim`. The database check asks nextpnr for the exact corner the Makefile declares (`nextpnr-ecp5 --25k --package CABGA381 --speed 6 --test`, ~4 s) rather than probing wherever a given install puts Trellis, and it reads the corner out of the Makefile rather than writing it down twice.

**Human follow-up, deliberately not done here: making `ecp5-timing` a required check is a branch-protection change**, which is a repository setting rather than a change to `ci.yml`.

## Decisions I made and why

- **No ADR.** The ticket sequences it with the sweep, and one placement is not a number this repo quotes. No Fmax ratchet either — the censuses gate, the frequency publishes.
- **The LPF ships.** The ticket left it optional ("costs nothing and removes a variable"); it costs 3.5% and removes a real variable, so it ships with that measurement written into its header.
- **`CLAUDE.md` amended**, minimally: `make ecp5-timing` added to the command list, "two instruments, two designs" → three, and one bullet describing what this instrument gates, what it publishes and what class it is. No existing rule weakened. Flagging it because the rulebook is the repo's ground truth and a new gate that is not in it is a gate nobody knows the rules for.
- **`soc/cell_census.py` gained two diagnostic-only flags** rather than being copied. Both default to today's strings, so the ice40 flow's messages are byte-identical.

## Tests

Seventeen new probes in `test/probe_gates.sh` force every one of the parser's red directions; two more pin that an ECP5 census names its own gate and its own declaration in both halves of the diagnostic. `PROBES_EXPECTED` 312 → 331. Every probe fails without the code it grades.

- `make test` — 70/70, `PROBES_EXPECTED=331`, 331 graded comparisons all green
- `make probe-gates` — green, including the 19 new ones
- `make soc-timing` — green, 12.71 MHz against the 12.00 floor, `SB_SPRAM256KA: 2`, `SB_RAM40_4K: 20` (the census flags default to the existing strings, so the ice40 flow is unchanged)
- `make lint` — clean in both passes
- `make ecp5-timing` — green, censuses as declared, 33.65 MHz


thejefflarson:
## [HIGH] `ecp5.config` could grade a stale run — fixed

The finding is correct, and it is this repo's recorded defect shape rather than a theoretical one. Reproduced before fixing: with the previous recipe, a stub `nextpnr-ecp5` that exits 1 without writing anything over a directory holding a good pair printed `Fmax : 40.00 MHz` from the *previous* run and `make ecp5-timing` exited **0**.

Three things had to line up and all three did: `|| true` drops the status, nothing deleted the outputs before re-running, and `test -e` cannot tell a fresh file from last week's. `ecp5.json` depends on the phony `soc-rom`, so yosys and the three censuses re-ran every invocation while the frequency being printed belonged to RTL no longer in the tree — and `ecp5.report.json` is not a make target at all, so `.DELETE_ON_ERROR` never touched it and nothing else ever invalidated it.

**What changed** (`d6084e7`):

- `@rm -f $@ ecp5.report.json` is now the recipe's first line.
- **`|| true` stays**, as advised. This flow hands nextpnr a constraint the design is *meant* to miss, so `ERROR: Max frequency ... (FAIL at 200.00 MHz)` and exit 1 is what a healthy run looks like here. Removing it would make every normal run red and would let `.DELETE_ON_ERROR` eat the configuration the measurement is read out of. The comment above the recipe now says which part of this is load-bearing and asks the next reader not to weaken `-s` back to `-e`.
- The guard is `@{ test -s $@ && test -s ecp5.report.json; } || { ...diagnostic...; tail -30 ecp5.pnr.log; rm -f $@ ecp5.report.json; exit 1; }`. It deletes the pair on the way out too, so a later run cannot inherit what this one refused.
- `clean` already removed both files; confirmed.

**Five probes**, driving the real recipe rather than a copy of the guard:

| probe | stub behaviour |
|---|---|
| control | writes both files, exits 1 — what the real tool does every run |
| stale pair | exits 1 writing nothing |
| the report goes too | exits 1 writing nothing; graded through a config written by the probe, so the parser gets past its cheapest check and the missing **report** is what speaks — the half `.DELETE_ON_ERROR` structurally cannot do |
| truncated config | `: > ecp5.config`, exits 1 — the `-e`-versus-`-s` case |
| half-written pair | writes the config but no report |

Hermetic via `make -o soc-rom`: `ecp5.json` hangs off that phony target, so without it make would rebuild the netlist and want yosys and a cross compiler for a file the stub never reads. Marking it old leaves `ecp5.json` up to date and `ecp5.config` genuinely out of date, which is the state a second run on a laptop is really in.

**Demonstrated red direction:** against the previous recipe the control stays green and all four go RED. That is in the commit message rather than only here.

`PROBES_EXPECTED` 317 → 341 (317 is main's after #200; 24 are this branch's).

## Also in this push

- **Rebased onto `d6a58b3`.** One conflict, `PROBES_EXPECTED`, textual as expected.
- **The ECP5 flow now takes a toolchain stamp**, following #200's convention: `ECP5_TOOLS := yosys nextpnr-ecp5`, an `ecp5-timing-toolchain` prerequisite so a synthesis that dies or a census that trips still records what was running, and `print-toolchain`'s default list gains it. It matters more here than for either up5k flow — this instrument is new, nothing about it is pinned, and its numbers have no derived band yet.

## Gates, re-run on the fixed tree

- `make test` — 70/70, 341 graded comparisons, all green
- `make ecp5-timing` from `make clean` — censuses as declared (36 / 32 / 4), **33.65 MHz**, toolchain stamped
- `make soc-timing` — 12.71 MHz against the 12.00 floor, `SB_SPRAM256KA: 2`, `SB_RAM40_4K: 20`
- `make lint` — clean in both passes

Not merging — held for your call.

PR #203, opened 2026-08-18
Build the mapped-netlist digest, and place what it claims (JEF-846)

thejefflarson:
Closes JEF-846.

A sixteen-seed sweep is about twelve minutes a side, and the shipping SoC's worst placement of sixteen clears 12 MHz by a fraction of a nanosecond with area measured twice not to buy any back. So a change that cannot have moved the placer's input should not cost a sweep. This builds the instrument that says which kind of change it is.

## The three measurements reproduced first

On `553a241` with the cached OSS CAD Suite (Yosys 0.68+48 `ff5817c34`, nextpnr-0.11 `62e659ed`):

- synthesis is byte-deterministic run to run — two builds of `littlecpu`, identical SHA-256;
- a comment-only edit moves the mapped JSON in **66 places, all `src`/`module_src`**;
- a dead wire yosys deletes still moves it — a `netnames` entry carrying `unused_bits` stays behind and renumbers every net bit id after it (104 differing lines injected into `rtl/regsel.v`, 1447 into `rtl/decoder.v`);
- `synth_ice40 -dsp; opt_clean -purge; write_json` with `src`/`module_src` dropped makes the dead-wire tree digest **identical** to base.

On the design that actually places, four different shipping netlists — base, a comment, a blank line, a dead tie-off — share **one canonical digest and one bitstream**; a one-bit `misa` change is a different digest, reported as −7 `SB_LUT4`, +3 `SB_CARRY`, −4 cells, +27 nets.

## The determinism control: the specified proxy failed, and the gate survived it

Control A (one netlist placed twice at one seed → byte-identical `.asc`) **passes**, over 7 309 930 bytes.

Control B as specified — place the purged and unpurged forms of one tree and require one bitstream — **fails**: the purged netlist places to `e47243d7…` where the shipping one places to `8da98943…`, same seed, same design, same tools. `opt_clean -purge` moves the placement.

That does **not** narrow the gate to strict equality of the unpurged form, because nobody places the canonical form. The proxy was sufficient, not necessary; the necessary question is whether the differences the digest forgives move the *shipping* placement, and they measurably do not. So control B asks that directly: a copy of the tree carrying all three forgiven edit classes is synthesised in the shipping form, its bitstream must equal the base's, and its digest must stay equal — plus a vacuity check that the mutant's shipping netlist **differs**, or the control passed having injected nothing. The purged-placement finding is recorded as the reason nothing may hand the canonical JSON to nextpnr.

## What landed

- `soc/netlist_digest.py` — the only reader of the canonical JSON. Digests it, and on a mismatch prints the structural difference (cell-type counts, ports, modules) or, where the summary has nothing to say, the differing paths. Missing / empty / truncated / not-a-netlist / no-top / no-cells are refused with exit 2 — never "equal".
- `soc/netlist_determinism.sh` — the control above, a prerequisite of the digest targets the way `pcloop_cover` is of `pcloop`. Three placements, about three minutes.
- `soc/netlist_base.sh` — builds another commit's canonical netlist from `git archive`, using **that tree's own** ROM recipe and synth script where it names one (a flag that moved would otherwise be invisible), and saying so when it names none.
- `make netlist-digest`, `make netlist-diff BASE=<ref>`, `make netlist-determinism`, part-parameterised (`NETLIST_PART`, up5k today; an unknown part is refused rather than digested).
- `SOC_SYNTH` / `SOC_PNR` are now named once in the Makefile and used verbatim by both the shipping flow and the digest, so the digest cannot describe a netlist nothing builds. `make fit` and `make soc-timing` are graded against exactly what they were before.
- The landing procedure where the sweeps are (`soc/baseline_sweep.sh`, `soc/timing_sweep.sh`): digest unchanged, no sweep owed; digest changed, the paired sixteen-seed sweep — and at today's margin a difference is a stop-and-redesign signal, not merely a sweep owed.
- ADR-0120, and a `Commands` entry in CLAUDE.md.

## Tests

36 new hermetic probes in `test/probe_gates.sh` against JSON fixtures and stub yosys/nextpnr — the forgiven classes, the not-forgiven ones (a non-source attribute, a dead net left in the file), the one-bit constant and its named path, all six refusals, and every red direction of the control including the vacuity check. `PROBES_EXPECTED` 312 → 348.

Checks run locally: `make test` 70/70 with the failure list matching `EXPECTED_FAIL` (probe-gates 348/348 inside it), `make fit` 3945 of 4088, `make soc-timing` 12.71 MHz against the 12.0 floor, `make netlist-digest`, `make netlist-diff BASE=origin/main` (equal, as it must be: this PR moves no RTL), and the same run with a one-bit `misa` mutation, which reports DIGEST-DIFFERENT and names the cell delta.

## Decisions taken without asking

- **The digest covers the SoC that places, not `fit`'s core-alone top.** A tie-off dead inside the SoC can be live at a top whose inputs are unconstrained, so an implication for `fit` does not follow from the SoC's digest. One part, one design, one claim.
- **The purge-versus-shipping placement is recorded, not graded.** Requiring it to keep differing would go red for a toolchain improvement.
- **CLAUDE.md gets the `Commands` entry only.** The rulebook paragraph under *Measurements and ratchets* is left for the architect, since the doctrine is stated in ADR-0120 and in both sweep scripts.


thejefflarson:
Held-PR fixes pushed (rebased onto `c38a9a7`, so `-device u` is folded into the single-sourced `SOC_SYNTH`).

**[CRITICAL] `place()` accepted a placement that never happened** — fixed, and the fix is not the one suggested, because the suggested marker was measured wrong here. `ICESTORM_LC:` is printed *before* placement (that is why `make fit` prints a utilisation table on a run that fails on IO), so it does not separate "finished" from "died mid-write". What does is nextpnr's own last line, `Info: Program finished normally.`, printed after the bitstream. The exit status is captured and carried into the diagnostic but cannot be the grader: measured here, a run whose own timing estimate fails exits **1** having written a bitstream **byte-identical** to the same placement made without the requirement, so demanding zero would void the control on exactly the marginal design it serves. The marker is a part-table entry (`NETLIST_PNR_DONE`). New probe `STUB_PNR_TRUNCATED`: non-empty bitstream, no marker → red.

**[HIGH] the vacuity check could not see its class** — fixed, and measuring it turned up something worse than reported: the dead tie-off **was not reaching the netlist at all**. Injected into `rtl/decoder.v`, a dead wire leaves *no trace* in this SoC's mapped netlist — the two netlists differ in `src` attributes and nothing else — so control B was exercising the comment class twice and the dead-net class never. The injection site moved to the **top** module (a part-table entry, `NETLIST_MUTANT`), where the wire costs one `netnames`/`unused_bits` entry (130 → 131) and seven non-attribute lines, and both halves are now asserted by name: present in the shipping netlist, **absent** from the canonical one, so the purge is demonstrably what forgives it. The injector's scope check follows the splice rather than the file. Three new probes, including a signal that is in the file but in an earlier module. ADR-0122's table row and its consequences are corrected — the old row overstated what had been exercised.

**[HIGH] `netlist_base.sh` collected `tail`'s status** — fixed: make's own status decides, the fallback fires only on `No rule to make target`, and any other failure prints the log and exits 2. New probe drives a base tree whose `print-%` rule fails. (Kept `#!/bin/sh` + `set -eu`, which is what `soc/baseline_sweep.sh` and `soc/timing_sweep.sh` use; the pipefail-carrying scripts in this tree are the bash ones. The fix removes the dependence on a pipeline's status rather than relying on an option.)

**[HIGH] the ref reached the shell as text** — fixed: `netlist-diff: export BASE`, and the recipe reads `"$$BASE"`. Verified with a canary — `BASE="x'; touch …; echo '"` is refused by `git rev-parse` as a ref and the canary file is not created.

Gates re-run on the rebased tree: `make test` 70/70 matching `EXPECTED_FAIL`, probe-gates **358/358** (`PROBES_EXPECTED` 317 → 358), `make fit` 3945/4088, `make soc-timing` 13.07 MHz, `make netlist-determinism` PASS with the dead net asserted through, `make netlist-digest`, and `make netlist-diff BASE=origin/main`.

Not merging — leaving that to the architect.

PR #204, opened 2026-08-18
Build the bus arbiter and prove it, with its environment assumption named (JEF-851)

thejefflarson:
Closes JEF-851.

Two harts on one bus need at most one driver per cycle, an AMO's two cycles indivisible, and no unbounded wait. Nothing in the design did any of it, and **starvation is the one property no torture program can catch** — a fixed-priority arbiter passes every one ever written for it and starves the loser anyway. So the module ships with its proof rather than after it.

## What is here

- **`rtl/busarbiter.v`** — a leaf module: free `request` and `mem_lock` inputs, a registered `grant`, and no core state read anywhere.
- **`formal/busarbiter.sv`** + the `busarbiter` task in `components.sby` — `mode prove`, `smtbmc`, **successful proof by k-induction**.
- **`formal/busarbiter_cover.sby`** — the anti-vacuity control, six goals, a prerequisite of the proof target.
- **`formal/busarbiter-probe.py`** — the two forced red directions, also a prerequisite.
- A step in the `components` CI job, an ADR, and 17 probes in `test/probe_gates.sh` grading the probe's own comparisons.

## Two bits of state, not three

The ticket budgeted three bits. The round-robin pointer turns out to be redundant with the grant — a tie goes to whichever hart did not hold the bus last cycle, which `grant` already says — so the whole module is two flops and one mux:

```systemverilog
assign winner = (request == 2'b11) ? (grant[0] ? 2'b10 : 2'b01) : request;
assign held   = |(grant & mem_lock);
always_ff @(posedge clk) grant <= reset ? 2'b00 : (held ? grant : winner);
```

That is not just smaller: **every bit of its state is on a port**, which is what lets the strengthening invariants be written in the harness over the module's outputs. A separate pointer register would be state the harness cannot see, `yosys` does not let a harness reach into an instance for it (`formal/pcloop.sv` records what happens when one tries), and the invariants would have had to live in the RTL's own `FORMAL` block or the pointer be exposed as a port nothing else reads.

**The price is an asymmetric bound, and it is stated rather than hidden.** An idle bus reads `2'b00` and `2'b00`'s tie goes to hart 0, so hart 0's worst wait is **2** cycles and hart 1's is **3**. Both are bounded, which is the property. Each hart's harness carries its own constant, and each constant is *covered* as well as asserted, so a bound loosened by a later edit goes red as an unreachable cover goal rather than passing wide.

## THE ASSUMPTION IS NOT DISCHARGED, and that is the honest state of it

The wait bound is proved under an environment assumption: **no hart raises `mem_lock` on two cycles running.** A lock held forever holds the bus forever and no arbiter can take it back.

**Nothing in this tree discharges that today.** The core publishes no `mem_lock` at all — the output and the duration property that owes this assumption its proof land with the core-surface ticket (JEF-850), which is not built. Until it is, **the bounded-wait result is conditional**: read a PASS as "granted within the bound, *provided* the core really does lock for one cycle". That is said in three places — at the assumption site in `formal/busarbiter.sv`, in the ADR, and here.

What JEF-850 owes this module, precisely: **`mem_lock[i]` high on the cycle an AMO issues** (not the write cycle that follows — the grant is registered, so the arbiter must be told before the cycle it must cover), **low on the write cycle, never two cycles running.**

Bounding the lock inside the arbiter instead — a bit refusing a second extension — would delete the assumption and is deliberately declined: it would cut an AMO's bus tenure short rather than report anything, turning a fairness bug this proof catches into a torn atomic nothing in the machine records.

## The graders, and their forced red directions

`mode prove`, `smtbmc`, both legs pass:

```
summary: engine_0 (smtbmc) returned pass for basecase
summary: engine_0 (smtbmc) returned pass for induction
summary: successful proof by k-induction.
DONE (PASS, rc=0)
```

The cover reaches all six goals (last at step 4):

```
Reached cover statement in step 2 at busarbiter_check: busarbiter.sv:176  (x2 -- both harts granted)
Reached cover statement in step 3 at busarbiter_check: busarbiter.sv:177  (x2 -- a lock spans two cycles while the other hart asks)
Reached cover statement in step 4 at busarbiter_check: busarbiter.sv:179  (x2 -- each hart's bound is reached)
DONE (PASS, rc=0)
```

Both red directions are forced, not argued (`formal/busarbiter-probe.py`, one line of `rtl/busarbiter.v` per mutation):

```
  shipping       prove: PASS, assertions failed at none
  shipping       cover: PASS, goals unreached at none
  fixed-priority prove: FAIL, assertions failed at [130, 155, 157, 159]   <- 159 is the wait bound; the lock arm (123) is NOT red
  grant-mid-lock prove: FAIL, assertions failed at [123]                  <- the lock arm alone
  grant-mid-lock cover: FAIL, goals unreached at [177, 179]               <- 177 is the lock goal
```

The last line is the point the swarm's security review raised: **the anti-vacuity control has its own forced red direction.** The one mutation that makes the cover blind is the same one that makes the lock arm false, so the control cannot quietly stop seeing the class it exists for — and the probe requires the lock goal by *line*, not just a FAIL.

The strengthening invariants are measured rather than believed too: **at `depth 2` the proof closes with them and returns UNKNOWN without either.** The task runs at sby's default depth, which reaches the same verdict by unrolling — so what they buy is the argument written down and a proof whose depth does not grow with it. A third invariant was written and deleted after measuring that it was free at both depths.

## Checks

| check | result |
|---|---|
| `make -C formal components_busarbiter` (cover + probe + proof) | PASS, ~4s |
| `make -C formal components_traps`, `components_decoder` | PASS (the `components.sby` edits touch nothing else) |
| `make test` | 70/70, matches `EXPECTED_FAIL` |
| `./test/probe_gates.sh` | 334 graded comparisons, every failure path executed |
| `make lint` (svlint, both passes) | clean |
| `test/port_connect_test.py`, `test/retired_term_test.sh` | clean |
| `make fit` / `make soc-timing` | not run, and cannot have moved: `rtl/busarbiter.v` is in neither `FIT_SRCS` nor `SOC_SRCS` and nothing instantiates it |

## Decisions taken here, for the record

- **Two bits rather than the ticket's three**, for the reason above. Recorded in the ADR.
- **`grant` is `$onehot0` and not `$onehot`** — the arbiter parks on nobody, so an idle bus has no driver and a hart pays one cycle to be handed a bus nobody was using. Parking on the last owner would buy that cycle back; it is a CPI question, CPI is only measurable in the dual top, and the ticket asked for `$onehot0`. Left for the integrator, flagged in the ADR's consequences.
- **`--keep-going` went on the shipping task, not on the probe.** The probe needs the whole red set (a starving arbiter breaks the handoff at step 3 and the bound at step 6); putting the flag on `components.sby`'s `busarbiter` task means the probe grades the engine that ships instead of a second one of its own. Free on a passing run.
- **No unit bench.** The proof is exhaustive over this module's whole state space and the ticket names the graders exactly; the module gets two frontends anyway (yosys under sby, sv-parser under `make lint`). Stated in the ADR.
- **CLAUDE.md is untouched.** Its formal-commands list would take one line for `components_busarbiter` — left for the architect rather than edited from a ticket branch.

## Conflicts to expect

`test/probe_gates.sh`'s `PROBES_EXPECTED` is 334 here (317 + 17). Sibling PRs move the same literal, and #201 also touches `.github/workflows/ci.yml` and `.gitignore`. The ADR was renumbered **0120 → 0123** on rebase after main claimed 0120 and open PRs claimed 0121 and 0122.


thejefflarson:

PR #211, opened 2026-08-18
Turn mhartid into a HART_ID parameter (JEF-849)

thejefflarson:
Closes JEF-849.

`mhartid` was one address in a five-address read-only-zero arm of `rtl/csrs.v`. It becomes a `HART_ID` parameter defaulted to `0`, reaching `csrs` through `littlecpu`; the other four addresses stay in the shared arm. Hart 0 keeps id 0 because the privileged spec requires some hart on the system to have it, so the default build is the design that was there before.

**Based on #203, not on `main`** — that branch provides `make netlist-diff BASE=<ref>`, which this ticket's acceptance criteria use. A squash-merge of #203 makes the rebase onto `main` the same tree.

## The digest reports, and it says DIFFERENT

This is the first real edit through the new gate, so the verdict is the interesting output. It came out **DIFFERENT**, which the gate's own procedure reads as "spend the seeds". Decomposed with two isolation syntheses:

| tree | SoC `SB_LUT4` | canonical digest |
|---|---|---|
| A — base (`f37b641`) | 4217 | `56a21175…` |
| B — **this tree** (`MHARTID: rdata = HART_ID;`) | 4249 | `808fdf7c…` |
| C — this tree with the arm spelled `32'b0` | 4249 | `808fdf7c…` — **equal to B, byte for byte** |
| D — base plus the arm split only, no parameter anywhere | 4188 | `ac8348dd…` |

**B ≡ C is the load-bearing row.** `HART_ID` at its default folds to exactly the constant it replaced: the tree that ships and the tree that hardcodes zero produce the same canonical netlist and the same SHA-256. So none of the +32 is the parameter's value, and none of it is logic.

What moved is text shape, in two pieces that partly cancel: splitting the arm is −29 on its own (D against A), and the parameter *derivation* — yosys deriving a `$paramod` for an override that resolves to the module's own default — is +61 (B against D). The two tops disagree in sign for the same edit, `littlecpu` −62 `SB_LUT4` and `littlesoc` +32, which is this repo's recorded signature for ABC re-mapping rather than for a design change.

**A finding about the gate, not about this ticket:** a parameter override that resolves to the default is a change class the canonical form does not forgive, and one that provably cannot alter behaviour — the digest proves it here itself, since B and C are the same netlist. Dropping `src`/`module_src` does not neutralise `$paramod` derivation, and arguably should not: the reordering really does reach ABC. Worth an ADR line rather than a code change. There is no cheaper spelling available either — threading a parameter *is* an override, so the churn is inherent to the feature as specified. Making it a port instead would avoid the derivation and was not taken: the ticket specifies a parameter, and a port would move `test/port_connect_test.py` and every harness.

## The sweep the digest owes

Paired sixteen seeds, one tree apart, same toolchain and program (`soc/baseline_summary.py … --allow-mismatch`, the `base` field differing being the point of a before/after):

| | worst | median | best | spread | packed LC |
|---|---|---|---|---|---|
| before `f37b641` | 80.00 ns — **12.50 MHz** | 75.63 ns — 13.22 MHz | 74.23 ns — 13.47 MHz | 7.8% | 4658 |
| after this tree | 80.64 ns — **12.40 MHz** | 77.97 ns — 12.83 MHz | 75.85 ns — 13.18 MHz | 6.3% | 4691 |
| delta | +0.8% | +3.1% | +2.2% | | +33 |

**Sixteen of sixteen placements hold 12 MHz**, the worst at 12.40 with 3.3% of margin. Every column is inside the ~3.6% edit-churn band, so by this repo's own rule the median is a null in both directions — and it sits near the top of that band rather than in the middle, which is worth reading with the row above it: the netlist that produced these placements is the same netlist a hardcoded `32'b0` produces, so the +3.1% cannot be a property of `HART_ID`.

## The mutation is the whole grader, and it was forced red

A wrong `mhartid` has no other consequence — nothing in the datapath reads it, and at the default the mutation changes no bit anywhere — so `test/MUTATION_DETECTORS` says so in the entry rather than leaving the reader to assume coverage. Forced, not asserted:

```
== mhartid-ignores-hart-id
   edits rtl/csrs.v
   ok   caught by bench csr_tb

MISMATCH mhartid reads HART_ID: got=00000000 expected=80000001
MISMATCH ...and a non-zero HART_ID survives one too: got=00000000 expected=80000001
```

`make mutation-check` in full: 10 mutations, 20 pairings, each caught by exactly the detectors it is paired with — so the new one is graded in both directions, and the whole `.S` suite staying green under it is a measurement rather than a gap.

`test/csr_tb.v` holds a second `csrs` at `HART_ID = 0x8000_0001` sharing every input with the DUT, so the parameter is the only thing that can differ between the two reads. Top bit and bottom bit are both set, so a value that arrived truncated does not read as this one. Three vectors: the read, the ignored write, and `mimpid` beside it still reading zero — the last is what says the split left the four neighbours in the shared arm. The task compares the value only; `implemented` has no `HART_ID` in its cone, so checking it there could not fail for a reason the DUT's own read at the same address has not already covered.

## Checks

- `make test` — green. 70/70 programs, failure list matches `test/EXPECTED_FAIL`, suite shape matches `test/OBSERVED_FLOOR`, 358 graded comparisons with every failure path executed. `PROBES_EXPECTED` is unchanged at 358 — this adds a mutation, not a probe.
- `make test-units` — all benches green, `csr_tb` included.
- `make mutation-check` — 10 mutations, 20 pairings, exact set match.
- `make fit` — 3883 LC against the base tree's 3945 measured locally on the same toolchain, −62 and inside the band; ratchet 3883 of 4088.
- `make netlist-diff BASE=f37b641` — DIFFERENT, decomposed above. `netlist-determinism` PASS as its prerequisite.
- Paired sixteen-seed `soc/baseline_sweep.sh` — above; `SOC_MIN_MHZ` met at every placement on both sides.

Not run locally, left to CI: the riscv-formal checks and Sail co-simulation. `mhartid` is not on `formal/checks.cfg`'s `[csrs]` list, the generated CSR model treats it as `const` rather than as zero, and Sail's config is single-hart at 0 — so no verdict on either leg should move, and the netlist being identical to the hardcoded-zero spelling is the stronger form of that argument.

## One residual, named

The `littlecpu` → `csrs` pass-through is graded by elaboration only. A connection wrongly hardcoded to `32'd0` there would elaborate and be invisible, because no test instantiates `littlecpu` at a non-zero `HART_ID` and none can without diverging from Sail and from `test/asm/csr.S`. Closing it needs the second hart that motivates the parameter, so it is recorded here rather than papered over with a check that cannot fail.


thejefflarson:

PR #212, opened 2026-08-18
Add the dual-window instruction memory and the timer's NHARTS (JEF-852)

thejefflarson:
Closes JEF-852.

Step 4 of the multi-core brief: the shared-storage platform pieces, both tied off
to today's design at their defaults.

**Based on `thejefflarson/jef-846-...` (#203), not on `main`**, because the
acceptance criteria are graded with `make netlist-diff`. Rebase onto `main` when
#203 merges.

## What changed

`rtl/imemory.v` takes `NHARTS` and publishes one fetch window per hart — address,
word pair, stall bit and fault bit packed low window first, so at one hart every
width is the scalar it has always been and no integrator changes. A **load** is
answered out of window 0's copy and steals that window alone; a **write** stalls
every window, because both parts define a same-address write-during-read as
returning *invalid* data on the reading port, and Zifencei permits fetching the
old or the new instruction but not garbage. No new stall reason and no new arm.

`rtl/timer.v` takes `NHARTS` and gives each hart an `mtimecmp` and an `mtip`
against the one `mtime` the privileged spec mandates. The window is two words
plus two per hart, rounded up to a power of two — four at one hart, eight at two,
top two reading zero — and the elaboration alignment check moves with it.

## The timer/UART collision, resolved

The UART (#206) is not in this base, so nothing collides today — but at
`NHARTS = 2` the timer's window would run to `0x0002_0020` and swallow the UART's
base at `0x0002_0010`. Resolved by **reserving the timer's widest window, 32
bytes, at every `NHARTS`**, rather than by moving the timer (whose base is pinned
by abutment with the top of the data RAM and copied into four other files).

`test/memmap_test.sh` now reads *every* `BASE` parameter declared under `rtl/` and
refuses one inside the span, naming `0x0002_0020` as the address to move to. That
is a list nothing has to maintain, and it is the check that will go red when #206
merges. **The UART's base has to move to `0x0002_0020`** — a one-line change in
`rtl/uart.v`, `rtl/littlecpu.v`, `formal/traps.sv` and `test/asm/riscv_test.h`,
which whoever merges second owns. Without the reservation the UART would work
perfectly on the single-hart machine and the collision would surface only when the
dual top was built, which is exactly the quiet failure the reservation exists to
bring forward. Both directions are probe-gated: a device inside the span is red, a
device above it is green.

The 8-word layout is deliberately **not** a CLINT's, and firmware written against
it does not port to one. Recorded as a choice with its cost in the ADR and in
`rtl/timer.v`'s header.

## Censuses

`rtl/imemory.v` alone, 2048 words:

| | one window | two windows |
|---|---|---|
| ice40 `SB_RAM40_4K` | 16 | **32** |
| ECP5 `DP16KD` | 4 | **8** |
| ECP5 `LUT4` / `TRELLIS_FF` | 352 / 96 | 517 / 108 |

The one-window ECP5 row reproduces the brief's remembered probe exactly. **32
against the up5k's 30 block RAMs is why the dual configuration can never be built
on that part** — a measurement now, not an occupancy argument.

`SOC_EXPECT_EBR` stays **20** and `SOC_EXPECT_SPRAM` stays **2**, both re-measured
on the SoC's synthesis log.

## "Both copies change on one edge", forced rather than argued

That claim is about the **mapped** netlist and about nothing in the source: at RTL
there is one array and every window reads it by construction, so no simulation of
`rtl/imemory.v` can confirm the replication or fail on its absence. New
`test/rom_replication.py` groups the mapped block RAM cells by the nets on their
write port — 16 groups of 2 on ice40, 4 of 2 on ECP5, every copy on the same
enable, address, data and edge, and the cells in a group differing on their read
port so they are copies serving different windows rather than one counted twice.

`test/imem_share_test.sh` runs it on both parts and **forces all three of its
verdicts red**: a mutant whose second window reads a pair of banks the write does
not reach, a declared count that does not match, and a single-copy run that would
pass without looking. It checks that each of the mutation's four edits applied, so
the red direction cannot quietly synthesise the shipping design, and it pins its
own case count the way `probe_gates.sh` pins its probe count. On `make test`.

## The digest verdict — DIGEST-DIFFERENT, and the architect should read this

**Bit-identity is not reachable here**, and that is a finding rather than a
shortfall: a width-one *vector* port cannot digest equal to a scalar one, because
the canonical JSON records the declaration. Every spelling tried leaves either a
`single_bit_vector` attribute or the net keeping the submodule's name.

What *was* avoidable was measured away one variable at a time against the base
tree's canonical netlist:

| edit, at `NHARTS = 1` | SoC `SB_LUT4` |
|---|---|
| the fetch window folded into a loop that includes window 0 | **+30** |
| naming `\|mem_wstrb && text_range` once and using it twice | **+64** |
| the timer's read mux written once, hart 0 an arm of it | **+19 to +25** |
| an in-process `for` loop over the harts, either half | **+36 / −20** |
| **an inert generate loop — zero iterations, nothing elaborated** | **−18** |
| the port widths, the `NHARTS` parameter, the bit selects | **0** |
| the read mux inside a *taken* `generate if` arm | **0** |

So window 0 and hart 0 are spelled exactly as they ship, and everything above the
first lives in a generate the single-hart build does not take. What remains is
**−18 `SB_LUT4`, 4217 → 4199** — and the fifth row is the whole of it: an inert
generate loop that elaborates to nothing moves the same count by the same −18.
Same modules, same ports, same `SB_CARRY` 688, `SB_DFFESR` 646, `SB_DFFSR` 234,
`SB_DFF` 166, `SB_RAM40_4K` 20, `SB_SPRAM256KA` 2.

**DECISION NEEDED.** The repo's landing procedure says a moved digest owes the
paired sixteen-seed sweep, and I could not spend it: `nextpnr-ice40` on this
machine dies in dyld against a boost it was not built for, so `make fit`,
`make soc-timing`, `make netlist-determinism` and `make netlist-diff` cannot run
locally. The canonical-netlist comparison above was taken by hand with the
Makefile's own synth script, which is the digest's content but not its gate. CI's
`fit` and `soc-timing` jobs are the measurement; the argument for the delta being
churn rather than cost is the inert-probe row, not a substitute for the sweep.

## Tests

- `test/imem_tb.v` gains a second instance at `NHARTS = 2`: each window answering
  about its own address on one cycle, a per-window fault, a load stealing window 0
  alone while window 1 keeps fetching *both* its words, **a write raising
  `fetch_stall` on both windows**, and the written word fetched back from both.
  Two of those were forced red by hand against mutated RTL before being kept.
- `test/timer_tb.v` gains a second instance at `NHARTS = 2`: `mtimecmp` per hart
  read back and not disturbing each other, `mtime` shared and still one tick per
  cycle, each `mtip` posted and lowered on its own, the two reserved words reading
  zero and swallowing stores, and the word past the widened window not aliasing
  `mtime`.
- `make window-test` gains the pair that says the alignment check moved with
  `NHARTS` — **one base, accepted at one hart and refused at two**, in both
  frontends. That pair rather than the diagnostic text, because iverilog's
  elaboration tasks take one string literal and the number cannot be in it.
- `make probe-gates`: three new probes, `PROBES_EXPECTED` 358 → 361.

`make test` 70/70 green, `make lint` clean in both passes, `make elaborate-strict`
with no promoted warnings. `formal/` is untouched: the core did not change, and
`formal/arbiter.v` transcribes window 0's behaviour, which did not move either.

## Scope notes

- ADR **0124**. 0120–0123 are claimed by merged or open PRs; renumber if another
  lands first.
- The commit message carries no ticket id, per the repo's rule.


thejefflarson:

PR #213, opened 2026-08-18
Add the core's tied-off multi-hart surface, and settle the seventh stall reason's arm (JEF-850)

thejefflarson:
Closes JEF-850.

**Stacked on #203** (`thejefflarson/jef-846-…`), which provides `make netlist-diff`. Review the two commits on top of it; rebase onto main once #203 merges.

## What lands

Three ports the second hart needs and one stall reason, all tied off, nothing dual built.

| | |
|---|---|
| `bus_wait` (in) | the shared bus is somebody else's this cycle → the seventh stall reason |
| `snoop_write`/`snoop_addr` (in) | another master's write, ORed into the reservation's clear term |
| `mem_lock` (out) | high on an AMO's read cycle, low on its write cycle, so a grant can span both |

## The arm is settled by construction, not inherited

The wait **BUBBLES**. `rtl/executor.v` publishes `stalled` as an *output* and takes no input that freezes it, so a divider stall is the executor being busy in its own FSM rather than a hold the decoder can borrow. A bus-grant wait leaves the executor idle, so a held `decoder_out` is consumed again — one instruction retiring twice, which `rtl/decoder.v` already warns about where it zeroes `out`. The HOLD spelling would have needed a freeze input on the module that is the named oracle for real mul/div arithmetic; that module is untouched here. A wait coinciding with a divide still HOLDS, and both halves are asserted in the decoder's `FORMAL` block and vectored both ways round in `test/decoder_tb.v` — which drives the input directly, because with it tied off nothing single-hart can otherwise tell which answer shipped.

## The six declaration sites

1. `rtl/decoder.v` — the `bus_wait` port, the `stall` OR, the publish bubble arm, and two `FORMAL` asserts (hold-with-divide, bubble-alone)
2. `test/decoder_tb.v` — the OR-identity check (now eight terms, seven reasons) and both-ways vectors, plus one that an interrupt waits for the grant
3. `test/cxxrtl.cc` — the `bus` bucket
4. `test/stall_report.py` — `REASONS` and `HEADINGS`
5. `formal/pcloop.sv` — `f_may_stall`, plus an assertion that an ungranted cycle holds the pc (an excuse that asserts nothing is not one)
6. `CLAUDE.md` commitment 8

`make cycles` reports the column at zero with `unattributed` still zero.

## The arbiter's contract, discharged

`#204`'s bounded-wait proof assumes no hart raises `mem_lock` on two cycles running, and nothing discharged it. `components_accessor` now does, with `assert(take_amo == past_mem_lock)` — both halves at once: every locked cycle is followed by the write-back it bought, and every write-back had a lock in front of it — plus `assert(!(mem_lock && past_mem_lock))`, the assumption verbatim. The snoop-clear property is asserted with `snoop_write`/`snoop_addr` free inputs.

## Tie-off baseline

`formal/MULTIHART_TIE_OFF` + `formal/check-multihart-tie-off.py`, on the `INTERRUPT_TIE_OFF` model and red in both directions: a harness with no line, a line with no harness, a declared harness that stops tying off, a `PORT` line naming a port `rtl/littlecpu.v` does not have, and — the direction that rots — **an input every harness holds at a constant that no baseline declares**. `irq_timer` is an `ELSEWHERE` record so that sweep has something to subtract. The Verilog is parsed by `test/port_connect_test.py` rather than by a second parser. Nine probes in `test/probe_gates.sh` (367 total). The baseline states the depth consequence: the table describes the single-hart configuration, the dual one has no generated checks, and G grows in it.

## Measurements

* **F = 6, G = 6** — `make -C formal remeasure-fg`, both flip points, both trigs, exactly as declared.
* **`make netlist-diff` is DIFFERENT**, and the cause is measured rather than assumed. The two tied-off **inputs** fold before mapping: the coarse cell census is identical, cell for cell. The unread **output** does not — a one-line probe adding *any* unread output port to `rtl/littlecpu.v` moves the SoC **+44 `SB_LUT4`** on its own. So the seeds were owed and spent instead of the digest.
* **Paired sixteen-seed sweep** (base `f37b641` vs `c4a685d`, one toolchain, `soc/baseline_sweep.sh`):

  | | worst | median | best | packed LC |
  |---|---|---|---|---|
  | before | 80.00 ns / **12.50 MHz** | 75.63 ns | 74.23 ns | 4658 |
  | after | 79.87 ns / **12.52 MHz** | 76.75 ns | 74.56 ns | 4706 |
  | delta | **−0.2%** | +1.5% | +0.4% | +48 |

  All thirty-two placements clear 12 MHz; the worst-of-sixteen margin is unmoved. Median is inside the ~3.6% churn band. `--allow-mismatch` is needed to subtract, since a before/after necessarily names two commits.
* `make fit` **3931** of a 4088 budget (−35 against the recorded 3966 — the same change is −35 on the core's top and +48 on the SoC's, which is the documented top-dependence).
* `SOC_EXPECT_SPRAM` 2 and `SOC_EXPECT_EBR` 20, both exact.

## Everything else run

70/70 suite; `probe-gates` 367 with every failure path executed; 86/86 generated riscv-formal checks against `EXPECTED_FAIL`/`EXPECTED_CHECKS`; `components_{decoder,executor,accessor,pcloop,traps}`; `complete`, `complete_cover`, `dmemcheck`, `imemcheck`, `cover`, `nonperturbation`; both tie-off checkers; `port_connect_test.py` green at all 8 `littlecpu` sites (the ticket said ten — 8 is what the tool counts, and the other two files instantiate `bench_littlecpu`); iverilog elaboration and its graded run; `svlint` clean in both passes; `make elaborate-strict` with no warnings.

Red directions demonstrated by hand for each new grader: `bus_wait` out of the OR (5 mismatches), out of the bubble arm (1), moved into the hold arm (1), the snoop clear deleted (4), `mem_lock` tied low (9), and a lock held two cycles (9).

## Decisions taken, for the record

* **No request output.** The ticket names three things and the arbiter's `request` is a free input in its own proof, so nothing here is left owing. The dual top still needs a decode-level request flag; that is step 5's, and it is the one loose end.
* **The snoop clear outranks the `lr.w` arm** rather than joining the clear term. A store-conditional that fails spuriously is permitted anywhere; one that succeeds over another master's write is not. Under the arbiter the two cannot coincide anyway — only one hart is granted.
* **`mem_lock` is not in the tie-off baseline.** An unread output cannot weaken a check; `components_accessor` is what says it is right.


thejefflarson:

PR #214, opened 2026-08-18
Raise causes 5 and 7 for a plain load or store (JEF-855)

thejefflarson:
Closes the core's last conformance gap: a plain load or store to an address no memory answers now raises **load access fault (cause 5)** or **store/AMO access fault (cause 7)**, committed in decode with every other cause.

**Read the measurement before the feature. It does not meet the board clock.**

## The re-take, and it is the headline

The 12.04 MHz worst-of-sixteen this branch shipped with **does not reproduce**. Sixteen placements a side, paired by seed, one tree, one toolchain, against a one-line control (`assign ls_fault = 1'b0;`, which deletes the whole region cone and nothing else):

| | worst | median | best | under 12.00 MHz | packed `ICESTORM_LC` |
|---|---|---|---|---|---|
| control | 12.63 | 12.93 | 13.36 | 0 of 16 | 4692 |
| with the fault | 11.52 | **11.97** | 12.28 | **9 of 16** | 4787 |

**Median of the per-seed deltas: +9.40% of period. 16 of 16 seeds slower, two-sided sign test p = 3.05e-05. +95 cells.**

The **median** is under the requirement, not just the tail. That is far outside this part's churn band (~3.6%) and its placement spread (4–9%), and the sign test carries it without leaning on either. It also reproduces ADR-0116's bit-12 figure of +9.10% on a different tree — so the earlier 12.04 was a sample, not a faster tree, and "the tree got faster underneath it" is not what happened.

**CI's own `soc-timing` job agrees: 11.87 MHz, red.** So this is not a finding that needs the sweep to be believed — the gate on this PR reports it. The sweep is what says how much of it is the feature (all of it: 16 of 16 seeds, p = 3.05e-05) rather than a placement. Note the toolchain gap while reading the two: the same default seed reads 12.06 MHz locally and 11.87 on the runner, which is the `fit`-job argument applied to the other instrument — quote CI's number.

`SOC_MIN_MHZ` has not moved and must not. `docs/adr/0126` records the measurement and names the three ways out, none of them measured. **The merge decision is the architect's; the RTL is complete and correct, and the period is the open question.**

## What was wrong with the branch, and what fixed it

**`uart.S` trapped to zero, and the UART's address was not the reason.** The UART did have to move — `0x0002_0010` sits inside the eight words the base branch's map reserves for one `mtimecmp` per hart, so it is at `0x0002_0020` now at every site that states the map. But `uart.S` still trapped after that: its case 9 stores to `8(t0)`, the word past the device's two registers, on purpose — "a device that transmitted on any store to its neighbourhood would pass everything above". That store used to be dropped silently and now raises cause 7. The case reads the fault instead, and keeps its control: the status register after it must still read idle.

**The monitor gating was owed, and the shape it was in was the bug.** `test/testbench.v` hid a refused access's whole retire from the monitor. That leaves a hole in `rvfi_order`, the reorder buffer's cursor stops at the missing number, and every retire after it is graded against the wrong shadow — which is what `loadfault.S` and `storefault.S` were really reporting, not the trap comparison. The retire is shown to the monitor now and `rvfi_mem_fault` goes with it. `test/sanitize_monitor.py` gains three rules — the port, the channel wire, and the gate on error 101 — plus a compensating check that a retire carrying that flag must also report a trap, so the gate cannot be spent for nothing. Both are vectored in `test/monitor_tb.v` and probed in `test/probe_gates.sh`.

**The generated checks were red for a contract this core had never had to satisfy.** `checks/rvfi_insn_check.sv` has a memory-fault contract, and the eleven A encodings have no spec model to reach it, so it had never once been evaluated here. The twelve plain load and store encodings do reach it: all twelve went red because a refused access retired with `rvfi_mem_addr` zero (it never gets as far as `rtl/accessor.v`) and a write mask of four ones whatever the width was. Decode carries both down the shadow now, and the check compares the address against the spec model's own effective address and the strobe against its mask. **That is a real oracle for the new causes' arithmetic, and the atomic half never had one.**

**Co-simulation closed without a baseline entry.** The Sail config's own comment gave one reason its map did not split loads and stores: the measurement this branch overturns. The model is this machine now — the 48 KB between the text window and the data RAM stops being a region, the peripheral region becomes one page — and `loadfault.S` and `storefault.S` **AGREE**. Sail rejects a region that does not end on a 4K page, so a sub-page refusal is not co-simulable; those two programs drop that one case each and name where it is graded instead (`test/decoder_tb.v`, and the generated checks, which reach it with a free base register).

Two defects in the branch's own new programs, both from `riscv_test.h`'s handler resuming at `mepc+4` whatever the instruction was: the compressed cases returned into the middle of the next word. Each carries a `c.nop` now. The `loadstore-region-ignored` mutation was paired with `FAIL 2` and really produces `FAIL 3`; `mutation-check` has never been able to run on this tree to say so.

## Base

Merged onto `thejefflarson/jef-853-…` (#216) rather than main, so this rebases once. `test/probe_gates.sh` was reconstructed from the base's copy with this branch's one changed group spliced in, and `PROBES_EXPECTED` read from the run rather than computed.

## Checks

**`soc-timing` is RED on this PR and that is the result, not an accident**: 11.87 MHz against the 12.00 MHz requirement, at CI's own placement.

`make test` 73/73 · `make -C formal check` 86/86 · all six component proofs · `traps-region-probe` both arms red at their own lines · `nonperturbation` PASS · `make cosim-suite` 69/73, matching the baseline · `make probe-gates` 456 · `make fit` 4000 of 4088 · `make lint` clean · `monitor-check` clean.

**`mutation-check` is red on the base and stays red**: `mtip-fires-a-tick-early` no longer applies to `rtl/timer.v`, which this branch does not touch (byte-identical to #216's). #216's own CI reports the same failure.

Closes JEF-855


thejefflarson:
Superseded by #218, which took the deferred spelling of the same feature.

This branch answers the region question in the ISSUING cycle. Its own sixteen-seed sweep measured **+9.40% of median period, nine of sixteen placements under 12.00 MHz with the median at 11.97**, and CI's `soc-timing` read 11.87 — so it was held rather than merged, because `SOC_MIN_MHZ` is the board's clock and does not slide.

#218 answers a cycle late instead and costs **nothing** in period: −1.23% of median, 5 of 16 seeds slower, sign test p = 0.21, **0 of 16 under 12.00**. Same causes 5 and 7 for the same twelve encodings, at 13.79% of Dhrystone's cycles rather than the board clock.

**Nothing here is lost.** The measurement that closed the same-cycle direction is ADR-0128, on `main`; the verification work — `loadfault.S`, `storefault.S`, the `rvfi_mem_fault` plumbing, `formal/traps.sv`'s must-trap region arm and the Sail map split — all went in with #218.

Closing as superseded, not as wrong: it is the reason the deferred version exists.

PR #223, opened 2026-08-24
Claim Zkt: the grader first, the ISA string second

thejefflarson:
## What

Claims Zkt (data-independent execution latency) for this core, grader-first. This PR went through several rounds: the RTL-text grader, three defects found and fixed against it, the rebuild onto the elaborated netlist, two exemptions mechanized that were previously argued only by comment, a polarity bug found in the leaf-walking functions and fixed, and — this round — the recognition that the polarity bug's own fix (and the two checks it lived in) were symptoms of using the wrong mechanism, so those two checks moved to a solver-proved RTL assertion instead.

### This round: move region_stall's own gate and ls_access's encoding set to a proof

Three rounds each defeated a different structural spelling of the same two facts — graded `ls_access` by NAME, then treated a NOT as polarity-preserving, then let a MUX read as an AND under odd polarity. Each fix was correct and each new spelling got through, because both facts are **single-trace exact-set-equality invariants with no VALUE comparison in them at all** — the same shape as `assert(out.is_amo == (out.is_amoswap || ...))`, which already sits in `rtl/decoder.v`'s own `ifdef FORMAL` block. A structural netlist walk is an over-approximation built for a 2-safety question (which the taint half of this script still needs it for); handed an exact 1-safety equality instead, it has edges an exact decision procedure does not.

`rtl/decoder.v` now states both facts as assertions:
```systemverilog
always_comb if (clocked) assert(!region_stall || ls_access);
always_comb if (clocked)
  assert(ls_access == (instr_lb || instr_lbu || instr_lh || instr_lhu ||
    instr_lw || instr_sb || instr_sh || instr_sw));
```
`make -C formal components_decoder` proves both by k-induction. `formal/decoder-zkt-probe.py` is the demonstrated red direction — the same pattern `make -C formal traps-region-probe` sets for `formal/traps.sv`: it builds two cores one line of `rtl/decoder.v` apart and requires each to fail at its own assertion, wired as a prerequisite of `components_decoder`. Its own grading is a new `test/probe_gates.sh` group (13 probes against a stub `sby`, mirroring `traps-region-probe`'s own).

**All eight historical defeats from the three prior rounds were re-run by hand against the new proof and every one still fails** — seven at the new assertions directly, and the eighth (a NOT wrapped around the whole `ls_access` OR) also trips a pre-existing sign-extension assertion nearby, confirmed to fail the new one too with `--keep-going`.

`test/zkt_isolation_test.py` loses the two checks and their machinery entirely (`and_tree_leaves`, `bool_tree_leaves`, `LS_ACCESS_TRANSPARENT`, `LS_ACCESS_ENCODINGS`, `GATE_TERM`, and 8 probes), net +5 probes after the 13 new ones (15 in the Python group, down from 23). What remains is genuinely 2-safety: forward reachability (from `reg_rs1`/`reg_rs2`/`executor_out.rd_data`, twice) and port coverage.

**No RTL behaviour changes**: both new assertions sit inside `rtl/decoder.v`'s `ifdef FORMAL` block, and neither `make fit` nor `make soc-timing`/`netlist-digest` ever defines `RISCV_FORMAL` — confirmed by reading their yosys recipes (`SOC_SYNTH`, `fit.json`'s recipe), which pass no `-D` flags at all. Could not run `make fit`/`make netlist-digest` locally this round — this machine's `nextpnr-ice40` is linked against a stale `libboost` (`dyld: Symbol not found`, reproduces on the toolchain-version-check step before any RTL is read, unrelated to this change). CI's pinned OSS CAD Suite is unaffected.

### Everything from prior rounds (unchanged this round)

**The grader reads the elaborated netlist, not the source text.** `test/zkt_isolation_test.py` follows `formal/check-nonperturbation.py`'s template: `yosys -q -s`, a JSON netlist, cells/nets read off it, zero regexes.

- **Forward reachability**, a flip-flop's D to its Q included, with `region_stall` and `CONTROL_FIELDS` (`out.rd`/`out.is_amo`/`out.valid`/`executor_out.rd`/`executor_out.valid`) blocked as taint sources past their own one hop. `CONTROL_FIELDS`' comment states the real justification: every value-dependence reaching those fields is TRAP-MEDIATED (architecturally visible), not merely narrow — the 5-bit width bound is a tripwire on top of that, not the reason.
- **Port coverage, both ways**, off the netlist's measured widths.

### The ISA string

`rv32imac_zicsr_zifencei_zkt`. `test/march_test.sh` passes at all 7 sites.

### ADR

`docs/adr/0134-claim-zkt-the-grader-first-the-isa-string-second.md` is the first round's record. `docs/adr/0137-the-zkt-grader-moves-to-the-elaborated-netlist.md` got a clean rewrite this round (rather than another patch-on-patch amendment) separating the taint half (unchanged) from this round's move-to-a-proof, with the three-round defeat history told once as a pattern.

## Testing

- `make test`: **74/74 programs pass, exit 0**.
- `make test-units`: all unit benches pass.
- `make probe-gates`: **550/550 probes pass** (`PROBES_EXPECTED` read from the script's own error output every round, never computed).
- `bash test/march_test.sh`: passes at all 7 sites.
- `formal/decoder-zkt-probe.py` run directly: both mutated cores fail at their own assertion's line.
- All eight historical Zkt defeats re-run by hand against the new RTL proof: all still fail.
- The shipping gate and a semantically-equivalent De Morgan re-spelling of it both still pass the proof (verified by hand — alternate correct spellings pass automatically now that a solver decides it, one of the reasons this is better).
- Formal set: `make -C formal check` (86/86, baselines match exactly), `make -C formal nonperturbation` (PASS), `make -C formal components_decoder components_executor components_pcloop components_traps` (all `DONE (PASS, rc=0)`).

## Scope notes

- **No RTL behaviour changes** — the two new assertions are inside `ifdef FORMAL`, confirmed by source inspection (`make fit`/`make netlist-digest` unmoved, could not run locally due to an unrelated broken `nextpnr-ice40`/`libboost` link on this machine; CI's pinned toolchain is authoritative here).
- No use of "master" introduced or reintroduced.

Closes JEF-868.


thejefflarson:
## HELD — the grader cannot fail for the three shapes it exists to catch

Reviewed as architect. All 11 required checks are green and `make test` is 73/73, so this is not a CI hold. It is held because the PR's headline claim — `_zkt` in `-march`, the core's only runtime statement of data-independent latency — rests entirely on `test/zkt_isolation_test.py`, and I reproduced three separate ways to defeat that gate. This repo's own rule is that a grader that cannot fail is not a grader.

**Today's RTL still satisfies the Zkt property.** What is broken is the grader, not the hardware. That is why this is a hold on the claim's warrant rather than a bug report against `rtl/`.

Evidence below is from running the PR's own script against the PR's own `rtl/decoder.v`, one mutation at a time.

### 1 · CRITICAL — `taint_closure` treats a procedural copy of the seed as a clean leaf

`test/zkt_isolation_test.py:171` walks continuous `assign`s only. Anything written in `always_ff`/`always_comb` has no entry in `assigns`, so it is a leaf. The docstring calls this "the deliberately conservative reading" — it is the opposite: a signal that *carries* the seed value is read as unable to propagate it.

`rtl/decoder.v:1083` is `out.rs1 <= reg_rs1;` — a verbatim procedural copy. `branch_taken` (`:966-976`) is an `always_comb` case over `cmp_eq`/`cmp_lt`, both of which read `reg_rs1`/`reg_rs2`.

The script's own declared red direction fails correctly:

```
assign hazard = hazard_rs1 || hazard_rs2 || serialize || reg_rs1[0];
  -> ZKT STALL ISOLATION: FAIL   (exit 1)
```

Both laundered spellings of the same defect pass:

```
assign hazard = hazard_rs1 || hazard_rs2 || serialize || out.rs1[0];
  -> ZKT STALL ISOLATION: PASS   (exit 0)
assign hazard = hazard_rs1 || hazard_rs2 || serialize || branch_taken;
  -> ZKT STALL ISOLATION: PASS   (exit 0)
```

Each is a real operand-value-dependent stall on a Zkt-listed instruction. The fail-closed check at `:222` does not help — it covers only the ~16 names in `MUST_NOT_DEPEND`/`EXPECT_TAINTED`/`GATED_SIGNAL`, and `out.rs1` is not among them.

### 2 · HIGH — `||` binds looser than `&&`, and the splitter only splits on `&&`

`top_level_and_terms` (`:120`) splits on `&&` at depth 0. One line defeats **both** halves of the gate at once — the `ls_access` conjunction check and the taint check, since `region_stall` is blocked as a taint source at `:233`:

```
assign region_stall = ls_access && !ls_settled && !ls_answer_valid || reg_rs1[0];
  -> "reg_rs1/reg_rs2 reach only region_stall among the stall reasons,
      and it stays gated on ls_access."
     ZKT STALL ISOLATION: PASS   (exit 0)
```

That mutation makes `region_stall` assert on rs1's low bit for *every* instruction, Zkt-listed ones included, and the gate reports that it stays gated on `ls_access`.

### 3 · HIGH — `SEEDS` is two hardcoded names, and the likeliest future change is invisible

`:53` seeds only `reg_rs1`/`reg_rs2`. `executor_out.rd_data` is a 32-bit register value already on a decoder port (`rtl/decoder.v:27`) and is not a seed. That is exactly the forwarding-network shape this repo has costed out three times (ADR-0083, ADR-0092, ADR-0100) — the change most likely to reintroduce operand-value-dependent timing is the one the gate cannot see.

### What would clear the hold

The three compose: all of them blind the same grader to the same future change. Any of these directions is fine, but the gate has to be able to fail:

- extend the taint walk to non-blocking and `always_comb` assignments, or refuse to run when a `MUST_NOT_DEPEND` signal's cone touches a procedural block it cannot trace (fail closed, the way the `:222` missing-name check already does);
- split on `||` as well as `&&`, or normalise the expression before checking the gate term;
- seed every register-file data output reaching the decoder, `executor_out.rd_data` included.

Then add the three mutations above to the probe group so they are demonstrated red — `test/probe_gates.sh` is where this repo requires that, and the two currently-passing ones are the red directions this gate is missing.

Nothing here asks for an RTL change, and the ISA-string half is fine on its own once the grader behind it can go red.

thejefflarson:
Fixed the three ways the grader was reported to be unable to fail, per the architect's review. Pushed as a new commit (branch rebased onto latest main, b165442).

**Fix 1 — taint severed at procedural signals.** Added a backward fan-in walk (`fanin_leaves`) over the seven value-blind stall reasons plus `region_stall`: any identifier it reaches that is not itself a continuous assign must be in a new `KNOWN_CLEAN_LEAVES` allowlist or in `SEEDS`, or the run stops (exit 2) rather than reading it as clean. Verified `|| out.rs1[0]` and `|| branch_taken` (the two reported counterexamples) now both fail.

**Fix 2 — `top_level_and_terms` ignores `||`.** `region_stall`'s own assign is now checked for a depth-zero `||` beside its `&&` chain, independent of taint (`stall`/`stall_other` read `region_stall` by name, never its assign body). Verified `assign region_stall = ls_access && !ls_settled && !ls_answer_valid || reg_rs1[0];` now fails.

**Fix 3 — SEEDS was two hardcoded names.** `SEEDS`/`NON_VALUE_INPUTS` are now checked (both directions) against the decoder's own port list, parsed with matched parentheses the way `test/port_connect_test.py` reads `littlecpu`'s — every input wider than a register NUMBER (5 bits) must be classified or the run stops. `executor_out.rd_data` is now in `SEEDS`. Verified `|| executor_out.rd_data[0]` in a stall reason now fails.

Three new probes (on top of the existing 7) demonstrate each fix failing at its own site — `PROBES_EXPECTED` is 516, as reported by the script itself. `make test`, `make test-units`, and `make probe-gates` all pass. Fixed the docstring and CLAUDE.md's Zkt paragraph, which overstated what "hard error, not read as clean" covered before this fix. Renumbered the ADR to 0134 since 0131-0133 landed on main first.

thejefflarson:
## HELD — the substrate, not the three findings

CI is green on all eleven required checks and I verified that independently. The hold is not
about a red check. It is about what backs the claim.

### Severity-ordered

**1. `test/zkt_isolation_test.py:131` — the allowlist re-admits the bug it closes.**
`ls_answer_valid` sits in `KNOWN_CLEAN_LEAVES`, under a comment stating every entry there holds
"register NUMBERS or control state, never a register-file DATA output". That is false for this
entry. Verified against `rtl/decoder.v`:

```
assign ls_block    = reg_rs1[31:LS_BLOCK_BITS];        // :519
assign ls_settled  = ls_text_deep || ls_ram_deep;      // :531
assign region_stall = ls_access && !ls_settled && !ls_answer_valid;  // :556
assign ls_capture   = region_stall && !stall_own;      // :557
... else if (ls_capture) ls_answer_valid <= 1'b1;      // :584-586
```

`ls_answer_valid` is a function of `reg_rs1`'s VALUE. Taint stops at it because it is an
`always_ff` leaf, so `assign stall_own = ... || (ls_answer_valid && x);` makes a Zkt-listed
instruction's stall depend on a register value while the gate prints PASS. Separately: `SEEDS`
and `NON_VALUE_INPUTS` are both checked for stale entries; `KNOWN_CLEAN_LEAVES` is not.

**2. `:214` — `parse_assigns()` rests on a false claim about the language.** The docstring says
"module-scope `assign` is the only continuous-assignment form Verilog allows". SystemVerilog also
has procedural continuous assignment (IEEE 1800 §10.6.1), and an `assign` inside an un-taken
`generate if (0)` or a non-`RISCV_FORMAL` `ifdef` is text this parser records but the elaborator
never builds. `generate if (0) begin assign x = 1'b0; end endgenerate` beside a real
`always_comb x = reg_rs1[0];` gives `x` a node with empty deps — not a leaf, so the allowlist
check passes; no taint, so `assign hazard = ... || x;` reports PASS.

**3. `test/probe_gates.sh:1450` — two new graded comparisons have no probes.**
`unclassified_ports` (`zkt_isolation_test.py:492`) and `stale_classifications` (`:501`) have never
once been executed. That is the half of the fix advertised as "checked BOTH ways", and it breaks a
named rule in `CLAUDE.md` word for word: "Every graded comparison must have a demonstrated red
direction that fails for the reason it was written; `make probe-gates` forces all of them." The
group has ten probes and none reaches either path. Live parser hole behind it: `RANGE` (`:339`) is
`\[([0-9]+):([0-9]+)\]`, so `input logic [XLEN-1:0]` computes width 1 via `bit_width(None)` and is
silently exempt from the classification the check exists to force. Latent today — every `decoder`
port is a numeric literal — and it defeats the check's entire stated purpose the day it is not.

### Why this is a re-ticket and not a fourth patch

The three findings are one category error. Each is a fact about ELABORATION that a text parser
cannot know: state through a procedural block, which `generate`/`ifdef` arm is live, constant
folding of a parameterised width. A fourth patch buys a fourth hole.

The file's own stated model is the answer. `CLAUDE.md` describes this check as "in the shape
`formal/check-nonperturbation.py`'s cone check already uses" — but that script contains **zero**
regexes (verified), runs yosys, writes JSON netlists and reads those, with `check=True`. This one
reads source text with regexes. The rulebook claims a substrate the implementation does not use.

The decisive point is what is being claimed. `-march=...zkt` is an outward-facing, hard-to-reverse
promise about a SECURITY property — data-independent execution latency — read by toolchains and by
anyone consuming this core. The evidence behind it is one script, and that script has now been
defeated three times in three rounds. A grader with a false allowlist entry and two never-executed
comparisons is not enough to stand behind a constant-time claim. Something too clever to fully
reason about is a reason to reject it, and this is the case for that.

`rtl/` is untouched, so nothing here says the core's timing behaviour is wrong. What is not
established is that the repo can KEEP it right, which is the whole point of claiming the extension.

### The shape the rework should take

Rebuild the Zkt isolation check on the yosys netlist, the way `formal/check-nonperturbation.py`
already does: elaborate, `write_json`, walk the cell graph. That removes findings 1, 2 and 3's
parser hole as a class rather than one at a time — a netlist has no un-taken `generate` arms, no
un-elaborated `ifdef` text, no unresolved parameters, and an `always_ff` is a cell with visible
inputs rather than an untraceable leaf. Then: a red direction for every graded comparison, and an
ADR recording that the Zkt grader runs on the netlist and why the RTL-text version was rejected.
ADR number 0134 is left free for it.

The ISA-string change should land WITH that grader, not ahead of it.

PR #224, opened 2026-08-24
Port CoreMark, simulated at 16 KB of ROM, and decline the Hazard3 comparison

thejefflarson:
## Problem

CLAUDE.md called `make dhrystone` "the only figure comparable to another project's." That stopped being true once Hazard3 (the RISC-V core inside the RP2350) became a core worth comparing against — Hazard3 publishes CoreMark/MHz, not Dhrystone.

**The obvious next step is also a mixed-configuration error.** Hazard3's published 4.15 CoreMark/MHz is its RP2350 build (Zba/Zbb/Zbs/Zbkb, a fast multiplier, a branch predictor). Its iCE40/iCEBreaker build (`EXTENSION_C=0`, no bitmanip, `MULDIV_UNROLL=1`, no branch predictor, at 12 MHz) has none of that, and has no published CoreMark/MHz of its own. So this PR does not compare to Hazard3 — it records what CoreMark reports on this core, states the configuration, and declines the comparison nobody has the other half of yet.

## What changed

- `test/bench/coremark/` vendors five CoreMark files **unmodified** from `eembc/coremark` at `1f483d5b` (Apache-2.0). CoreMark's trademark terms permit quoting the name only for an unmodified copy, so `test/bench/coremark/PINNED.sha256` is checked on every `make coremark` run — this repo's own manifest, plus (see below) `make revendor-coremark` for verification against upstream itself.
- `test/bench/core_portme.h` / `test/bench/coremark_port.c` are the porting layer, mirroring `test/bench/dhry_port.c`'s split for Dhrystone: `mcycle`/`minstret` time the run, the report goes into a RAM buffer and out `rtl/uart.v` when `COREMARK_UART` is set, and the build `#error`s without `COREMARK_FLAGS`.
- `test/bench/coremark.lds` links against `test/testbench.v`'s 16 KB simulated `ROM_WORDS` — CoreMark's image is 13,272 bytes and does not fit the shipping 8 KB. The figure is labelled "SIMULATED AT 16 KB OF ROM" everywhere it's printed (the Makefile target, the runner, and the program's own report), because it describes a machine that cannot be built until the deferred SPI-flash boot path lands.
- `make coremark` reports **1.811 CoreMark/MHz** at 100 iterations (default), self-check PASS against EEMBC's published 2K-performance CRCs. The ratio reproduced to three decimal places at 10, 40 and 100 iterations, so the default trades EEMBC's literal 10-second minimum (a rule for averaging out a real clock's jitter, which a cycle-exact simulator doesn't have) for a couple of minutes of `--stalls`-instrumented runtime instead of ~15.
- `test/march_test.sh`'s Makefile occurrence count moves 3→4 for the new `COREMARK_CFLAGS` site (a plain load/store, "silent when wrong," program that never executes an atomic — the same category `DHRY_CFLAGS` is already in), and `test/probe_gates.sh`'s matching probe fixture is updated to match.
- `docs/adr/0136` records the number, the configuration, the stall-reason decomposition (`--stalls` breakdown: hazard 28.9%, region 10.2%, operand 9.3%), and the Hazard3 comparison caveat.
- CLAUDE.md's dhrystone bullet is corrected and a new bullet + `make coremark` line added, kept minimal per the note about JEF-868/JEF-869 also touching this file concurrently.

## Security fixes

A review held this PR on four High findings that chain end-to-end into an attacker-chosen CoreMark score, PASS verdict and CPI table with every integrity check reporting green. All four are fixed:

1. **The PASS oracle could pass when it never ran** (`coremark_port.c`). `portable_fini()` used to trust `core_main.c`'s `err` field alone, and `err` stays zero whenever the seed/size CRC doesn't match one of EEMBC's five known configurations — `core_main.c` sets `known_id = -1`, skips the whole CRC comparison, and never touches `err`. `TOTAL_DATA_SIZE` is an overridable build flag, so a differently-sized run sailed through unvalidated. `portable_fini()` now re-derives "the checked 2K performance configuration actually ran" itself: `execs == ALL_ALGORITHMS_MASK`, the 2K seeds and size, the three CRCs compared directly against EEMBC's published values (`0xe714`/`0x1fd7`/`0x8e3a`), and `check_data_types()` folded into the conjunction. A run that can't be validated writes a third, distinct `tohost` code (`COREMARK_UNVALIDATED = 5`, vs `PASS = 1` / `FAIL = 3`) and prints its own "COULD NOT BE VALIDATED" line, so it can no longer exit the way a PASS does.
2. **The manifest check was one-way** (`run_coremark.sh`). `shasum -c` only verifies the names `PINNED.sha256` lists; it says nothing about a file dropped in beside them. Concretely: `coremark.h`'s `#include "core_portme.h"` is a quoted include, which searches the including file's own directory first, so an unlisted `core_portme.h` in the vendor directory would shadow this port's real header (which carries the timing hooks) for every vendored unit while `shasum -c` reported the tree unmodified. Fixed with a `comm -23`/`comm -13` two-way set comparison before `shasum -c` runs at all — the same idiom `test/check_suite_shape.sh` and `test/dual_build.sh` already use for a manifest against a directory.
3. **No `--strict`, and the format warning was discarded** (`run_coremark.sh`). `shasum -c` silently skips a malformed manifest line and still exits 0 if a surviving line matches; the pin-check output was captured into a variable that was only printed on failure, so the `WARNING: N line(s) improperly formatted` message never reached anyone. Fixed: both `shasum` and `sha256sum` branches now pass `--strict`, and the pin check's own output is printed on the success path too.
4. **The accounting grader was gated on guest-controlled text** (`run_coremark.sh`). `test/cxxrtl.cc` prints the guest's own console buffer *before* its own `STALLS`/`RETIRES` lines, and the extraction used first-match, so a guest program that happened to write text starting with `"STALLS "` or `"RETIRES "` could supply the published CPI table. Fixed two ways: the extraction now keeps the **last** match of each line in the log (the runner's own accounting is always printed last, since nothing follows it), and an empty stall/retire line is now a hard `exit 1` (adopting `run_dhrystone.sh`'s spelling) instead of silently skipping `stall_report.py`.

Added a `test/bench/run_coremark.sh` probe group covering all four manifest/pin paths — a mutated vendored byte, a deleted manifest line, an unlisted file (`core_portme.h`, the concrete exploit), and a malformed manifest line — each demonstrated red for its own reason, plus a control proving an unmodified vendor tree passes silently.

`docs/adr/0134`/`0135` landed on two other in-flight branches; the CoreMark ADR is renumbered to **0136**.

## The fifth finding: `make revendor-coremark`

**Implemented, not declined.** `PINNED.sha256` proved only "unchanged since committed" — which git already says — never that the committed bytes are `eembc/coremark`'s. An earlier pass at this PR declined a revendor target on "new trust boundary" grounds; that reasoning didn't hold up. This repo already fetches over the network in several places (`make sail-setup`, `make lint-setup`, `formal/pin.mk`'s clone) and `formal/pin.mk` in particular clones a third-party repo and **executes its Python**, committing the stdout as tracked source — a materially larger standing trust boundary than fetching a pinned commit and diffing bytes, which executes nothing it downloads.

`make revendor-coremark`:
- Pins `eembc/coremark` at `1f483d5b8316753a742cbf5590caf5bd0a4e4777` via `override COREMARK_PIN` (validated 40-hex, rejects command-line/environment override — the same shape `formal/pin.mk`'s `RISCV_FORMAL_SHA` uses).
- Is optional and off `make test`'s path, modelled on `make sail-setup`: nothing calls it, so a machine with no network still runs the full suite.
- Fetches the pinned commit from `codeload.github.com` and reads each vendored file straight out of the archive **by its known path** (`tar -O`), never extracting the tarball to disk — so there's no path-traversal surface to check, unlike `sail-setup`'s full-tree unpack.
- Reads the file list out of `test/bench/coremark/PINNED.sha256` rather than a second hardcoded copy in the Makefile.
- Fails closed: a network/tool failure, a file missing from the pinned archive, or any byte mismatch is a nonzero exit with the specific file(s) and both hashes named. A match is never assumed.
- Added a 4-probe group mirroring `make sail-setup`'s: a control (real archive matching the tree, endorsed), a substituted non-archive download (refused for every file), a real archive with one byte-mutated file (caught and named), and a real archive missing one member (caught and named, distinct from the corrupted-download case).

**Ran it against the real pin — all 8 files match exactly** — and it is what established the `coremark.md5` finding below, rather than asserting it.

### `coremark.md5`'s disagreement, resolved

Upstream's `coremark.md5` is stale in **exactly one of its six entries**. Fetched and diffed against the pinned commit:

| | `coremark.h` |
|---|---|
| `coremark.md5` claims | `8ca974c013b380dc7f0d6d1afb76eb2d` |
| actual (== upstream's own byte content at the pin) | `b0ec69b6c8e75853d06accb3b1bcf534` |

The other five entries (`core_list_join.c`, `core_main.c`, `core_matrix.c`, `core_state.c`, `core_util.c`) match. Upstream edited `coremark.h` at some point and never regenerated `coremark.md5`. This is upstream staleness, not tampering, and not an open question — `make revendor-coremark` reproduces the same one-file disagreement on demand. Recorded with both hashes and the one-file scope in `test/bench/coremark/PINNED.sha256`'s header and in `docs/adr/0136`, replacing the earlier "cannot be trusted... disagrees" wording that asserted the conclusion without the evidence.

## Testing

- `make coremark` — self-check reports **PASS (2K performance seeds and size, list/matrix/state CRCs matched against EEMBC's published values, check_data_types clean)**, exit 0, **CoreMark/MHz unchanged at 1.811** (55,200,557 cycles / 28,485,349 instructions / 100 iterations — identical every run across every fix in this PR; the stricter oracle validates the same measured configuration, it doesn't change what was measured).
- `make test` — 73/73 passed, exit 0.
- `make test-units` — all unit benches pass, exit 0.
- `make probe-gates` — **515 graded comparisons, every failure path executed**, exit 0 (up from main's 506: +5 for the `run_coremark.sh` manifest/pin group, +4 for the `revendor-coremark` group — each probe verified to fail for its stated reason before being accepted, and `PROBES_EXPECTED` read from the script's own error output, never computed by hand).
- `make revendor-coremark` — run against the real pin: all 8 files match `eembc/coremark` at `1f483d5b8316753a742cbf5590caf5bd0a4e4777` exactly, exit 0. Also verified both fail-closed directions by hand with a substituted `curl` (a non-archive response, and a real archive with one byte mutated) before writing the probes for them.
- `soundcheck:pr-review` on the diff — no Critical/High findings.
- `/simplify` pass: reused the repo's `comm -23`/`comm -13` idiom instead of a hand-rolled string-equality check for the manifest comparison, and replaced a `python3 -c` probe fixture with a one-line `sed -E` mutation matching the file's existing `ma_edit`-style idiom. Re-ran the gates after to confirm no behavior change.

## Scope notes

- Vendoring under a separate `Apache-2.0` `LICENSE.md` inside `test/bench/coremark/` alongside the repo's own MIT `LICENSE` is intentional and standard practice for third-party vendored code of a different license.
- `COREMARK_ITERATIONS` defaults to 100 rather than the ~120M-cycle figure a literal 10-second-at-12MHz run implies; the tradeoff and the measurement backing it are documented in the Makefile comment and the ADR. Raise it with `COREMARK_ITERATIONS=<n>` for a longer run.
- Neither `make coremark` nor `make revendor-coremark` is on CI, a gate, or a ratchet — same as `make dhrystone` and `make sail-setup`.

Closes JEF-867


thejefflarson:
## HELD — the figure is probably right; the gate that would catch a wrong one is not

Reviewed as architect. All 11 required checks are green, the port is careful, and **declining the Hazard3 4.15 comparison was the right call** — that is an RP2350 configuration, and Hazard3's iCE40 build is RV32IMA with no C, no bitmanip, `MULDIV_UNROLL=1` and `CSR_COUNTER=0`. Refusing to quote a number measured on another part against a different configuration is exactly this repo's rule that a product is only a measurement when both factors came off one tree.

The hold is not about the 1.811. That run used the standard 2K configuration, so the CRCs did compare and the figure is very likely real. The hold is that **nothing in the runner would have stopped a wrong one from being published as validated**, and a benchmark harness whose oracle can be switched off from the command line is the shape this repo has been burned by before.

### 1 · HIGH — the PASS oracle is skipped, not failed, on an unknown configuration

`test/bench/coremark_port.c:330` reports PASS as `res->err == 0`. But `err` initialises to 0 at `coremark/core_main.c:191`, and every write to it is inside `if (known_id >= 0)` (`:321`, accumulated at `:353`). An unknown seed/size sets `total_errors = -1` at `:318` and **skips the CRC comparison entirely** — so `err` stays 0, the port prints PASS, writes `tohost[0]=1` and exits 0, and the score is published as validated.

`COREMARK_CFLAGS` is an overridable make variable, so `make coremark COREMARK_CFLAGS=-DTOTAL_DATA_SIZE=...` disables the oracle from the command line. The port also reads `res->err` rather than `total_errors`, which drops `check_data_types()` (`:356`) and the 10-second rule.

Fix: gate PASS on `known_id >= 0 && total_errors == 0`, and refuse to print a figure at all for a configuration whose CRCs were not compared.

### 2 · HIGH — the pin check is over listed names, not set equality, and an added file shadows the porting header

`test/bench/run_coremark.sh:70` runs `shasum -a 256 -c PINNED.sha256`, which verifies the files the manifest lists and says nothing about a file that is not in it. `coremark/coremark.h:38` is `#include "core_portme.h"` — a quoted include, resolved from the including file's own directory first — and `:125` also passes `-I "$VENDOR_DIR"`. So an **unlisted** `core_portme.h` dropped into the vendored directory shadows the repo's porting header, which is where the timing hooks live, and the pin check still reports the tree unmodified.

Fix: compare the directory listing against the manifest as a set, both directions, the way this repo grades every other baseline.

### 3 · HIGH — `shasum -c` without `--strict`, with the warning discarded

`:57` omits `--strict`, so a malformed manifest line is skipped with a warning and the check still exits 0. `:70` captures both streams into `$pin_check` and prints it only on failure, so the format warning is discarded on the path where it matters.

### 4 · HIGH — whether the only CPI grader runs is decided by guest-written text

`:224` decides whether `test/stall_report.py` runs by matching STALLS/RETIRES lines in the program's own output; absent them, it is skipped and the script exits 0. `test/bench/run_dhrystone.sh:193-197` hard-`exit 1`s on the identical condition. This is a regression from an in-repo precedent rather than a choice, and it composes with the rest: `test/cxxrtl.cc:427-433` prints the guest console *before* the runner's own RETIRES/STALLS lines and the script takes the first match, so guest text can supply both the score and the CPI table with every check green.

### 5 · HIGH — `PINNED.sha256` has no upstream binding

`test/bench/coremark/PINNED.sha256:16` proves only "unchanged since committed", which `git` already says. Commit `1f483d5b` appears in a comment no script reads, and upstream's own `coremark.md5` disagreed with the vendored header and was discarded rather than reconciled. Compare `formal/pin.mk`: a 40-hex SHA, a make `override`, and a fail-closed `rev-parse` re-derivation. Either reconcile the upstream digest and record why it differed, or say plainly in the file that the pin is local-only — but the trademark argument in the failure message ("an unmodified copy of the benchmark") is a claim this file cannot currently support.

### Not blocking

`test/probe_gates.sh:30` — the ~14 graded failure paths in `run_coremark.sh` have no forced red direction. `make coremark` is on no workflow and `run_dhrystone.sh` never had a probe group either, so this follows bench convention rather than regressing a gate. Worth adding once the above lands, since by then the graders will be worth probing.

### What would clear the hold

Findings 1 and 4 are the two that matter; 2, 3 and 5 are the pin check being decorative. None of this requires re-measuring — if the oracle is repaired and the standard configuration still reports 1.811, the number stands as it is with a gate behind it.

PR #226, opened 2026-08-24
Reach the SPI flash for simulation, keep it off the board's shared pins, and fix the UART's own conflict (JEF-866)

thejefflarson:
## Summary

`rtl/spiflash.v` (a read-only SPI controller for the on-board configuration
flash) and `soc/pin_lockout.v` (a chip-select arbitration primitive) shipped
in this PR's earlier commits, wired into `soc/board_upduino.v` to share four
UPduino pins between the on-chip controller and the host programmer
(`iceprog`). Review held on two findings against that wiring:

1. **The arbitration's predicate can't deliver the property.** `released`
   answers "does pin 16 read high", which is electrically identical for a
   host that is genuinely absent and a host that is idle with its own chip
   select parked high — `iceprog` parks it that way, as a push-pull MPSSE
   output, for most of a programming session and not a corner of it. A grant
   built on that predicate could still drive a live driver.
2. **The grant never expired.** `want` was a level with nothing to lower it
   but firmware; firmware that trapped, hung, or busy-waited after asserting
   the chip select held the pins driven for the life of the chip.

## What changed

- **The on-chip flash controller's own pins are not wired to this board at
  all.** `rtl/spiflash.v`'s `sck`, `mosi`, and its own `cs_n` are tied off in
  `soc/board_upduino.v` (`spi_miso` high, the rest unconnected), so finding 1's
  unsound predicate is unreachable from the shipping design. `rtl/spiflash.v`
  and `soc/pin_lockout.v` both still ship — reachable in simulation
  (`test/asm/spiflash.S`, `test/spiflash_tb.v`) and graded standalone
  (`test/pin_lockout_tb.v`) — but wiring the controller to real pins is a
  future ticket's, and it owes a real-board measurement of the contention
  window described above.
- **Pin 14's own, older conflict is fixed rather than deferred.** That
  conflict predates this ticket: `soc/board_upduino.v` drove pin 14 (shared
  with the flash's own MISO) unconditionally from the day the board wrapper
  was written, which is what actually caused the observed hardware failure
  (`cdone` high after reset, a late JEDEC id, a write error) this PR opened
  with. The fix reads pin 16 (the flash's own chip select) through a
  free-running two-flop synchroniser and drives pin 14 with the UART only
  once that reads released — a sound predicate, unlike `released`'s: an SPI
  slave's own output driver is a deterministic function of its chip select,
  so this answers "is the flash's own driver off" with no ambiguity about who
  or what is holding the select at a given level.
- **`soc/pin_lockout.v` gains a bounded grant** (fixes finding 2): a new
  `busy` input and an idle timeout force a fresh sample if a request goes
  idle for a byte time, so a hung requester loses the pins within a bounded
  window once something wires it up, rather than holding them forever. The
  module's own comment states the one open question honestly: the
  sixteen-cycle bound is unmeasured against real compiled firmware's
  inter-byte overhead.
- **The false safety claim is corrected.** "The two controllers simply never
  share a wire" is removed from CLAUDE.md and ADR-0135; both now state the
  predicate's actual, narrower guarantee.
- **"Master" is replaced with "controller"** throughout the touched files
  (`rtl/spiflash.v`'s header, `soc/board_upduino.v`, `soc/pin_lockout.v`,
  CLAUDE.md, ADR-0135, `docs/adr/README.md`).
- `soc/board_elaborate.sh`'s two mutation probes move with the wiring they
  test (a port rename, and now the pin-16 synchroniser's input left
  unconnected) rather than describing arbitration that no longer ships.

## Ratified, not revisited

- The SPI controller is read-only by firmware convention rather than
  construction (unchanged).
- 12.06 MHz worst-of-sixteen is not a regression — `littlesoc`'s
  `make soc-timing` numbers are untouched by this change, since
  `soc/board_upduino.v` is not part of that graded flow.

## Testing

- `make test` — 74/74 `.S`/`.c` programs, `probe-gates` (522 probes,
  `PROBES_EXPECTED` updated from the script's own error), `board-elaborate`
  (2 mutation probes), `window-test`, `imem-share-test`: all green.
- `make test-units` — all unit benches, including `pin_lockout_tb` (new
  vectors: a real transfer's `busy` duty cycle never trips the bound; a hung
  request loses the grant within one byte time and gets it back
  uncontested; a host that takes the pin in the open window keeps the grant
  withheld) and `spiflash_tb` (unaffected).
- `make fit`: 4109 cells, within budget (core is untouched by this PR).
- `make soc-timing`: 13.00 MHz single-seed (untouched design; ADR-0135's
  existing sixteen-seed sweep for `rtl/spiflash.v` on the bus stands).
- `make board-elaborate` / `make bitstream`: elaborates and places cleanly
  (12.55 MHz single-placement estimate, no ratchet).
- `make -C formal check` (86/86), the four component proofs (decoder,
  executor, pcloop, traps — each with its own anti-vacuity prerequisite),
  `nonperturbation`, and `remeasure-fg` (F=6, G=6, both reproduce exactly) —
  all green. None of this PR's changes touch `rtl/` core files.

Closes JEF-866.


thejefflarson:
## HELD — a required check is red at this PR's own new gate. Three decisions ruled below.

Reviewed as architect. The engineering here is careful and the report leads with what it did not solve, which is the right way round. It is held on one hard blocker and one design finding; the two decisions escalated to me are both ruled in this PR's favour.

### BLOCKER · the required `test` check fails at `make board-elaborate`

[Job 97392519269](https://github.com/thejefflarson/little-cpu/actions/runs/32714414592/job/97392519269) — `test` is a required check and it is red:

```
== soc/board_upduino.v: the wrapper as it ships
FAIL the board wrapper did not elaborate cleanly (rc=1, warnings=no):

== and two ways of breaking it, each required to be caught
FAIL a port the SoC does not have was accepted, or refused for some other reason:
FAIL the enable reading a pin nothing drives was accepted, or refused for some other reason:

3 of 3 elaborations did not behave as required.
make: *** [Makefile:1211: board-elaborate] Error 1
```

All three cases fail, including the shipping file, which points at a shared cause rather than three separate ones. **I could not reproduce it locally** — `make board-elaborate` passes under Homebrew yosys `0.68+post` (`c12172fb`) and under the cached OSS CAD Suite `0.68+48` (`ff5817c3`). CI's is `0.68+120` (`a34d3baa-dirty`), so the difference is most likely the yosys build.

**The second defect is the more important one:** the grader printed no reason. `soc/board_elaborate.sh:66` and `:74` report with `grep -E '^(ERROR|Warning)'`, and nothing in the log matched — so whatever CI's yosys said, this script cannot show it. A grader whose failure output is empty is not diagnosable from the job log, and this file's entire purpose is to be a grader. Please make the failure path dump the tail of the log unconditionally when the grep finds nothing, then re-run to see the actual message.

Worth noting the gate has never been green in CI: it arrived in the last two commits of this branch, and this is the only CI run on it.

### HIGH · the chip-select arbitration is one-directional, and the obvious fix oscillates

`soc/board_upduino.v:110` is `assign own_flash = !soc_cs_n;` — derived purely from the on-chip master's own chip-select register, with no reference to `ssn_in`.

Pin 14 is genuinely protected: `OUTPUT_ENABLE(!own_flash && ssn_in)` releases the UART when the host selects the flash. Pins 15, 16 and 17 are not — all three are `OUTPUT_ENABLE(own_flash)`, so firmware asserting its own chip select drives them regardless of what the host is doing. CRESET is not wired on this board, so the old design keeps running throughout an `iceprog`, and firmware reading the flash during programming puts two CMOS drivers on `spi_sck` and `spi_mosi`. That is the same contention class already observed here — `cdone` high, a JEDEC id a byte late as `FF EF 70 16`, then a write error — which is the very failure this change set out to fix. Pin 14 is fixed; 15 and 17 are newly exposed.

Pin 16 is the narrower case: the FPGA only ever drives it low (`D_OUT_0(1'b0)`), so it contends only while the host drives it high.

**Do not apply the one-line fix.** `assign own_flash = !soc_cs_n && ssn_in;` closes a combinational loop through the pad: `own_flash` raises `io_ssn`'s output enable, which drives the pin low, which makes `ssn_in` read 0, which drops `own_flash`. The arbitration has to sample `ssn_in` on a cycle it is *not* driving and latch the decision — permission to select is granted once, before assertion, and held for the transaction. That is a small FSM, not an expression, and it needs its own vector in the bench.

### DECISION 1 · the read-only-by-convention SPI master — **RATIFIED**

The master is read-only by firmware convention rather than by construction: nothing stops it shifting `0x06` then `0x20` at the bitstream's sector. I agree this is not a finding, and I am recording the reasoning rather than just the verdict.

Enforcing read-only in hardware means the module has to know commands. It currently does not — a JEDEC id read and an `0x03` sequential read are the same eight clocks a byte — and that is the whole of why it is small. Adding a command decoder to enforce a policy braids two concerns that are better kept apart, and buys a guarantee against an actor (this repo's own firmware) that is not in the threat model. The write-enable latch needs two commands in the right order, recovery is a reflash rather than a dead board, and the trade is already documented at the call site. Keep the shifter dumb and the property firmware's.

### DECISION 2 · the 12.06 MHz tail — **NOT a regression**

`CLAUDE.md` decides this one directly. 12 MHz is a requirement, not a regression floor, and the record it already carries is that this design's own worst-of-sixteen has read 12.39, 12.21 and **12.03** on three earlier trees — "three samples and not a trend, every one meeting the requirement and none of them a regression". 12.06 sits inside that range. The median moved the *right* way at −1.60% with only 3 of 16 seeds slower, and 0 of 16 placements are under 12.00.

So this is another sample of a wide tail, not a cost. The tail does not hold this PR, and when it lands the ADR should quote the sweep it came from rather than the margin.

### The unreachable "~13 KB"

Leading with the headline problem being unsolved was right, and the reason is worth keeping: `rtl/imemory.v`'s range tests are reductions, so `ROM_WORDS` must be a power of two, and the next size above 8 KB is 16 KB in 32 block RAMs against 26 free. There is no ROM size between the one that ships and one that does not fit. That is a real finding and it belongs in the record whatever happens to the rest of this branch.

### To clear the hold

Fix the `board-elaborate` failure (and its blind failure path), and either latch the arbitration properly or narrow this PR to drop the shared-pin change. Everything else here — the master, the model, the bench, the `.S` program, F=6/G=6 reproducing, formal 86/86 — is in good shape.

Also: this branch takes `docs/adr/0131`, which is now merged as ADR-0131 from a different PR. Renumber to the next free number when you rebase; `0133` is my suggestion, and `CLAUDE.md`'s two citations of 0131 in this branch move with it.

thejefflarson:
## HELD — the arbitration predicate cannot deliver the property the ADR and CLAUDE.md claim

All thirteen required checks are green and I confirmed that independently. The hold is about two
real-hardware findings and about what this PR writes into the permanent record.

### Severity-ordered

**1. `soc/pin_lockout.v:28` — the predicate is wrong, not the FSM.** `released` is
`ssn_in`, pin 16's level read through `SB_IO`'s `D_IN_0`. A single-ended input read cannot
distinguish a pin the FTDI is actively driving HIGH from a pin nobody drives that the `SB_IO`
`PULLUP` holds HIGH. Both read 1. There is no spelling of "read the pin" that tells them apart.

`iceprog` drives chip select as an MPSSE push-pull output and parks it HIGH between its own
transactions for the whole session. So during programming: `released` samples 1, firmware lowers
`soc_cs_n`, `grant` fires, and `soc/board_upduino.v` drives pin 16 LOW through `io_ssn` against
the FTDI's active HIGH driver — plus SCK (15) and MOSI (17). Two CMOS push-pull drivers on three
pins, in the COMMON case during programming, not a corner.

This is the same contention class already observed on this board — `cdone` high after a reset that
should have lowered it, a JEDEC id arriving late as `FF EF 70 16`, then a write error. The
mechanism does not close the bug it was written to close.

The claims that overstate it: `board_upduino.v:56-58` ("the two masters are never driving pin 16 —
or the three pins its select gates — at once"), the same sentence in `CLAUDE.md`, and ADR-0135's
"the two masters simply never share a wire."

**2. `soc/pin_lockout.v:31` — the grant is unbounded.** `released` is sampled before driving and
frozen for the whole request, and nothing bounds `want`: it is `!soc_cs_n`, bit 0 of
`rtl/spiflash.v`'s control register, set and cleared only by firmware. A host that starts
programming AFTER the grant is invisible to the frozen synchroniser. Firmware that asserts select
then traps or hangs holds three pins driven forever. And the overlap is the normal workflow, not a
hypothetical: `soc/run_suite_board.sh` flashes while the previously flashed design is still
running (CRESET is not wired), and one of the programs it flashes reads the flash.

Finding 2 amplifies finding 1 — once the grant is wrongly given, nothing takes it back.

### Verified clean, and worth keeping

No combinational loop (`grant = want && released` reads only flop outputs; the `!grant` clock
enable breaks the pad path). The freeze does hold for the whole request. Reset leaves the master
not owning the bus. `test/pin_lockout_tb.v`'s red direction genuinely fails six times at the glitch
phase. `soc/board_elaborate.sh` is tightened, not loosened — `(ERROR|Warning):` cannot match
yosys's `Warnings: N unique messages` summary, which has no colon. `mutate()`'s `cmp -s`
anti-vacuity guard is intact and `PROBES_EXPECTED` is compared after all probes run.

`rtl/spiflash.v`, its two benches, the map work, the window-test extension and the
`board-elaborate` gate are all sound. None of that is what this hold is about.

### Why held rather than merged with a follow-up

Two reasons, and the second is the one that decides it.

The hazard is on real hardware. Nothing on CI or on `make test`'s path drives these pins —
`board_upduino.v` is board-only — so the repo's gates cannot go red for this. Flashing is a live
workflow here, and sustained push-pull contention is not something the iCE40 I/O is rated for.

More importantly, this PR would write a FALSE SAFETY PROPERTY into `CLAUDE.md` and into ADR-0135.
That file is the repo's ground truth and every later change gets reasoned against it. An invariant
recorded as held, which is not held, is worse than no invariant — it is exactly the "tests stay
green while the design rots" failure the design-commitments section exists to prevent.

### Why I am not taking the descope myself

Shipping `rtl/spiflash.v` + benches without wiring it into the board wrapper is the right shape and
I am recommending it. I am not doing it here: it means deleting `pin_lockout.v`, its bench and its
probes, re-deriving `PROBES_EXPECTED`, and re-authoring both the ADR and the CLAUDE.md paragraphs
to remove the arbitration claims. That is re-authoring the PR's argument, not resolving a
mechanical conflict, and it would invalidate the CI I just verified.

### The shape the rework should take

Split in two.

**Ticket A — the master, no shared pins.** `rtl/spiflash.v`, `test/spiflash_tb.v`,
`test/spiflash_model.v`, the memmap and window-test work, `test/asm/spiflash.S`. No arbitration, no
`SB_IO` output enables on the four dedicated pins, no `pin_lockout.v`. Nothing shared is driven, so
no contention is possible. The ADR states the flash is reachable in simulation and explicitly does
NOT claim the board can drive it while a host is attached.

**Ticket B — arbitration, as a design question.** The suggested fix — wire `io_sck`'s discarded
`.D_IN_0()` (`board_upduino.v:142`) in as a host-activity detector, with a sticky bounded inhibit —
is worth building, and I am NOT ratifying it blind. It is a heuristic: SCK toggling detects a host
that is CURRENTLY clocking, not one idle mid-session with CS parked high, which is the exact window
the grant fires in. Take seriously the possibility that no purely-electrical arbitration on these
pins is sound, and that the honest answer is a hard interlock or not sharing them at all. That is
an architectural question this ticket did not answer and cannot answer under merge pressure. It
needs a bench measurement on the real board with `iceprog` running, not a simulation.

ADR-0135's number is left free for whichever lands first.

thejefflarson:
## Follow-up: a second High from re-audit

Pin 14's own output enable was gated on `ssn_in` — the fully synchronised
(two-cycle-stale) copy of pin 16 — so a host pulling chip select LOW got up
to two clock periods (~167 ns at 12 MHz) where the FPGA kept driving pin 14
while the flash's own output driver was already live: two push-pull drivers
on one pin, on every chip-select assertion `iceprog` makes. Turn-on needed
the synchroniser's caution; turn-off did not.

Fixed with an asymmetric enable in the new `soc/miso_share_enable.v`:
`enable = released && now`, where `released` is the two-flop synchronised
copy (gates turn-on) and `now` is the raw, unsynchronised read of pin 16
(kills the enable combinationally, same edge the flash sees, turn-off
only). Verified this creates no loop through the pad — nothing downstream
ever writes pin 16.

`test/miso_share_enable_tb.v` grades it standalone, including the specific
turn-off-edge vector: chip select falls, enable must drop the *same* cycle.
Confirmed this vector goes red against the old synchronised-only spelling
(`enable = released`) before accepting the fix — 3 mismatches, all on the
turn-off vectors, reading the driver still enabled for two cycles after the
pin actually went low.

Narrowed CLAUDE.md and ADR-0135's prose to match: the "does pin 16 read
high" predicate is sound at DC (an SPI slave's output stage is a
deterministic function of its own chip select, regardless of who holds it
there), but a single synchronised copy of that level is stale on the
turn-off edge — that edge is covered by the combinational term, not by the
synchroniser.

Re-ran every gate: `make test` (74/74, 522/522 probes — unchanged, no new
probe-gates.sh target), `make test-units` (14 benches including the new
one), `make fit` / `make soc-timing` (untouched — this change is entirely
in the board wrapper), `make board-elaborate` / `make bitstream` (place
cleanly), `make -C formal check` (86/86), all four component proofs,
`nonperturbation`, and `remeasure-fg` (F=6, G=6, both reproduce exactly).

PR #228, opened 2026-08-24
Route fork PRs off the self-hosted pool instead of skipping

thejefflarson:
Closes JEF-870.

## The problem

`.github/workflows/ci.yml` ran eleven jobs (`elaborate`, `test`, `mutation-check`, `cosim`,
`components`, `lint`, `fit`, `soc-timing`, `ecp5-timing`, `nonperturbation`,
`monitor-freshness`) on `little-cpu-runners`, a self-hosted Kubernetes pool, triggered by
`on: pull_request` with no guard on where the PR's branch lives. The repo is public, and every
one of those jobs runs `make` recipes and shell scripts the PR head wrote. GitHub turns one
maintainer approval of a first-time outside contributor's run into a **standing grant** for that
contributor's later PRs, so approving one run to see green CI hands them durable code execution
inside the cluster. The pods' `runAsNonRoot`/`allowPrivilegeEscalation: false` hardening blocks
host takeover and cross-job persistence — not code execution inside the pod, lateral reach,
exfiltration, or abuse of the `actions/cache` entry that outlives a job. This gap predates this
branch (verified identical to `main` during the #223/#224/#225 pass).

## The fix: route, don't skip

Every job that ran on `little-cpu-runners` now picks its runner with:

```
${{ (github.event_name != 'pull_request' ||
     github.event.pull_request.head.repo.full_name == github.repository)
    && 'little-cpu-runners' || 'ubuntu-latest' }}
```

Same-repo PRs and pushes to `main` keep the fast self-hosted pool; a fork PR runs the **same job,
under the same name**, on GitHub-hosted `ubuntu-latest`. A bare `if:` skip was rejected: it leaves
a fork contributor with no CI signal at all, and a skipped *required* check either blocks the PR
forever or reads as passing depending on branch-protection semantics — worse than the exposure
this closes.

`formal` was already on `ubuntu-latest` (a memory-limit reason, unrelated) and needed no change —
it's the existing proof the toolchain installs and the jobs run there at all.

Five jobs assert the RISC-V cross compiler is baked into the self-hosted image via
`.github/actions/verify-toolchain` rather than installing it (installing on the self-hosted pool
is blocked by `runAsNonRoot`). Each gets one fork-only step, gated on the same condition, that
installs `gcc-riscv64-unknown-elf` (plus `clang` where the assertion also names `clang++`) before
the assertion runs — a no-op on the self-hosted pool.

The expression is repeated on all eleven `runs-on:` lines, with the reasoning stated once in a
comment above `jobs:` and every job pointing back to it, rather than factored through a shared
setup job — a `needs:` edge from all eleven onto one job would stall the whole workflow on it for
a value with no side effect to share.

## Three-case evaluation (also in ADR-0133)

- **`push` to `main`**: `github.event.pull_request` doesn't exist, so `github.event_name !=
  'pull_request'` is `true` and the `||` short-circuits before touching the (absent) right side.
  Self-hosted.
- **Same-repo `pull_request`**: every recent PR (#223–#227) pushed a `thejefflarson/jef-8xx-*`
  branch directly to this repo, so `head.repo.full_name == github.repository` is `true`.
  Self-hosted — unchanged for the pattern this repo actually uses.
- **Fork `pull_request`**: both disjuncts false → `ubuntu-latest`.

`actionlint` passes on the edited workflow with no findings.

## Required checks (verified live, not from a comment)

```
gh api repos/thejefflarson/little-cpu/branches/main/protection -q '.required_status_checks.contexts'
```
→ `elaborate`, `test`, `components`, `monitor-freshness`, `lint`, `formal`, `soc-timing`,
`nonperturbation`, `cosim`, `mutation-check`, `fit` — eleven checks, all still present because
every job kept its name. `ecp5-timing` is the twelfth self-hosted job and is not required; it's
routed the same way anyway, because the exposure has nothing to do with whether a check is
required.

## Docs

- **`docs/THREAT_MODEL.md`**: new "Which runner class a fork PR can reach" note under Trust
  boundaries, naming the self-hosted pool's hardening, what it does and doesn't block, and that a
  fork PR can no longer reach it.
- **`docs/adr/0133-fork-prs-route-off-the-self-hosted-pool-instead-of-skipping.md`**: the decision
  record. 0131 and 0132 are taken on `main`; 0131 is also independently claimed by three still-open
  PRs (#223, #224, #226) against their own pre-`main` trees — a known collision for the architect
  to resolve on merge order. 0133 is genuinely free (checked against `main` and against every open
  PR's file list).

## Human follow-up (cannot be done from a PR)

**"Require approval for all external contributors"** — the repository setting that makes approval
per-run rather than a standing grant — is the other half of this fix and lives in Settings →
Actions → General. No workflow file can set it. Flagged in `docs/THREAT_MODEL.md` and ADR-0133 as
a maintainer action.

## Testing

- `actionlint .github/workflows/ci.yml` — clean.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parses, 12 jobs.
- Manual reasoning through all three trigger cases (above).
- `bash test/retired_term_test.sh .` and `python3 test/band_source_test.py .` — both green (the
  doc additions don't trip either hermetic prose check).
- `gh api .../branches/main/protection` — required-check names confirmed unchanged.
- A workflow trigger's actual GitHub-side behavior can't be exercised without pushing a PR from a
  real fork; this is the acknowledged limit of what's testable pre-merge.

## What I did NOT change

No job's behavior once it runs, on either runner class. No `rtl/` change. No held branches
(`jef-866`/`jef-867`/`jef-868`) touched.


thejefflarson:
Architect review — merging.

Re-verified independently rather than taken from the engineer's report:

- **CI is genuinely green.** All 12 jobs pass on run 32783422615, `mergeStateStatus: CLEAN`. The earlier `BLOCKED` was pending checks, not a failure.
- **Required-check names, both directions.** Read live from branch protection: the 11 required contexts are a subset of the 12 job names, and the job-name set parsed from this head is *bit-identical* to `b2574bc`'s. Nothing silently lost a `name:`. `ecp5-timing` is the twelfth job and is not required — routing it anyway is right, since an unrequired job on the self-hosted pool is still code execution in the cluster.
- **Routing coverage is exact.** Zero bare `runs-on: little-cpu-runners` remain outside the explanatory comment; 11 routed jobs plus `formal` already on `ubuntu-latest`. `head.repo.full_name` appears only in `runs-on:` and `if:` expression contexts, never interpolated into a `run:`. No `pull_request_target`, triggers and `permissions: contents: read` unchanged, and `ci.yml` references **no secrets at all**.
- **Toolchain bridge is exactly scoped.** There are precisely five `verify-toolchain` assertion sites and precisely five fork-only install steps, each immediately preceding its assertion. The `tools:` list matches what each step installs, and `gcc-riscv64-unknown-elf` is the same package `CLAUDE.md` already documents as the Linux path.
- `actionlint` clean; `test/retired_term_test.sh` and `test/band_source_test.py` both green. ADR **0133** confirmed free — `main` tops out at 0132.

## Decision recorded: the fork path ships unexercised, and that is accepted

**A fork PR's rerouted job has not been run and cannot be, pre-merge.** This is asserted, not measured. Saying that plainly rather than implying otherwise:

- Proven: the OSS CAD Suite installs on `ubuntu-latest` — the `formal` job has always done it.
- Proven: `gcc-riscv64-unknown-elf` is the package name this repo already documents for Linux.
- **Not proven:** that the five install steps actually satisfy `verify-toolchain` on a hosted runner, and that hosted-runner speed fits the existing `timeout-minutes` budgets. On the self-hosted pool this run, `components` took 5m08s against a 10-minute budget — the tightest headroom of any routed job, and it has no install step to blame if a slower hosted runner overruns it.

Accepted anyway, because the failure mode is bounded and asymmetric. The added steps are `if:`-false on **every path this repository currently exercises**, so they are dead code for same-repo PRs and pushes to `main` — this change cannot regress the queue. The worst case is that a hypothetical fork PR goes red for an environment reason rather than a code reason: a nuisance, discovered the first time it happens, and strictly better than the standing arbitrary-code-execution grant it replaces. Holding an Urgent security fix to chase an unfalsifiable-until-used path would be the wrong trade.

## Human follow-up (not this PR)

1. **Set "Require approval for all external contributors"** (Settings → Actions → General). This is the other half and no workflow file can set it. Tracked in `docs/THREAT_MODEL.md`.
2. **Ticket: exercise the fork path once** — open a throwaway fork PR after this lands and confirm the five install steps satisfy `verify-toolchain` and that no routed job overruns its `timeout-minutes` on hosted hardware. `components` is the one to watch.
3. **Ticket: two stale comments.** `fit` and `nonperturbation` carry in-file notes saying "Not in main's required set, deliberately"; the live query says both ARE required. Nothing breaks — the file tells readers to trust the live query — but two comments now assert the opposite of reality.

Merged with a normal `--squash`. No `--admin`, no branch-protection bypass.

PR #229, opened 2026-08-25
Retire "master" from the bus vocabulary

thejefflarson:
## Summary

Replaces "master" throughout comments and prose with more precise terminology, per the maintainer's ask:

- **"bus initiator"** for the side of a shared bus that issues a transaction ("single-initiator"/"multi-initiator" for the design shape)
- **"controller"** for an SPI role

**41 occurrences across 22 files** — `CLAUDE.md`, `rtl/`, `test/` and `formal/` — all comments or prose. Confirmed with `grep -rn master`: no signal, port, or module ever spelled the word, so this is a comment-only sweep with no logic changes and no identifier renames. Several sentences were rephrased rather than mechanically substituted (e.g. `CLAUDE.md`'s shared-bus paragraph) to keep the meaning and sharpness of the original, per the repo's comment discipline.

## Two sites fixed directly, not exempted

Per the maintainer's review, these were not left as-is:

- **`.github/workflows/soundcheck.yml`** — `branches: [main, master]` → `branches: [main]`. This repo's default branch is `main`, so the fallback was inert; removing it changes no behaviour.
- **`docs/ideas/finish-the-rewrite.md`** — the riscv-gnu-toolchain link now uses GitHub's branch-agnostic `/blob/HEAD/README.md` instead of `/blob/master/README.md`. Upstream still uses `master`, so swapping the word for `main` would 404; `/blob/HEAD/` resolves to whatever upstream's default branch is and survives a future rename there. Verified it resolves (HTTP 200) before pushing.

## Left as-is, by design

`docs/adr/0125-*.md`, `docs/adr/0126-*.md`, `docs/adr/README.md` — dated decision records. Rewriting one to match a later vocabulary preference misrepresents history the same way moving a measured number in an ADR would.

## No new test machinery

An earlier version of this PR added a third term to `test/retired_term_test.sh` (mirroring the existing `ladder`/`RV32IMC` retired-term mechanism) to guard against regression. The maintainer reversed that decision: that machinery exists for terms that **misdescribe** the design — where the old word is a correctness problem — and a vocabulary preference isn't that. Gating it would mean a permanent allow-list whose granularity is a standing argument, for a failure mode that's just a word reappearing in a comment. So the final diff is the rename plus the two fixes above — no probe delta, no allow-list.

## Testing

- `make test` — 73/73 programs pass, failure list matches `test/EXPECTED_FAIL` exactly
- `make test-units` — all unit benches pass
- `make probe-gates` — 515/515 graded comparisons (main's count, unchanged — no probes added or removed)
- `test/retired_term_test.sh` — unchanged from main, 2 terms (`ladder`, `RV32IMC`) confined to their sites
- `test/band_source_test.py` — green, unaffected
- `make netlist-digest` / `make netlist-diff BASE=origin/main` — **DIGEST-EQUAL**, identical sha256 to origin/main (expected: nothing here is RTL)
- Verified the new `/blob/HEAD/` URL resolves (HTTP 200) before pushing


thejefflarson:
**Merged with a known residue, recorded here because nothing in the tree can go red for it.**

This sweep was correct against the tree it was written on: 41 occurrences, comments and prose only,
no identifiers, `PROBES_EXPECTED` unchanged at 515, and the RTL diff contains no non-comment line,
so the netlist cannot have moved.

It landed second, after #226. **#226 introduces 31 new occurrences of the same word in the same
sense**, across 15 files — so `main` carries them now:

```
Makefile:25             rtl/decoder.v:506       rtl/littlesoc.v:4,22,183
test/COSIM_EXPECTED_FAIL:248                    test/MUTATION_DETECTORS:103
test/asm/loadfault.S:224,230,232                test/asm/riscv_test.h:202
test/asm/spiflash.S:7,28,44                     test/decoder_tb.v:1198,1200
test/memmap_test.sh:150,171,241,406             test/probe_gates.sh:2154,2245,2249,2250,2257,2261,2265
test/spiflash_model.v:22,39,118                 test/spiflash_tb.v:9,286
test/testbench.v:46
```

`rtl/spiflash.v` itself already says "SPI **controller**", so the module and the references to it
disagree.

**This is the exact reintroduction path `test/retired_term_test.sh`'s own header documents** — "the
word was back on main within hours, in a comment written on a branch that predated the sweep and
merged after it, with nothing anywhere to object." Merge order is why, and no care during a sweep
closes it. That is an argument for the vocabulary gate this PR removed at review; it is the
maintainer's call, so it is recorded rather than reinstated.

**A follow-up sweep is not mechanical — one site is coupled to a grader.** `test/probe_gates.sh:2250`
matches the literal substring emitted by `test/memmap_test.sh:171`:

```
"the UART ends at 0x00020028 and the SPI master starts at"
```

Renaming in `memmap_test.sh` without moving the probe's expected text in lockstep turns `probe-gates`
red, and `probe-gates` is a prerequisite of `make test`, a required check.

**Separately, #226 introduced the paired retired term**, which this PR's scope did not cover:
`CLAUDE.md:1026`, `soc/board_upduino.v:39` and the `docs/adr/README.md` row for 0135 all say "an SPI
slave's output stage". Out of scope here by the maintainer's framing ("the bus vocabulary"), so it is
flagged, not swept.

PR #235, opened 2026-08-26
Put Hazard3's iCE40 build in soc/compare/: clock factor measured, CoreMark cycle factor deferred

thejefflarson:
## Status: clock factor shipped, cycle factor deferred

Puts Hazard3's iCE40 configuration in `soc/compare/`: vendored, wired into
the existing three-way harness, and measured for the clock factor. The
CoreMark cycle factor -- the other half of the product ADR-0136 asked
for -- is explicitly **not** built this pass; see "What's deferred" below
for why and what it needs.

### Vendoring

`soc/compare/hazard3_pin.mk` pins `github.com/Wren6991/Hazard3` at
`8af992930f71a69b0e06c38734c1094f41a05ca0`, in `formal/pin.mk`'s exact
shape: 40-hex SHA, `override` so the pin can't be defeated from the
command line, fail-closed re-derivation if a stale clone is on disk at
the wrong SHA, gitignored clone fetched on demand and never committed.
**Licence: Apache-2.0** (Hazard3's own `LICENSE`, no narrowing `NOTICE`),
checked before vendoring.

### The harness

`soc/compare/bench_hazard3.v` instantiates `hazard3_cpu_1port` with
`example_soc/fpga/fpga_icebreaker.v`'s extension/CSR/muldiv parameters
copied verbatim -- `EXTENSION_C=0`, `MULDIV_UNROLL=1`, no bitmanip, no
branch predictor, `CSR_COUNTER=0`, the same 12 MHz target -- behind an
AHB5 adapter onto `rtl/memory.v`. Three named deviations from
`fpga_icebreaker.v`'s own SoC integration (no JTAG debug module, text at
address 0 rather than the SoC's 0x40 bootloader offset, `CSR_M_MANDATORY`/
`CSR_M_TRAP` kept on since the core cannot execute without them) are in
the file's own header comment and in docs/adr/0139.

### The anti-fraud check caught a real bug

`make compare-smoke` is not a formality here: the first working adapter
placed, synthesised at a plausible 0.95x ratio, and published three
all-X words where littlecpu's and VexRiscv's own sequences were real
numbers. The bug: the adapter fed a write's data into `rtl/memory.v` on
the SAME cycle as the address, but AHB5 presents write data one cycle
LATER (the data phase overlaps the next transfer's address phase), and
`soc/compare/bench.S` hits that exact collision on every inner-loop
`sw`-immediately-followed-by-`lw`. Fixed by holding `hready` low for one
cycle after a write's address phase and performing the write against the
captured address once the data is valid. Verified: Hazard3's six
published values now match littlecpu's exactly. This is documented as
its own section in docs/adr/0139, because it's the actual demonstration
that the anti-fraud checks work on the new core, not just that they're
wired up.

### The clock factor (measured)

Five seeds on the existing hx8k comparison harness, `soc/compare/bench.S`
(RV32I only -- shared datapath, not either core's differing extensions),
this tree:

| | worst | median | best | placed `ICESTORM_LC` | standalone `SB_LUT4` | ratio |
|---|---|---|---|---|---|---|
| littlecpu | 30.90 MHz | 32.01 MHz | 32.69 MHz | 7254/7680 (94%) | 6517 | 1.11x |
| Hazard3 (iCE40 config) | 32.60 MHz | 33.63 MHz | 33.89 MHz | 3346/7680 (43%) | 3505 | 0.95x |

Hazard3's iCE40 build clocks **1.06x this core's worst placement, 1.05x
median**. Both anti-fraud ratios clear the 0.80x floor. Five seeds is a
look, not a verdict -- this repo's own convention wants twelve to sixteen
before treating a gap this size as settled.

### What's deferred: the cycle factor

Hazard3's iCE40 build has `CSR_COUNTER=0` -- no `mcycle`, no `minstret`
-- so it cannot self-time CoreMark. The preferred approach (harness-side
cycle counting, changing neither core) is already built once in this
tree for VexRiscv's identical gap: `soc/compare/dhry_port.c` publishes a
marker to a fixed RAM word before/after the measured loop, and
`soc/compare/dhry_tb.v` counts cycles between the two markers on each
core's own bus. The same shape closes Hazard3's gap. It isn't built this
pass because CoreMark is EEMBC's much larger benchmark (ADR-0136:
55.2M cycles at 100 iterations, >100x a Dhrystone run) and getting a
first cross-core placement plus a real, working anti-fraud gate landed
in the time available left no safe margin to also stand up
`coremark_tb.v`, a matching port, and a linker script -- run it right the
first time -- and trust the number. **A wrong cross-core figure is worse
than none, which is the exact position ADR-0136 already took declining
the raw comparison.** docs/adr/0139 records the concrete follow-up plan.

### Testing

- `make compare-timing COMPARE_CORE=hazard3` -- placed, ratio-checked,
  five seeds swept (see table above).
- `make compare-smoke` -- all three cores publish the identical six-value
  sequence off `soc/compare/bench.S` (this is the check that caught the
  write-timing bug above, before any number was quoted).
- `make probe-gates` -- 584 graded comparisons, every failure path
  executed, exit 0. `PROBES_EXPECTED` unchanged at 584 (this diff does
  not touch that count).
- `make test` -- full suite green: 13/13 unit benches PASS, 74/74 `.S`/`.c`
  programs PASS, failure list matches `test/EXPECTED_FAIL` exactly.
- `test/adr_numbering_test.sh` -- 136 ADR files, one row each, no number
  twice.

`test/probe_gates.sh` needed one surgical fix: `ecp5_stale_fixture()`
copies a bare `Makefile` + `formal/pin.mk` to run `make` against, and the
root Makefile's new second `include` line means that fixture now also
needs `soc/compare/hazard3_pin.mk`. One `mkdir` path widened, one `cp`
line added; nothing else in that file touched.

### ADR

`docs/adr/0139-hazard3s-ice40-build-joins-soccompare-clock-measured-cycles-not-yet.md`
records the vendoring, the configuration deviations, the write-timing
defect and fix, the measured clock factor, the decomposition ADR-0136
asked for (so far as it can be stated without the cycle factor), and the
concrete follow-up plan for the cycle factor.

Closes JEF-875


thejefflarson:
Merged at 18dbf60, after a rebase onto a27b954 (#236) and a full re-run of all 11 required checks on the rebased head.

**The WIP status is honest, which is the thing I was checking.** ADR-0139 says what it did not measure in the places a reader would otherwise assume it had: no cross-core product is quoted anywhere, the five-seed clock figure is labelled "a look, not a verdict" against this repo's own twelve-to-sixteen convention, and "Neither core's critical path was walked for this ADR" forecloses the reading that the area column explains the clock gap. The detail that convinced me it is a real record rather than a favourable one is that it discloses the pre-fix adapter reported a *larger* gap (33.08 → 32.60 MHz worst, 34.48 → 33.63 median), so the bug that `compare-smoke` caught had been flattering this side. `geometry_test.sh` staying a two-core check is disclosed rather than quietly left, with the reason — keeping this diff's footprint in `test/probe_gates.sh` at zero while #236 was live on that same file, which was the right call for merge safety.

**Independently re-verified the supply-chain surface**, since this vendors a second third-party core and a self-report is not evidence:

- `8af992930f71a69b0e06c38734c1094f41a05ca0` is Hazard3's `v1.1.1` tag and `stable` head, not an arbitrary commit.
- Apache-2.0 confirmed by reading `LICENSE` at that SHA through the API; no `NOTICE` in the tree to narrow it.
- The pin file is `formal/pin.mk`'s shape verbatim — `override`, 40-hex validation, fail-closed on a stale clone, atomic clone-then-rename, gitignored.
- Hazard3 carries **six submodules** and the clone recipe does not recurse. None sit under `hdl/`, so every source this repo compiles is in-repo and the trusted surface is smaller than a recursing clone's. Worth keeping that way on any re-pin.
- `hazard3_pin.mk`'s claim that nothing executes code out of the clone holds empirically: zero `$system`/`$fopen`/`$readmem*`/plusargs across the compiled `hdl/` files, so iverilog compiling them adds no execution surface beyond RTL semantics.

**`PROBES_EXPECTED` was measured on the merged tree, not computed** — the rebase touched `test/probe_gates.sh` in a different region from #236's and reported clean, which is exactly the silent-wrong-literal case JEF-895 exists for. `./test/probe_gates.sh` on the rebased tree reads **586 graded comparisons, every failure path executed**, exit 0, matching the literal; both of #236's new probes and this PR's `ecp5_stale_fixture()` `cp` line are present. CI's own `test` job agrees.

**Follow-up this owes, beyond the cycle factor the ADR already plans.** `CLAUDE.md`'s Commands block still describes a two-core harness in three places — `make compare-timing` as "this core and VexRiscv in ONE hx8k harness; COMPARE_CORE picks the side", `make compare-smoke` as "both harnesses", and the cross-core bullet as "this core against the VexRiscv Verilog in the pinned riscv-formal clone". All three are now wrong by one core. Descriptive text rather than an invariant, so not a hold, but `CLAUDE.md` is this repo's ground truth and it should not be the last place to learn there are three working points.

PR #236, opened 2026-08-26
Isolate tracked_ignored_test.sh from per-developer git excludes

thejefflarson:
## The bug

`test/tracked_ignored_test.sh` runs `git check-ignore --no-index`, which also consults exclude sources this repository does not carry: a developer's own `core.excludesFile` (their personal `~/.gitignore`) and `$GIT_DIR/info/exclude` (per-clone, never committed). A rule on either one — an ordinary `*.log` or `*.json` line — matched a file this tree tracks on purpose and turned `make test` red on that one machine only. CI was unaffected, which made it harder to see rather than easier.

## The fix

- `-c core.excludesFile=/dev/null` on both the `ls-files` and `check-ignore` invocations neutralizes the first source.
- **`$GIT_DIR/info/exclude` needed its own handling — `--no-index` does not already exclude it.** Measured directly: a pattern written straight into `.git/info/exclude` still matched a tracked path even with `core.excludesFile` already neutralized, because `--no-index` reads `info/exclude` out of whatever `--git-dir` names and needs no object database to do it. So `check-ignore` is now pointed at a throwaway bare git directory (created fresh in `$tmp` on every run) that carries no `info/exclude` of its own, while `--work-tree` stays on the real repo so every actual `.gitignore` rule still resolves and matches exactly as before.

## Tests

Two new probes in `test/probe_gates.sh`'s `test/tracked_ignored_test.sh` group, one per exclude source, each asserting a **green** result under a hostile config (a tracked file matching a rule from `core.excludesFile` / `$GIT_DIR/info/exclude` alone, with no matching rule in the tree). This is the inverted shape from every other probe in the file, so each was proven to actually go red against the pre-fix script: I reverted `test/tracked_ignored_test.sh` to `origin/main`'s version in a scratch copy, re-ran `probe_gates.sh`, and both new probes reported RED (`exited 1, expected 0`, naming the fixture's `.json` file and its rule), then restored the fix and re-ran to confirm both go green again.

The six existing probes for this script are untouched and stay green, including the one demonstrating the original defect (a tracked file matching its own **in-repo** `.gitignore` rule is still caught and named).

`PROBES_EXPECTED` moved from 584 to **586**, read from the script's own mismatch error after adding the two probes, not computed.

Also ran `make test` end-to-end in a throwaway clone with `core.excludesFile` pointed at a fixture naming `*.json`, reproducing the exact false-positive shape the bug report described, and confirmed the suite stays green with the fix in place.

## Scope

Only `test/tracked_ignored_test.sh` and `test/probe_gates.sh` changed. No ADR — this is a bug fix in a grader, not an architectural decision. `make netlist-digest` is unaffected (no `rtl/` change).

Closes JEF-894

### DIFF
--- a/test/probe_gates.sh
+++ b/test/probe_gates.sh
@@ -29,7 +29,7 @@ REPO=$(cd "$HERE/.." && pwd)
 # Pinned as a literal: a probe that is deleted, or that stops being reached by
 # an early `return`, would otherwise cut this file's coverage while it kept
 # printing a green summary.
-PROBES_EXPECTED=584
+PROBES_EXPECTED=586
 
 tmp=$(mktemp -d "${TMPDIR:-/tmp}/littlecpu-probe.XXXXXX") || {
   echo "error: could not create a temporary directory under ${TMPDIR:-/tmp}." >&2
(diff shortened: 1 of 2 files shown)


thejefflarson:
Merged. One residual, recorded here rather than left for someone to rediscover — not a blocker, and strictly smaller than what this closed.

`git init --bare` honours `init.templateDir`, so the throwaway git directory this now points `--git-dir` at can itself inherit an `info/exclude` from a developer's own init template. Measured on this machine, the same way the rest of this PR was measured rather than argued:

```
mkdir -p $W/tmpl/info && printf 'tracked.json\n' > $W/tmpl/info/exclude
git -c init.templateDir=$W/tmpl init -q --bare $W/bare
cat $W/bare/info/exclude   # -> tracked.json
```

That is the third per-developer exclude source, in the same class as the two this PR closed and reachable by the same shape of defect: red on one laptop, green on CI and everywhere else. It is a rarer configuration than `~/.gitignore`, which is why this is a follow-up and not a hold — but the fix is one flag on the `git init` (`--template=` pointed at an empty directory, or `-c init.templateDir=`), plus a third probe in the same group, taking `PROBES_EXPECTED` to 587.

Independently re-verified the three load-bearing claims before merging, since the surprising one is doing the real work here:

- `--no-index` does read `$GIT_DIR/info/exclude` even with `core.excludesFile=/dev/null` — the correction to the ticket is right, and the ticket's one-flag fix would have been insufficient.
- Pointing `--git-dir` at a throwaway bare repo suppresses it.
- A real in-tree `.gitignore` rule still matches through `--work-tree`, so the check did not go blind to buy the isolation.

Also confirmed the shared-worktree config hazard is closed on this machine: `core.excludesFile` is unset locally and `.git/info/exclude` carries only git's default comment block, no rules.

PR #238, opened 2026-08-27
Close init.templateDir as a third per-developer exclude source

thejefflarson:
## What

`a27b954` (#236) isolated `test/tracked_ignored_test.sh` from two per-developer
exclude sources — `core.excludesFile` and `$GIT_DIR/info/exclude` — by pointing
`--git-dir` at a throwaway bare repository with no `info/exclude` of its own.

That throwaway repository is itself built with `git init --bare`, which honours
`init.templateDir`: a template directory carrying its own `info/exclude` gets
copied into the fresh `$GIT_DIR`, reintroducing the exact false positive the
throwaway directory exists to prevent.

## Fix

`--template=` on that one `git init --bare` call, pointed at a freshly-created
empty directory. Per `git-init(1)`'s TEMPLATE DIRECTORY precedence
(`--template` > `$GIT_TEMPLATE_DIR` > `init.templateDir` > the default template
directory), `--template=` is the only one of the three that outranks every
other source — `-c init.templateDir=` would still lose to a developer's own
`$GIT_TEMPLATE_DIR` environment variable, which sits one level above
`init.templateDir` in that same order.

## The exhaustive exclude-source enumeration

`gitignore(5)` documents exactly four pattern sources, highest to lowest
precedence:

1. Patterns passed on the command line to the command doing the matching —
   N/A here, this script never passes any.
2. In-repo `.gitignore` files — the one source this check exists to grade,
   left untouched.
3. `$GIT_DIR/info/exclude` — isolated via the throwaway `--git-dir` (#236).
4. `core.excludesFile` — neutralised via `-c core.excludesFile=/dev/null`
   on both git invocations (#236).

All three non-in-repo sources are now closed, including #3's own supply
chain: a template directory is not itself one of the four documented
*pattern* sources, but it is a way to populate #3 at the moment the
throwaway repository is created, which is exactly the gap this ticket closes.
I also checked that `-c core.excludesFile=/dev/null` suppresses the *default*
global-ignore location git falls back to when `core.excludesFile` is unset
anywhere (`$XDG_CONFIG_HOME/git/ignore`) — confirmed empirically, not a
distinct fifth source. **No fourth exclude source remains that this check can
still inherit.**

## Testing

- Added a third probe to `test/probe_gates.sh`, matching the shape and
  fixture pattern of the two #236 added: a **green** control under a hostile
  `init.templateDir` (a template directory whose `info/exclude` matches a
  tracked path), scoped to that one probe invocation via `GIT_CONFIG_GLOBAL`
  rather than touching any shared config.
- **Proved the fix load-bearing in a fully isolated clone**, the way #236's
  probes were proved: built a fixture repo tracking a file a hostile
  `init.templateDir`'s `info/exclude` also names, ran the fixed script
  (green), then reverted the `--template=` flag in a scratch copy and ran it
  again against the same fixture — red, naming the throwaway repository's own
  `info/exclude` as the source, then restored.
- The original defect (a tracked file matching an in-repo `.gitignore` rule)
  is still caught — its probe is unchanged and still passes.
- `PROBES_EXPECTED` read from the script's own error output after adding the
  probe: **587** (main is 586).
- `make test`, `make test-units`, `make probe-gates` (587/587) all green.
- `make netlist-digest`: could not run — this machine's `nextpnr-ice40`
  binary is broken independent of this change (`dyld: Symbol not found:
  __ZN5boost15program_options3argE`, a stale binary against a newer Homebrew
  boost, dated before this session). No `rtl/` file is touched by this diff,
  so the mapped netlist is unaffected by construction; I could not produce a
  fresh digest to confirm mechanically.

Closes JEF-897.

### DIFF
--- a/test/PROBES_EXPECTED
+++ b/test/PROBES_EXPECTED
@@ -420,7 +420,9 @@ control: a peripheral above the reserved span is accepted
 control: a placement holding the whole core is green
 control: a report clearing the floor is green
 control: a rule from this clone's $GIT_DIR/info/exclude is not this repository's to grade
+control: a rule from this developer's $GIT_TEMPLATE_DIR is not this repository's to grade
 control: a rule from this developer's core.excludesFile is not this repository's to grade
+control: a rule from this developer's init.templateDir is not this repository's to grade
 control: a stamped sweep summarises
 control: a stamped sweep with its reports behind it is binned
 control: a suite matching its manifest is green
--- a/test/probe_gates.sh
+++ b/test/probe_gates.sh
@@ -2914,6 +2914,24 @@ git -C "$d" add -f excluderule.json
 probe "control: a rule from this clone's \$GIT_DIR/info/exclude is not this repository's to grade" 0 \
   "tracked files, none matching a .gitignore rule" "$TI $d"
 
+# A THIRD such source: a developer's own init.templateDir. `git init --bare`
+# copies a template directory's contents -- including an info/exclude of its
+# own -- into the fresh $GIT_DIR the check builds to isolate itself from the
+# CLONE's info/exclude, so a rule sitting in the template reaches that same
+# throwaway repository unless something stops it. GIT_CONFIG_GLOBAL, scoped
+# to this one probe invocation, stands in for that developer's real global
+# config without touching this session's.
+d=$(ti_fixture)
+template=$(new_case)/hostile-template
+mkdir -p "$template/info"
+printf 'templaterule.json\n' > "$template/info/exclude"
+printf 'built\n' > "$d/templaterule.json"
+git -C "$d" add -f templaterule.json
+hostileconf=$(new_case)/hostile-gitconfig
+printf '[init]\n\ttemplateDir = %s\n' "$template" > "$hostileconf"
+probe "control: a rule from this developer's init.templateDir is not this repository's to grade" 0 \
+  "tracked files, none matching a .gitignore rule" "GIT_CONFIG_GLOBAL='$hostileconf' $TI $d"
+
 begin_group "test/march_test.sh"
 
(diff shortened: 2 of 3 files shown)


thejefflarson:
Integration note (architect, INTEGRATE mode).

Rebased onto `d87895b` (#239), which landed first. Two changes made during integration:

**1. Merge collision resolved.** #239 deleted the `PROBES_EXPECTED` scalar this PR bumped to 587. One conflict hunk in `test/probe_gates.sh`, resolved in favour of the manifest: the bump is dropped and this PR's probe label is inserted into `test/PROBES_EXPECTED` instead. The count was **read from the grader on the merged tree, not computed** — `./test/probe_gates.sh` reports **588**.

Worth recording that the new mechanism paid for itself on its first real merge: instead of "expected 588, got 586" it named both missing labels by name:

```
392a393
> control: a rule from this developer's $GIT_TEMPLATE_DIR is not this repository's to grade
393a395
> control: a rule from this developer's init.templateDir is not this repository's to grade
```

**2. A second probe added, for the middle precedence rung.** `GIT_CONFIG_GLOBAL` is honoured only by git ≥ 2.32, so on an older git the `init.templateDir` probe passes having demonstrated nothing — a grader that cannot fail. `$GIT_TEMPLATE_DIR` needs no version support and is also the rung that decides the flag choice.

That flag choice was argued in this PR from `git-init(1)`. It is now **measured** rather than cited:

| delivery | `-c init.templateDir=` | `--template=<empty>` |
|---|---|---|
| `$GIT_TEMPLATE_DIR` | rule still lands | rule blocked |

So `--template=` is the correct spelling and the `-c` form would indeed have lost. Red direction demonstrated: with `--template=` removed from `tracked_ignored_test.sh`, **both** template probes go red, each naming its own rule file.

`make test` green locally (74/74, baseline exact) and all 11 required checks re-verified after the rebase.

PR #240, opened 2026-08-28
Build Zknh's SHA-256 half as a cost probe: +191 cells, marginal opcode costs more

thejefflarson:
Closes JEF-907

## What this is

A measurement, not a feature. Continues ADR-0131's `sh2add` probe and is the first real user of ADR-0132's/ADR-0138's per-encoding completeness exclusion predicate.

**Question**: ADR-0131 measured `sh2add` (adder contact) at +110 packed cells on `littlesoc`, a period null. Is that a flat per-opcode tax, or does it track datapath contact? This probe builds `sha256sig0`/`sig1`/`sum0`/`sum1` — three constant rotates XORed together, no adder, no shifter at all — to find out.

**Answer**: flat, and worse. +240 packed cells for four instructions (60/instruction), and the marginal cost *grows* with each one added (+23, +36, +63, +69 — the fourth is three times the first). `FIT_MAX_LC` trips (4300 against 4219). Unlike `sh2add`, this probe also costs period: 13 of 16 seeds slower (p = 0.021), and one placement of sixteen misses the board's 12 MHz clock where the base held all sixteen.

Full writeup, both encodings' provenance, the formal-exclusion mechanics and the per-instruction breakdown: **[docs/adr/0141](docs/adr/0141-sha256-is-measured-and-the-marginal-opcode-costs-more-not-less.md)**.

## What shipped vs. what was reverted

Per the ADR-0100/0101/0113/0131 convention — build it, measure it, revert it, keep the ADR:

- **Kept**: `docs/adr/0141-*.md` and its `docs/adr/README.md` row.
- **Built, measured, reverted** (three commits: build → record → `git revert`): the decode/execute RTL, `test/decoder_tb.v` vectors, `test/asm/sha256.S`, the `test/OBSERVED_FLOOR` line, and `formal/complete.sv`'s/`formal/COMPLETE_EXCLUSIONS`'s per-encoding exclusion.

`git diff origin/main -- rtl/ test/ formal/` is empty on this branch.

## ADR-0132/0138's predicate: first real user

All four SHA-256 encodings share one `opcode/funct3/funct7` triple (`0010011/001/0001000`), told apart only by the field a normal R-type reads as `rs2` — which the predicate doesn't need a field for, since no riscv-formal-modelled instruction lives inside that triple at all. One exclusion line covers all four; `check-complete-exclusions.py` accepted it on the first attempt (correctly distinguishing our funct7 from SLLI's funct6-wildcarded one at bit 25); `make -C formal complete` passes with the exclusion in and fails at the same assertion ADR-0131 measured without it; `complete_cover` still passes at its original thirteen goals (no new anti-vacuity goal needed, since OP-IMM's own vacuity is already covered by ADDI/SLLI/etc.).

## Testing

- `make fit` incrementally (1, 2, 3, 4 instructions live) on `littlecpu`: 4109 → 4132 → 4168 → 4231 → 4300.
- `make soc-timing` single-seed on `littlesoc`: packed `ICESTORM_LC` 4943 → 5183 (deterministic across all 16 swept seeds, both sides).
- `make netlist-diff BASE=origin/main`: DIGEST-DIFFERENT (sweep owed), pre-place `SB_LUT4` +242.
- Sixteen paired up5k placements (`soc/baseline_sweep.sh`, `SOC_MIN_MHZ=0` so a sub-12MHz seed is recorded rather than aborting the sweep): worst/median/best/spread and per-seed sign test, in the ADR.
- `make -C formal complete` and `complete_cover`: PASS with the exclusion.
- `make test`, `make test-units`, `make probe-gates`: green **after the revert** (74/74 `.S`/`.c` programs; while the probe RTL was present, `make test` was also green except for `test/probe_gates.sh`'s own `check-complete-exclusions` fixture generator, which anchors on the literal pre-change text of `formal/complete.sv`'s `insn_excluded` line and breaks on any real edit to it — resolves automatically on revert, not a defect in the checker itself, which was separately verified standalone).
- `make elaborate-strict`, `make waves` (iverilog leg): clean, no warnings.

## Verdict

The arithmetic really is free — `rtl/executor.v`'s diff was four `assign` wires with no cells of their own, matching ADR-0087's finding that a constant rotate is wiring. What costs is the decode row and op-select slot, because the cone they extend already carries roughly forty other flags and sits in the fetch loop — the same shape ADR-0117 already named. **Zknh (ten opcodes) stays dead on this part.** A bare rotation, apart from the extension, is not the thing that costs.


thejefflarson:
**Architect interim record (pre-merge), so nothing is lost to another interruption.** All numbers on the engineer's own toolchain (OSS CAD Suite 20260811, yosys 0.68+48, nextpnr 0.11-1-g62e659ed), after first reproducing both of the ADR's endpoints exactly: base `fit` 4109, four-flag probe 4300.

**1. The one-flag re-spelling does NOT flatten the curve: `fit` = 4296 vs the four-flag 4300 — a null at the ±50 band.** Built as one `is_sha256` row flag (gated on `instr[24:22]==0`) plus a 2-bit `sha_sel <= instr[21:20]` beside it, executor picking among the four XOR networks off the select. Full suite green with it, 75/75, `sha256.S`'s spec-derived vectors included. Reverted per convention.

**2. Why: the cost is the functions' own LUT floor, not the decode row.** Synthesized standalone for ice40 (no decoder, no fetch loop, nothing else in the module): one σ/Σ function is exactly **32 SB_LUT4 — one LUT per output bit**, ADR-0119's own floor for bitwise arms; all four plus their 4-way select are **187 SB_LUT4**; the sum0/sum1 subset plus select is **96**. Against the probe's whole-core +191 fit cells / +242 pre-place SB_LUT4: the XOR datapath is the whole bill. A constant rotate is wiring; a 3-input XOR is not.

**3. The 'marginal cost grows' reading does not survive.** +23/+36/+63/+69 are differences of single `fit` runs on an instrument whose churn band is ±50; their mean 47.75/instruction is the flat floor 187/4 ≈ 47. The monotone pattern is churn. The *total* is real and intrinsic.

**4. Base 16-seed sweep re-run here reproduces the ADR's base column** (seed 2 worst: 82.89 ns / 12.06 MHz, packed LC 4943 — identical), so the engineer's sweep is validated, not just trusted.

**Verdict direction: ADR-0141's bottom line (Zknh dead on this part) STANDS, on harder ground — an intrinsic one-LUT-per-bit floor cannot be re-spelled away. Its mechanism paragraph (decode row / op-select cone congestion) and the growing-marginal claim do not survive and will be corrected by an Amendment 1, not a rewrite.** Amendment + merge to follow.

thejefflarson:
**Architect interim record 2: the one-flag variant's period, 11 of 16 paired seeds (labelled as eleven — a hair short of the repo's 12–16 go/no-go convention, posted now so it survives interruption).**

Same toolchain and stamps as the base sweep (OSS CAD Suite 20260811); base is my own 16-seed re-run on `ab5af01`, which reproduces the ADR's base column exactly (worst 82.89 ns / 12.06 MHz, lc 4943).

| | base (11 common seeds) | one-flag (11 seeds) | Δ |
|---|---|---|---|
| worst | 82.89 ns, 12.06 MHz | 80.73 ns, 12.39 MHz | **−2.61%** |
| median | 76.53 ns, 13.07 MHz | 77.42 ns, 12.92 MHz | +1.16% |
| best | 74.79 ns, 13.37 MHz | 75.59 ns, 13.23 MHz | +1.07% |
| slower | — | 6/11 seeds | sign test p = 1.000 |
| under 12.00 MHz | 0/11 | **0/11** | — |

**The one-flag spelling of the same four instructions is a period NULL where the four-flag spelling was not** (four-flag: 13/16 slower, p = 0.021, one seed at 11.99 MHz). Its worst placement is *faster* than base's own worst. So the four-flag period signature was a property of that text, not of the instructions — the same spelling-dependence ADR-0106 measured at 3.3% between two texts of one design.

**Area does not move, in either spelling**: fit 4296 (one-flag) vs 4300 (four-flag), and `littlesoc` packed lc 5259 vs 5183 — the two tops disagree at the band, which is churn per ADR-0094/0097. The +\~190–240-cell floor is intrinsic (one LUT per output bit × 4 functions, measured standalone at 187 SB_LUT4), so **`FIT_MAX_LC` still trips at 4296 and the area verdict stands regardless of spelling; the period half of ADR-0141's verdict does not**.

Five tail seeds (11–15) still to land; will be reported labelled as sixteen if they complete. RTL will be reverted per convention either way.

thejefflarson:
**Architect final measurements (supersedes the 11-seed interim; all RTL reverted after this).**

**1. One-flag variant, all four instructions, sixteen paired seeds — a period NULL:**

| | base (16) | one-flag (16) | Δ |
|---|---|---|---|
| worst | 82.89 ns, 12.06 MHz | 80.73 ns, **12.39 MHz** | −2.61% |
| median | 77.00 ns, 12.99 MHz | 77.30 ns, 12.94 MHz | +0.39% |
| best | 74.79 ns, 13.37 MHz | 75.59 ns, 13.23 MHz | +1.07% |
| slower | — | 9/16 | sign test p = 0.804 |
| under 12.00 MHz | 0/16 | **0/16** | — |

(My base column reproduces ADR-0141's base sweep seed-for-seed — same toolchain, same worst/median/best — so the two candidate sweeps are directly comparable.) The four-flag text's period signature (13/16, p = 0.021, one seed 11.99) **was the spelling, not the instructions**: a spelling of the same four Zknh instructions exists that holds 12 MHz at all sixteen placements with its worst seed faster than base's.

**2. Area is intrinsic and unchanged by spelling**: one-flag fit 4296 vs four-flag 4300 (null); the floor is one LUT per output bit per function (32/function standalone, 187 for four + select). `FIT_MAX_LC` 4219 trips either way.

**3. The sum0/sum1 subset (the compression loop's Σs): fit = 4225 — 6 cells over the 4219 budget**, against a standalone floor of 96 LUTs. It consumes the entire remaining headroom and lands within re-mapping churn of the ratchet line: not affordable without raising `FIT_MAX_LC` for cause or harvesting cells somewhere the area ceilings say is already closed. No separate period sweep taken for the subset; the all-four one-flag null at sixteen seeds bounds it (strictly smaller change, same spelling) — an inference, labelled as one.

Amendment to ADR-0141 and merge decision follow.

thejefflarson:
**Architect final: ratchet recommendation, HSM ranking, and disposition.** (Amendment 1 recording all second measurements is already pushed as `40a28ae` — the original Decision text is untouched, per convention. Auto-squash-merge is armed on the required set.)

## The ratchet question

**Do not raise `FIT_MAX_LC` now.** The subset's 4225-vs-4219 miss is 6 cells on an instrument with a ±50 churn band, so it is not a pass/fail measurement in either direction — and it is also toolchain-dependent by up to ~54 cells (the budget is graded on the CI job's suite; 4225 was measured locally). What the number does establish is that the subset consumes essentially all of the 110-cell headroom, and a ratchet must sit outside the band relative to the shipping count — so shipping it *requires* moving the budget, to roughly **4330** (4225 + a span re-derived on that tree, per the `FIT_MAX_LC` derivation rule), with an ADR naming the purchase. That raise would be legitimate — the ratchet exists to catch *unexplained* drift, and this drift would be named and measured — **but only when something the product needs buys it, and today nothing does** (next item). A ratchet raised for a capability nothing can use is headroom spent on inventory.

## Is Zknh the binding constraint on the HSM? No — third, and the wrong lever even at third.

1. **ROM is binding, by an order of magnitude.** 8 KB is a hard ceiling with no size between it and a 16 KB that does not fit the part (ADR-0135); `rvc.S` at 12,256 bytes already cannot run on silicon. `ed25519-dalek` alone is tens of KB of rv32 text; the full stack is several times the whole ROM. Flash boot is simulation-only and the SPI controller's pins are not wired to the board — reaching them owes the real-board contention measurement ADR-0135 defers. Until code loads, no ISA extension changes anything.
2. **Entropy is second, and it is functional, not performance.** No TRNG; `Zkr`'s `seed` CSR is an interface, not a noise source. Deterministic Ed25519 *signing* with an injected key works without one; key generation and `snow`'s per-handshake X25519 ephemerals do not. An HSM that cannot make an ephemeral key cannot run Noise at all.
3. **Performance is third, and there Zknh's SHA-256 half is the wrong half.** Ed25519 hashes with **SHA-512** — these four instructions never execute in it; RV32's SHA-512 half is six different, wider instructions. Ed25519's dominant cost is 255-bit field arithmetic built on MUL/MULHU — the shipped M extension with its parallel multiplier is the accelerator that matters. ChaCha20 wants 32-bit rotates (Zbb `rori` territory); Poly1305 wants M again. The SHA-256 four accelerate HMAC/HKDF and the handshake hash — small, infrequent slices — and the software fallback is already constant-time under Zkt, so they buy speed only, no security property.

**The HSM direction is not dead on this measurement.** It is blocked on ROM and entropy, both platform work; SHA acceleration is a later, optional purchase whose price is now known exactly (96 LUTs for the Σs, ~187 for all four, period-free in the right spelling).

## Disposition

- All probe RTL reverted, probe branch deleted; the PR diff is docs-only (`docs/adr/0141-*`, `docs/adr/README.md`). Both the main checkout and the worktree are clean; nothing leaked.
- Merge decision: **MERGE** — the record is correct with Amendment 1 attached; verdict stands on area, on harder ground; the period half and the mechanism paragraph are corrected on the record, not rewritten.

thejefflarson:

re-run green.

PR #241, opened 2026-08-29
Makefile: cut comments to tripwires/pointers, audit for dead weight

thejefflarson:
## Summary (updated -- this PR grew well beyond its original scope)

This started as "move three long comment derivations to ADRs" and grew into a full pass at CLAUDE.md's comment rule applied to the Makefile, plus an audit for genuinely unused targets/variables, after review found the first pass's reduction (1911 -> 1838 lines) far short of what was wanted.

### Comments: 654 -> 162 top-level lines (-75%)

Every remaining comment is one of:
- a **tripwire** naming a specific fixed defect or Make-semantics footgun (a backslash split inside a quoted yosys script that dies differently across `make`s, an order-only clone prerequisite, `.DELETE_ON_ERROR` + `|| true`, a **shell-injection guard** on `netlist-diff BASE=`, a once-shipped wrong-number flags-quoting bug),
- a **one-line ADR pointer** for a derived number (`ADR-0142`/`0143`/`0144`, unchanged from the previous round -- just renumbered),
- or a **one-line label** where a target's name doesn't say what it does.

Everything else -- rationale, history, restatement of what a target's own name or a script's own header comment already says -- is gone. Most of it duplicated CLAUDE.md's own **Commands** section almost verbatim (every `test`-path grader had its own "hangs off `test` because it needs no cross compiler..." paragraph, repeated ~15 times with minor variation); CLAUDE.md is the canonical doc for those, not a second copy in the build file.

**Left completely untouched, per explicit instruction to stay out of concurrent work**: the one comparison-harness comment line already fixed in the previous round, and the `SOC_EXPECT_SPRAM`/`SOC_EXPECT_EBR` block (including its own comment) -- both regions two other engineers are actively editing. This freeze is why the final numbers (1344 lines / 162 comments) land above the ~950/~100 rough targets: the frozen region alone is 282 lines, 86 of them comments.

### Audited for dead weight -- found none among targets, one among variables

Cross-referenced all **146 real targets** (every `.PHONY`, every file target with its own recipe) against four sources of truth: `.github/workflows/*.yml`, CLAUDE.md's Commands section, `make test`'s own prerequisite chain, and every `make <target>` call inside `test/*.sh`, `test/*.py`, `soc/*.sh` and `soc/compare/*.sh`. **Every target has a caller somewhere in that set.** Two judgment calls, both kept:
- The **hardware-flashing pathway** (`bitstream`, `prog`, `board.*`, `suite-board`, `dhrystone-board`/`-rom`, `noop-rom`, `ftread`) isn't named in CLAUDE.md's fenced Commands block, but it's real product functionality -- flashing the physical UPduino this whole project targets, documented at length in CLAUDE.md's surrounding prose and the maintainer's own notes on using `make prog`. Not cruft.
- `print-toolchain` is likewise unreferenced by any script, but `.github/actions/setup-oss-cad-suite/action.yml`'s own failure message tells a human to run it.

Audited all **121 variables** the same way (usage count + where defined). Nearly every apparently-single-use one is a deliberate `?=` override point (several named in CLAUDE.md itself -- `DHRY_RUNS`, `SOC_SEED`, `PROG`) or a named ratchet/census value referenced by name in prose elsewhere (`FIT_MAX_LC`, `SOC_MIN_MHZ`) -- inlining either would delete real, intended functionality. Exactly **one**, `ELABORATE_STRICT_OUT`, had no override use anywhere; inlined it (zero `make -n` effect, since the expansion is textually identical either way).

Evaluated collapsing the near-duplicate cross-compiler-detection boilerplate shared by `soc-rom`, `dhrystone-rom` and (frozen) `compare-rom` into one canned recipe. **Declined**: Make's `define`/backslash-continuation rules are an easy way to silently break a working single-shell recipe, and `soc-rom` sits on `make test`'s path -- not worth ~10 saved lines on functionally important targets.

## Testing

- **`make -n` byte-identity sweep across all 146 real targets**, captured before this change and re-captured after (twice -- once after the comment purge, once more after the final tripwire fix-up): **zero differences**, byte for byte, on every target. This is the load-bearing proof that only comments (and the one inlined variable, whose expansion is identical) moved -- every recipe, every variable assignment, every dependency list is unchanged.
- `make test`: green end to end. 588/588 probe-gates comparisons (every failure path executed), 74/74 `.S`/`.c` suite programs, `test/adr_numbering_test.sh` clean (140 ADR files, no collisions), failure list matches `test/EXPECTED_FAIL` exactly.
- `/soundcheck:pr-review`: no Critical/High findings. Specifically checked that no comment stating an actual security *constraint* (as opposed to just explaining one) was dropped -- the `netlist-diff BASE=` shell-injection guard's comment is deliberately kept; the SHA-256/path-traversal checks in `sail-setup`/`lint-setup` are enforced by recipe code (untouched, proven by the zero-diff sweep) with their own inline error messages, not by the prose comments that were removed above them.
- `/simplify`: direct self-review (comment-only diff, not application code); no further changes found.

## Before/after (cumulative, from this ticket's start on `main`)

| | before | after |
|---|---|---|
| total lines | 1911 | 1344 (-29.7%) |
| comment lines | 727 (38.0%) | 162 (12.1%) |
| blank lines | 159 | unchanged |
| recipe/variable lines | 1025 | unchanged (proven by the `make -n` sweep) |

Closes JEF-910


thejefflarson:
Renumbered the three new ADRs from 0141-0143 to 0142-0144: PR #240 (open, auto-merge armed) claims 0141 for the Zknh SHA-256 cost probe and branched first, so this PR moves out of its way. Updated the README rows, the three Makefile pointer comments, and the one cross-reference between the two Makefile-derivation ADRs. `git mv` used throughout so history follows. Re-verified `test/adr_numbering_test.sh` (140 files, no collisions) and byte-identical `make -n` output post-renumber.

thejefflarson:

I re-derived the central claim independently rather than reading the report. Extracted `origin/main`
and this branch into two clean trees, enumerated targets from `make -pn` on each (100 by my
extraction; the set is identical both ways), ran `make -n` for every one of them in both trees
capturing stdout, stderr and exit status, and diffed after normalising only the tree path.
**Zero differing targets.** The behaviour of this Makefile is byte-identical to main's.

The substance also checks out directly: stripping comment-only and blank lines from both versions
leaves exactly one difference, the documented one —

```
-ELABORATE_STRICT_OUT ?= /tmp/elaborate-strict.cc
-	yosys -p '... write_cxxrtl $(ELABORATE_STRICT_OUT)'
+	yosys -p '... write_cxxrtl /tmp/elaborate-strict.cc'
```

Nothing else in the repository references `ELABORATE_STRICT_OUT`, so the inlining loses only an
unused `?=` override hook, and it is a two-way door.

Spot-checked that the deletions kept what CLAUDE.md's comment rule says to keep: `FIT_MAX_LC`'s
tripwire sentence survives verbatim, `SOC_MIN_MHZ` and `FIT_MAX_LC` both still carry their pointers,
and the three ADRs are cited at the sites the derivations came out of.

This PR is executing a rule CLAUDE.md states about itself — comments earn their place, delete
restatement and history and section banners — and moving three derivations to ADRs is the right
destination for them, since an ADR is a dated record and a Makefile comment is not.

**Merge order.** This should land before #243: both touch the Makefile, and #243 adds
`compare-coremark` plus `COMPARE_COREMARK_*` in the comparison-harness region this PR deliberately
left alone. #243 is held on separate grounds, so that ordering costs nothing.

re-run green. Note that whichever of #240/#241/#242/#244 lands first will leave the others needing a
one-line rebase of the `docs/adr/README.md` row — they all insert at the same anchor. That
resolution is graded rather than trusted: `test/adr_numbering_test.sh` goes red on a row without a
file or a file without a row, in both directions, and it runs inside the required `test` check.

thejefflarson:
Held — not on CI (the required set is untested here only because the branch is behind), but on two content findings. The bulk of this PR is right and sanctioned: `CLAUDE.md` does say to delete restatement, history, section banners and emphasis furniture, and most of the 643 lines are exactly that. Two things in it are not.

**High — four deleted comments are tripwires, which is the one exception the rule carves out.**

`CLAUDE.md`: *"Comments earn their place or go. ... Delete restatement, history (git has it), section banners and emphasis furniture. ... One exception: tripwires — a two-sentence warning that stops a fixed defect from being silently reintroduced."*

Checked against the resulting file rather than the diff, so these are gone rather than reworded:

- `Makefile:555` (on main) — *"A `.gitignore` rule never applies to a file git already tracks, so a tracked file matching one is always a mistake."* This guards a defect class fixed **twice in the last five merges** (#236, #238). Deleting it is the precise case the exception exists for.
- `Makefile:15` — *"Both sim legs build from this list. Do not copy it anywhere else — a second copy goes stale and the gate then checks a different design than it says it does. The CI job had one; it missed rtl/imemory.v and rtl/memory.v when they..."* — names the defect it caught, in the comment.
- `Makefile:818` — *"Count logic cells from nextpnr, never cell counts from yosys."* — the ADR-0112 finding that `SB_LUT4` and packed `ICESTORM_LC` disagree in magnitude and in sign.
- `Makefile:1551` — *"THE DIGEST REPLACES A SWEEP, NEVER A GATE."*

Fix: restore these four. Terser is fine; the warning and its reason are what must survive.

**Medium — two comments are replaced by a bare ADR pointer, which the rulebook names directly.**

`CLAUDE.md`: *"**No ADR numbers and no invariant numbers in comments** — state the mechanism instead; `git blame` finds the record."*

- `# DHRY_CFLAGS: ADR-0143.`
- `# iterations and this exact COREMARK_CFLAGS: ADR-0144.`

These say nothing the code does not, and cannot be read without leaving the file — failing both halves of *"a comment must say something the code does not, in one or two plain sentences readable without leaving the file."* The PR title's model, "cut comments to tripwires/**pointers**", is where this comes from: a pointer is not one of the two permitted shapes.

Not flagged: `# 4219 = 4097 + 68 + 54, each term measured and cited: see ADR-0142.` states the arithmetic before it points, and the `ADR-0139` line states the mechanism (*"has no mcycle to self-time a Dhrystone run with"*). Drop the numbers from both if you like, but neither is a bare pointer. The three `ADR-` strings already on main are `@echo` **recipe output**, not source comments — user-facing text at the moment someone is about to quote a number — so they are not precedent for this.

Also note, mechanically: this branch conflicts with `main` on `docs/adr/README.md` (append-vs-append). #248 has since landed a `merge=union` driver for that file, but it does **not** fix this one — `.gitattributes` is added *by* the merge, so it is not in the working tree when git picks a strategy. Re-anchor the row by hand once; subsequent merges get the driver. Numbers 0142/0143/0144 are free and uncontested.

No `-march=` value moves in this diff, so `test/march_test.sh` is unaffected — but note that #250 has since changed its exception format to `<path> <ISA> <count>`, so rebase after it lands.

thejefflarson:
Addressed the hold: restored the four tripwires verbatim from \`main\` (SIM_RTL_SRCS's "second copy goes stale" defect, the \`.gitignore\`-vs-tracked-file class fixed twice in #236/#238, ADR-0112's \`SB_LUT4\`-vs-nextpnr finding, and the netlist digest's "sound in one direction only"), and replaced the two bare \`ADR-0143\`/\`ADR-0144\` pointer comments with one-line mechanism statements (why 2000 Dhrystone runs, why 100 CoreMark iterations) per CLAUDE.md's "state the mechanism instead" rule — also applied the same fix to the FIT_MAX_LC pointer (\`see ADR-0142\`), which wasn't named explicitly but is the identical violation.

Left the one \`(ADR-0139)\` citation inside the frozen cross-core-comparison-harness section alone — it predates this correction and touching it means editing that section again, which stays off limits while JEF-906 is still working there. Flagging this as a known, deliberate exception rather than fixing it silently.

**Rebased onto \`main\` (#250)**, which brought in #248's \`merge=union\` driver for \`docs/adr/README.md\`. The union merge kept both the pre-renumber and post-renumber rows for this branch's own three ADRs (a rebase replaying two of this branch's own commits, not two branches colliding — but the same known gap ADR-0150 names). Removed the three stale rows; \`test/adr_numbering_test.sh\` is clean (142 files).

Re-verified after every change: \`make -n\` swept across all 146 real targets against the freshly-rebased \`main\` — zero differences. \`make test\` green end to end: 590/590 probe-gates (grew from 588 as main's own probe count moved), 74/74 suite programs, ADR numbering clean.

**Final counts**: 1911 → 1366 lines from this ticket's start (-28.5%), 727 → 184 comment lines (-74.7%).

Force-pushed the rebase to the same branch.

thejefflarson:
Re-checked after the restore commit (`7fb2c0c`).

**The four tripwires are back and verbatim** — the copy-the-list warning, the "a `.gitignore` rule never applies to a file git already tracks" note, "count logic cells from nextpnr, never cell counts from yosys", and "THE DIGEST REPLACES A SWEEP, NEVER A GATE". That hold is closed.

Two things still stand between this and a merge.

**1. One comment still carries a bare ADR number.** `main` has zero ADR references in `Makefile` comments; this branch introduces one:

```
Makefile:1091
# has no mcycle to self-time a Dhrystone run with (ADR-0139) and compare-dhrystone
```

CLAUDE.md: *"No ADR numbers and no invariant numbers in comments — state the mechanism instead; `git blame` finds the record."* The mechanism is already stated in full right beside it ("Hazard3's iCE40 configuration sets `CSR_COUNTER=0`, so it has no `mcycle` to self-time a Dhrystone run with"), so `(ADR-0139)` is carrying nothing — deleting the parenthetical loses no information. The `@echo` lines at 735/738/1329 are program output to the reader, not comments, and are fine as they are.

**2. Still `CONFLICTING` against `main`.** Worth holding the rebase until the other `Makefile`-touching PRs land — #243 (CoreMark targets) and #255 (`geometry_test.sh`'s derived file list) both edit it, so rebasing now buys a second conflict. I am sequencing those first; re-push after they merge and I will re-check.

No content objection beyond the one line.

PR #242, opened 2026-08-30
Price Zbb's rotate, the extension the stack actually wants

thejefflarson:
## Summary

Continues ADR-0141's Zknh SHA-256 cost probe with the extension this stack's own workload names — ChaCha20's quarter-round and software SHA-256/SHA-512 are add-xor-rotate, unlike the SHA-256 four-instruction subset the prior probe found dead here. `ror`/`rol`/`rori` were built, decoded, executed and measured as a probe, then reverted per the ADR-0100/0101/0113/0138/0141 convention. This PR is **docs-only** — `git diff origin/main -- rtl/ test/` is empty.

## What was measured

- **The existing shifter cannot serve a rotate.** Its 33-bit fill carries only one bit (a sign or a zero); a rotate needs up to 31 from the word's own other end. Rotate needs its own barrel network.
- **Standalone synthesis** (ADR-0141 Amendment 1's method): ROR alone (shared outright by `rori`) is 192 `SB_LUT4`; ROL alone via the reversal trick is *identically* 192, confirming ADR-0087's "a reversal is wiring" for a variable-amount rotate; both directions with a runtime select is 280; the "obvious" reuse of the existing SRL/SLL shifter — two shifts OR'd — measures **worse**, 332, not a saving.
- **`make fit`**: 4347 against a 4219 budget, 128 cells over. Unlike SHA-256, no subset is a near miss — `rori` alone already needs the full 192-cell network, clearing the ~110-cell headroom by 82 on its own. There is no ISA-legal single-instruction Zbb subset cheaper than the whole barrel network.
- **Sixteen paired seeds**: period is a null — 12/16 slower (sign test p = 0.077, short of significance), worst placement 2.3% *faster* than base's own worst, both sides hold 12 MHz at all sixteen placements.
- **Zkt**: `test/zkt_isolation_test.py` passes with both new op flags in the netlist — rotate would stay Zkt-clean if it were ever affordable.

## Verdict

Zbb rotate — full set or any subset — is not affordable within the current `FIT_MAX_LC` ratchet. Area alone is the wall; the period was never the obstacle. `FIT_MAX_LC` is not raised.

## Testing

- `make test-unit-decoder_tb` / `make test-unit-exec_tb`: pass, plus a throwaway directed bench cross-checking five concrete rotate vectors
- `make -C formal components_decoder` / `components_executor`: pass by k-induction, including both tasks' `-zkt-probe` prerequisites, with two new independent formal reference assertions for the rotate arithmetic
- `test/zkt_isolation_test.py`: pass
- `make lint`, `test/adr_numbering_test.sh`, `test/band_source_test.py`, `test/retired_term_test.sh`: pass
- `make fit`, `make soc-timing`, standalone `yosys synth_ice40` on isolated rotate modules, and a sixteen-seed paired `soc/baseline_sweep.sh` sweep (base vs. rotate) — all reported in the ADR

## ADR numbering

Per the coordinator's central allocation for the four in-flight PRs, this uses **ADR-0147** (0141/0142-0144/0145/0146 reserved for other tickets).

Closes JEF-909.


thejefflarson:

re-run green.

PR #243, opened 2026-08-30
CoreMark's cycle factor lands for Hazard3; VexRiscv cannot run it

thejefflarson:
## Summary

Closes JEF-906. Builds the CoreMark cycle-factor mechanism ADR-0139 explicitly deferred: `soc/compare/coremark_tb.v` / `coremark_compare_port.c` / `coremark.lds` mirror `soc/compare/dhry_tb.v`'s marker-counting split, timing Hazard3's iCE40 configuration (`CSR_COUNTER=0`, no `mcycle`) and littlecpu on their own bus.

**DECISION (see ADR-0146 for full reasoning): VexRiscv is not a third DUT.** The pinned riscv-formal clone's VexRiscv build is RV32IC with no M extension at all — not slow, absent — so it cannot execute CoreMark's multiply/divide-heavy matrix and state-machine benchmarks. Dropping the shared image to RV32IC (the way `COMPARE_DHRY_CFLAGS` works around VexRiscv's gap for Dhrystone) would erase littlecpu's and Hazard3's own hardware multipliers, which is exactly what this pair exists to compare.

## Measured, one tree, one session

- **Cycle factor**: Hazard3 takes **1.491×** littlecpu's cycles for the same CoreMark work (1 iteration, RV32IMA). Both cores' list/matrix/state CRCs match EEMBC's published 2K-performance values; their 16 KB data RAMs are bit-identical afterward. CoreMark/MHz: 2.086 (littlecpu), 1.399 (Hazard3).
- **1 iteration is exact, not undersized**: CoreMark's `iterate()` is the identical seed-determined call pair every time with no cache on either core; verified linear at 2 iterations (ratio reproduces to 3 decimals).
- **Clock, re-taken on this tree in this session** (not reused from ADR-0139's stale run): littlecpu 30.90/32.01 MHz worst/median (reproduces ADR-0139 exactly), Hazard3 32.07/32.45 MHz (a different 5-seed sample).
- **Product — the first for this pair**: littlecpu's CoreMark score is **1.437×–1.471×** Hazard3's iCE40 configuration's.
- **Bonus, isolated on littlecpu alone** (Hazard3 can't decode compressed instructions at all, so an rv32imac build can't be safely run on it in the shared harness): the C extension costs **2.37% more cycles** on this CoreMark build while shrinking `.text` **29.4%** — the opposite of "smaller code is faster code" on this pipeline.
- **Found and fixed along the way**: Hazard3's regfile is `RESET_REGFILE=0` in this harness, so `x0` is never written by hardware and stays X in simulation unless zeroed at time 0 — the same class of catch ADR-0139's AHB write-timing bug was.

## Tests

- `make probe-gates`: 607/607 (was 588; added 19 new probes for `coremark_fit.py`/`coremark_dmips.py`, mirroring `dhry_fit.py`/`dhry_dmips.py`'s own groups exactly — each demonstrated red).
- `test/march_test.sh`: passes (added two exceptions for `Makefile rv32ima` and `soc/compare/run_coremark_compare.sh (empty)`, mirroring the existing `rv32ic`/`run_dhrystone.sh` entries).
- `test/adr_numbering_test.sh`: 138 ADR files, no collisions, ADR-0146 reserved per the coordinator's numbering allocation.
- `make compare-smoke`: unaffected, still PASS (all three cores agree on `bench.S`).
- `make compare-coremark`: ran end-to-end multiple times, reproducing the same numbers.
- Manual security self-review (soundcheck:pr-review reported not directly invocable in this subagent context): no network input, no shell injection, no secrets in the new shell/Python/C/Verilog tooling.

## Scope / risks

- No `rtl/` file changed.
- `soc/compare/sweep.sh`'s default `COMPARE_CORES` is left as `"littlecpu vexriscv"` — not changed by this ticket.
- CLAUDE.md's Hazard3 CoreMark bullet and the comparison-harness command list are updated in this same commit to state the new numbers; the Makefile's own comparison-harness *prose comment* near `compare-timing` (flagged as PR #241's neighbourhood) was **not** touched.
- DECISION NEEDED / already resolved: see "VexRiscv is not a third DUT" above — flagging for the architect's awareness since it's a interpretation of the ticket's own suggestion, not a literal implementation of it.


thejefflarson:
## HELD — two items must land before this merges; two are follow-up tickets

The mechanism here is good work and the ADR is unusually honest about its own limits: the VexRiscv
exclusion is argued rather than asserted, the 1-iteration sample is shown exact rather than claimed
stable, the `x0`-on-reset catch is a real one, and the `RESET_REGFILE` question is answered the right
way — `coremark_tb.v:107` zeroes `dut_haz.core.core.regs.real_dualport_noreset.mem[i]` from the
testbench by hierarchical path, so Hazard3's pinned RTL is untouched and ADR-0139's "run the pinned
Verilog unmodified" discipline holds. I confirmed that from the diff: no file under the Hazard3 pin,
and `soc/compare/hazard3_pin.mk` and `bench_hazard3.v` are both unchanged.

What holds it is not the mechanism. It is that the number the mechanism publishes is biased, the
bias is larger than the precision the number is quoted to, it runs in this project's own favour, and
neither the ADR nor CLAUDE.md says so.

### 1. Blocking — `soc/compare/bench_hazard3.v:205` with `soc/compare/coremark_tb.v:114,128`

**Problem.** The harness's two slaves do not have the same write timing. `bench_hazard3.v:205`
(`assign hready = !wr_pending_q;`) holds `hready` low for exactly one cycle after every write address
phase — correctly, and for the reason its own comment gives, since AHB5 presents `hwdata` a cycle
after `haddr` and `rtl/memory.v` has one port. But `bench_littlecpu.v` drives `bus_wait(1'b0)` and
its stores cost nothing extra. So **every Hazard3 store in this comparison is charged one bus cycle
that littlecpu never pays**, and that cycle lands in the numerator of the published 1.491×.

Sizing it from the ADR's own figures: at roughly 15.7k stores (littlecpu's count for the same
program) against Hazard3's 714984 cycles, that is ~2.2% of Hazard3's measured cycles belonging to
the harness rather than to the core. 1.491× is really nearer 1.459×, and the product 1.437×–1.471×
moves with it. The bias is about 2.2%; the figures are quoted to four significant digits.

**And the one printed fact that could bound this is broken in the same direction.**
`coremark_tb.v:114` counts littlecpu's stores as `if (|dut_ours.mem_wstrb)` — any strobe — while
`:128` counts Hazard3's as `if (dut_haz.mem_wstrb_mux == 4'b1111)` — full words only. So
`writes=15701` and `writes=13350` are not the same measurement, `haz_writes` under-counts by however
many sub-word stores CoreMark makes, and a reader who tried to bound the harness penalty from the
printed line would under-estimate it. (`bench_tb.v:50,55,60` uses `== 4'b1111` symmetrically on all
three cores, so this asymmetry is new here, not inherited.)

The CRC and `ramdiff=0` gates do not touch this. They say both cores computed the right answer; they
say nothing about which cycles were charged to whom.

This matters more than it would in an ADR alone because the same commit writes 1.491× and the
1.437×–1.471× product into **CLAUDE.md**, which is the file every later reader and agent treats as
ground truth — three lines above the paragraph being edited, that file already says the only
cross-core comparison that means anything is one harness, "same part, **memories**, program,
toolchain and seeds." The memories are not the same here in a way that costs one core cycles.

**Fix — either is acceptable, and the cheap one is fine:**

- *Minimum, and I would take it:* make the two write counters symmetric (`|dut_haz.mem_wstrb_mux`),
  and add the wait-state asymmetry to ADR-0146 as an explicit distortion with its size, the way
  ADR-0098 enumerates its nine for the VexRiscv product. Then quote the ratio as a bound —
  "1.491× as measured, of which up to ~2.2% is the harness's write wait state, so ≥1.46× is what the
  cores account for" — and carry the same qualifier into the CLAUDE.md bullet. A disclosed
  distortion of known size is a measurement; an undisclosed one is not.
- *Better if it is cheap:* subtract it. `coremark_tb.v` already watches Hazard3's write bus; counting
  the stalled cycles directly and printing them as their own fact would let `coremark_dmips.py`
  publish both the raw and the harness-corrected ratio, and would make the correction a graded number
  rather than an argument in prose.

Do not simply delete the wait state — the adapter comment records that zero wait states produced
silently wrong store data, and that catch is worth more than the 2.2%.

### 2. Blocking — `soc/compare/run_coremark_compare.sh:85-97`

**Problem.** The comment at `:85` says "Membership is a two-way match against PINNED.sha256, the same
reason `test/bench/run_coremark.sh` checks it: a file dropped in beside the vendored tree that shasum
was never told to look at is invisible to a one-way check." The code at `:97` is `shasum -c` and
nothing else. The comment travelled from `test/bench/run_coremark.sh:63-86`; the `comm` membership
check it describes did not.

The attack that comment names is live here: `coremark.h`'s `#include "core_portme.h"` is a quoted
include, so an unlisted `core_portme.h` dropped into `$VENDOR_DIR` shadows this port's real header
for every vendored unit — including its timing hooks — while `shasum -c` reports the tree unmodified
and the CRC, marker and `ramdiff` gates all still pass. (`TOTAL_DATA_SIZE` is not reachable this way;
`coremark.h:27` defines it before the include at `:38`.)

A comment asserting a check the code does not perform is worse than no comment, and it is the exact
failure mode this repo names as "a grader that cannot fail is not a grader."

**Fix.** Lift the `manifest_files` / `tree_files` / `comm -23` / `comm -13` block from
`test/bench/run_coremark.sh:63-86` verbatim ahead of the `shasum -c`, or factor the two into one
shared helper. Roughly ten lines.

### 3. Follow-up ticket, not a hold — `test/march_test.sh:145`

`Makefile rv32ima` exempts **every** `-march=rv32ima` occurrence in the Makefile forever: exception
entries carry a file and an ISA string but no occurrence count, while the required table is
exact-count. This is a pre-existing property of the exception mechanism (`Makefile rv32i` and
`Makefile rv32ic` have it too) that this PR inherits rather than introduces, and
`run_coremark_compare.sh:147-153`'s embedded-flags check fences it off from the cross-core number, so
the blast radius is littlecpu's own single-core figures. Worth a ticket to give exceptions counts;
not worth blocking this.

### 4. Follow-up ticket, not a hold — `soc/compare/geometry_test.sh:67`

`for top in bench_littlecpu bench_vexriscv` omits `bench_hazard3`, so that core's `ROM_WORDS`/
`RAM_WORDS` can drift from `COMPARE_ROM_WORDS`/`COMPARE_RAM_WORDS` ungraded. Latent rather than live
for CoreMark — `coremark_tb.v:82-88` overrides the geometry explicitly, so drift cannot reach this
PR's number; only `bench_tb.v` is exposed. Pre-existing from the Hazard3 PR and already recorded as a
known separable follow-up in ADR-0139:96-102.

### 5. Not a finding, but it will mislead a later reader

`coremark_fit.py`'s `cf_fixture` comment says it inflates Hazard3's block count "the same shape
`dhry_fit.py`'s own fixture does for vexriscv" — but `df_fixture` uses VexRiscv's *real* 18 blocks.
The inflation to 20 is legitimate and disclosed; the sentence claiming a precedent that does not
exist is what to fix. One line.

### Also required before merge, independent of the above

`make test` was not run. `probe-gates` (607/607), `march_test.sh`, `adr_numbering_test.sh` and
`compare-smoke` are all good and all necessary, but the diff touches `test/probe_gates.sh`,
`test/PROBES_EXPECTED` and `test/march_test.sh` — test infrastructure that `make test` is the grader
for — and CLAUDE.md requires the full suite before work is called done. It is also a required check,
so this is not optional in practice either.

### What is explicitly cleared

- **The VexRiscv decision stands as recorded.** RV32IC with no M extension is a fact about the pinned
  clone, not a choice, and both alternatives are correctly declined in ADR-0146 — dropping the shared
  image to `-march=rv32ic` would erase the hardware multipliers this pair exists to compare, and
  editing the pinned Verilog would break `soc/compare/`'s founding rule. Ratified; no change wanted.
- **The `x0`/`RESET_REGFILE` fix is applied the right way** — testbench poke by hierarchical path, no
  Hazard3 parameter moved, pin unchanged.
- **The Hazard3 clock difference (32.07/32.45 against ADR-0139's 32.60/33.63) is disclosed and
  qualified correctly** as a different five-seed sample. One thing to note for whoever re-takes it:
  littlecpu reproduced *exactly* on the same tree and Hazard3 did not, which is a slightly odd shape
  for two sweeps of an unchanged netlist. If the two sides were swept with different seed sets, say
  the seed sets; if the toolchain moved between them, CLAUDE.md's own rule about
  `soc/baseline_summary.py` refusing to subtract sweeps with disagreeing stamps applies.
- Diff scope is correct: no `rtl/` file changed, and nothing outside this ticket's surface.

### Note on CI

Nothing on this PR can merge right now for reasons that have nothing to do with it — the
`little-cpu-runners` self-hosted pool has zero registered runners, so ten of the eleven required
checks queued for 24 hours and were cancelled. Only `formal` (ubuntu-latest) ran, and it passed. The
work above is what to do while that is being restored, not a consequence of it.

thejefflarson:
The content hold stands — the undisclosed harness bias in a published cross-core ratio, and the comment asserting a `PINNED.sha256` two-way check the code does not perform. I have not rebased this branch, deliberately: resolving the ADR-index conflict would make it look ready when the reason it is held has not moved.

Three notes for whoever picks it up.

**1. #247 is now the worked example of the disclosure standard this is held to.** It measures Hazard3's Dhrystone cycle factor in the same three-way harness and handles the same AHB5 write-data wait state by **counting it directly (9.01% of measured cycles) and printing it beside the row rather than folding it in**. That is the shape the fix here wants — a disclosed, counted bias, not a corrected-away one — and it is a stronger answer than a prose caveat.

**2. Rebase for #250, which is now on `main`.** `test/march_test.sh`'s exception entries are `<path> <ISA> <count>` and are graded at an exact count the same `grep -cxF`/`-ne` way required sites already are. This branch adds two entries in the old two-field format:

- `Makefile rv32ima` → needs a count
- `soc/compare/run_coremark_compare.sh (empty)` → needs a count (`(empty)` is a value; only `(any)` is countless)

This is fail-closed rather than silent — an entry with no count fails the numeric comparison and the run goes red — so it will not sneak through. Worth knowing that #250 was written **because** `rv32ima` was coming: it is one keystroke from the declared ISA, and an exception that exempted every occurrence rather than a counted one would wave a second, unnoticed near-miss through. The exception this PR adds is exactly the case that motivated the counting.

**3. The ADR-index conflict is now cheap.** #248 landed a `merge=union` driver for `docs/adr/README.md`. It does not resolve *this* conflict — `.gitattributes` is added by the merge itself, so it is not in the working tree when git picks a strategy — but once this branch has merged `main` once, subsequent merges get the driver. Re-anchor the 0146 row by hand that one time. `test/adr_numbering_test.sh` grades the result both ways; 0146 is uncontested.

thejefflarson:
## Holds addressed

Rebased onto current `main` (picked up #244, #248, #249 cleanly, no conflicts — #250 has not merged, so the `march_test.sh` exception-count conflict the earlier note anticipated does not exist yet). `make test` and `make mutation-check` both run in full this time (they were the two things explicitly missing before).

### Hold 1 — undisclosed harness bias

Measured it rather than assuming the ~2.2% estimate: `coremark_tb.v` now counts cycles spent with `dut_haz.wr_pending_q` high the same way the Dhrystone sibling does, and it comes out to **14,176 of 714,984 measured cycles — 1.98%**, not 2.2% and nowhere near Dhrystone's 9.01% (different write mix). Took route 1 (publish the stall as its own counted fact) — `coremark_dmips.py` prints it as its own line and a probe pins the exact sentence. Both the raw and bias-corrected numbers are now quoted together everywhere they appear: ADR-0146, CLAUDE.md's Hazard3 bullet, and `docs/adr/README.md`'s row — **1.491×/1.437×–1.471× measured, 1.462×/1.408×–1.442× as a lower bound with the adapter's own cost taken out.**

### Hold 2 — asymmetric write counters

Fixed: `coremark_tb.v`'s Hazard3-side counter was gated on `mem_wstrb_mux == 4'b1111` (full word only); littlecpu's was `|mem_wstrb` (any strobe). Changed Hazard3's to `|dut_haz.mem_wstrb_mux`, symmetric with littlecpu's and with the pre-existing `bench_tb.v`/`dhry_tb.v` pattern, and moved the marker/verdict address compares out from under the width test per the finding. Re-measured: `writes=15701` on both sides now (previously 15701/13350) — cycles and the 1.491× figure are unaffected, since CoreMark's markers are always full-word stores by construction; only the previously-undercounted `writes=` figure moved.

### Hold 3 — manifest check is one-way

Lifted the `comm -23`/`comm -13` two-way membership check from `test/bench/run_coremark.sh` into `run_coremark_compare.sh` ahead of `shasum -c`, and swapped the `>/dev/null 2>&1` for a `mktemp` capture so the checker's own diagnostics reach the operator on both the pass and fail path (confirmed: a `shasum --strict` WARNING that doesn't fail the run now surfaces instead of being swallowed). New probe group mirrors `test/bench/run_coremark.sh`'s own four probes, including the concrete `core_portme.h`-shadowing case.

### Nit — `cf_fixture` comment

Fixed: it no longer claims `dhry_fit.py`'s own fixture also inflates a block count (it doesn't — it uses vexriscv's real 18).

## Also done

- `make mutation-check`: 11 mutations, each caught by exactly its paired detectors.
- `make test`: 74/74.
- `make probe-gates`: 613/613 (was 607 before this PR; +6 net from the new coremark_dmips.py wait-state probe and the new run_coremark_compare.sh manifest-check group, offset by the wording fix costing nothing).
- `make compare-smoke`: PASS, all three cores agree.
- `make compare-coremark`: re-run end to end twice (once to establish the fixed numbers, once after the `/simplify` pass) — identical both times.
- `/simplify` pass on the touched files: fixed a redundant post-`grade()` guard in `coremark_dmips.py` and restored a comment that got dropped when lifting the manifest check (explains why the pin-check output is `cat`'d even on the success path). Declined two larger refactors two independent review passes both surfaced (folding `coremark_dmips.py`/`coremark_fit.py` into shared modules with `dhry_dmips.py`/`dhry_fit.py`, and factoring the manifest check into a cross-directory shared script) — both are real duplication, but both are pre-existing patterns or explicitly-sanctioned-as-acceptable by the original review, and both would expand this PR's diff into files/architecture the hold didn't ask for. Worth a follow-up ticket if the duplication becomes a live maintenance cost.
- `/security-review` (soundcheck pr-review): no Critical/High findings — this diff has no network-facing surface, no user input, no credentials.

## Scope note

The altitude review pass also correctly points out that Dhrystone's own comparison (`dhry_tb.v`/`dhry_dmips.py`) has no equivalent wait-state disclosure on `main` yet — but that's #247's own work-in-progress (its branch already has it, at 9.01%), not something to duplicate here.

thejefflarson:
Update: after posting the above, #250 (`test/march_test.sh`'s exact-count exception format) merged to `main` while this was in flight. Rebased again — one real conflict, not a text one: `probe_gates.sh`'s own 'a second occurrence of a counted exception value is red' probe had planted a throwaway `Makefile rv32ima 1` exception and injected two fake occurrences, on the (then-true) assumption that no such exception existed yet. This PR's real `Makefile rv32ima 1` entry made that assumption false — the fixture's copy of the real Makefile now already carries one real occurrence, so the probe's two fakes summed to three, not two, and its pinned message went unmatched. Fixed by using the now-real exception instead of planting a duplicate, and injecting only the one occurrence the label already asks for.

Re-verified after: `make probe-gates` (614/614), `make test` (74/74), `make compare-smoke` (PASS). Branch is `MERGEABLE` against current `main`.

PR #244, opened 2026-08-30
Price a 12 KB text window: 24 block RAMs against the 26 free (declined on period)

thejefflarson:
## Summary

Prices ADR-0135's 8 KB text-window ceiling directly. Builds shape (a) from the ticket — two power-of-two windows ORed (3072 = 2048 + 1024) — measures it end to end, and declines it: the period does not survive.

## What was built and measured (then reverted)

- `rtl/imemory.v`: `ROM_WORDS` generalized to accept a power of two, or a power of two plus a second, smaller power of two. Both range tests stay reductions, ORed. The bank arrays go non-power-of-two (1536 words each); a bank-index clamp folds an out-of-range truncation back in before the array read, since that index is never a valid fetch address but the array still needs an in-bounds one.
- `rtl/decoder.v`: `ls_supported`/`ls_text_deep` needed the identical two-window shape — a plain load or store into the text region is classified by the decoder's own copy of the map, and that copy is mask-based with the same power-of-two assumption.
- `rtl/littlecpu.v`: its `LS_TEXT_WORDS` default had to move to 3072, because `test/memmap_test.sh` requires it to equal `rtl/littlesoc.v`'s actual value, and `formal/wrapper.v` instantiates `littlecpu` with no override — so this also moved every riscv-formal check's default configuration.
- Correctness checked out clean: all 86 generated riscv-formal checks passed at the new default, `components_traps` passed by k-induction, `test/zkt_isolation_test.py` passed unchanged, `SOC_EXPECT_EBR` measured at exactly 28 off the SoC's own synthesis log, and `test/asm/rvc.S` linked to 12256 bytes against the new 12288-byte ROM (32 bytes to spare).
- **The period did not.** Sixteen paired placements, same tree, base (2048, one window) against candidate (3072, two windows): median **−5.89%**, **16 of 16 seeds slower** (sign test p ≈ 3×10⁻⁵), **5 of 16 under the 12.00 MHz floor** (worst 11.86 MHz against the base's 12.06). `make fit`'s core-alone count was a null (4094 against 4109), so the cost is the fetch loop, not area — every candidate placement's critical path starts or ends inside `rtl/imemory.v`.
- Shape (b), a magnitude comparison, is declined without its own sixteen-seed sweep: `rtl/imemory.v`'s own header comment already measured that shape at a quarter of the whole period for the adjacent `next_is_last` test — same fetch loop, same class of non-power-of-two constant — and shape (a) already misses the floor.

## Decision

**Declined.** The shipping ROM stays 2048 words / 8 KB. `test/asm/rvc.S` still does not run on silicon, and `soc/run_suite_board.sh` still batches. All RTL, Makefile and test changes made while pricing shape (a) are reverted in the same commit that adds the ADR — see `docs/adr/0145-a-12-kb-text-window-two-windows-ored-not-a-magnitude-compare.md` for the full write-up, including what would change the answer (moving the new logic off the same-cycle fetch path, the way ADR-0129 did for the region fault).

## Testing

- `make window-test`, `make imem-share-test`, `make fit` (both trees), `make -C formal components_traps` and `make -C formal check` (86/86 pass at the candidate's default) — all run against the candidate before reverting.
- `make test` run clean on the fully-reverted tree (byte-identical to `main` per `git diff main -- <touched files>` = empty except the ADR and its README index row).
- Sixteen paired `make soc-timing` placements, base and candidate, same tree, same seeds — table and per-seed data in the ADR.
- `test/adr_numbering_test.sh` passes (no collision, no orphan; the gap at 0141–0144 is expected, those are other in-flight PRs).

## Scope notes / risks

- No RTL ships from this PR — it is a measurement and a decline, recorded as ADR-0145.
- `remeasure-fg` could not independently confirm F/G are unmoved for the (unshipped) candidate: `btorsim` is absent from this machine's boolector install, a pre-existing environment gap unrelated to this change (confirmed absent repo-wide). The 86/86 generated-check pass at the existing depths is suggestive but is not the same statement `remeasure-fg` makes; moot here since nothing ships.
- Fixed two environment issues on the shared build machine while pricing this (both outside the repo, affecting any engineer's `make soc-timing`): a `nextpnr-ice40` linked against a stale boost (fixed via `brew reinstall nextpnr-ice40`) and `icetime` looking for its chipdb under a `share/icebox` path this icestorm bottle doesn't ship (worked around with a symlink to the actual `share/icestorm/chipdb`). Neither is a repo change.

Closes JEF-908.


thejefflarson:

re-run green.

PR #246, opened 2026-08-30
Stop hand-refreshing the cross-core product: stamp it, detect staleness, re-take it on a schedule

thejefflarson:
Closes JEF-918.

## What this builds

The cross-core throughput product (Dhrystone against VexRiscv today, CoreMark
against Hazard3 once #243 lands) has gone stale on record twice — once inside
a day, once by being re-derived by hand a second time — because nothing
watched either factor. This ticket builds the watcher.

- **`make compare-product`** measures both factors of every cross-core pair
  in one run and writes `soc/compare/product.json`. Schema: **a benchmark
  names a list of cores**, not a fixed pair — each core carries its own clock
  distribution and cycle factor, and a product is derived per `(target core,
  other core)`. This is a mid-flight scope widening from the ticket's
  original two-core sketch: **JEF-919 is making the Dhrystone pair three-core
  in parallel**, and the artifact needed to accept that without a second
  schema change. Every measured pair stamps the commit, whether the tree was
  dirty, the seeds, the CFLAGS, **the bare `-march=` the shared image was
  built at**, and the resolved toolchain.
- **`soc/compare/product_check.py`** refuses to print a product once `rtl/`,
  `soc/compare/`, or a `--current FIELD=VALUE` has moved past the stamp, or
  the stamp was measured on a dirty tree. `make compare-dhrystone` calls it
  after its own measurement.
- **`soc/compare/product_diff.py --require-news`** reuses `product_check`'s
  own staleness question against the *older* snapshot, rather than inventing
  a percentage "moved enough" threshold — no churn band has ever been derived
  for the hx8k part this harness places (`soc/bands.py`'s own entry says so),
  so a magnitude gate would have been an unmeasured number standing in for
  one.
- **`.github/workflows/compare-product-schedule.yml`** re-measures the pair
  weekly, opens a PR only when the old stamp would call itself stale against
  the checked-out tree, and reports "no change" just as visibly when it does
  not (a silent no-op looks exactly like a job that never ran, which is what
  hid the runner pool going to zero for a day).
- CoreMark is feature-detected, not assumed: `make compare-coremark` and
  `soc/compare/coremark_dmips.py` are still on the held #243. Where absent,
  the pair is recorded `not_yet_measured`; where present, `run_product.sh`
  attempts the same measurement and falls back cleanly (with a diagnostic) if
  the attempt doesn't match the expected log shape, rather than blocking the
  whole run on a guess about an unmerged interface.
- CLAUDE.md gets one new sentence pointing at the mechanism. **The measured
  prose itself stays hand-written** — this PR does not regenerate any number
  in CLAUDE.md or an ADR.

## Tests

- `make test`: 74/74, full run, green.
- `make probe-gates`: 609/609. New groups for `product_check.py`,
  `product_write.py`, `product_diff.py` — every staleness/validation path has
  a demonstrated red direction, and `test/PROBES_EXPECTED` carries all the new
  labels.
- `test/adr_numbering_test.sh`, `test/band_source_test.py`: green (no new
  ADR; CLAUDE.md's one new sentence doesn't restate a band figure).
- `test/march_test.sh`: green, with new named exceptions for the artifact and
  the two scripts that reference `-march=` in help/comment text rather than
  as a real flag.
- Ran the full pipeline for real: `soc/compare/product.json` ships an actual
  four-seed sample (`seeds: "default 1 2 3"`, transparently a look and not a
  twelve-seed verdict), reproducing ADR-0098's own recorded cycle counts
  exactly (littlecpu 304428, VexRiscv 408758). It's stamped `dirty: yes`
  because it was measured mid-development of these very files under
  `soc/compare/` — which the freshness check correctly calls stale against
  itself in this PR's own diff. That's the check working, not a bug: once
  merged, the first `make compare-product` (by hand or the scheduled job)
  stamps a clean one.

## Decisions made without asking

- **Schema is `benchmark → cores[]`, products per pair, ISA stamped
  separately from full CFLAGS** — per the mid-task scope note above, to stay
  compatible with JEF-919 landing a third core.
- **No numeric "moved enough" threshold.** `--require-news` asks whether the
  older stamp would call itself stale against the current tree (same
  question `product_check.py` already answers), instead of a churn-band
  percentage nobody has measured for hx8k.
- **The scheduled workflow opens a PR** (not an issue, unlike
  `riscv-formal-pin-bump.yml`'s precedent) because that's what the ticket
  asked for and `soc/compare/product.json` is data, not RTL — but
  GITHUB_TOKEN-authored pushes don't trigger `pull_request`/`push` workflows,
  so the PR it opens will show no CI until nudged (documented in the
  workflow's own header).
- **CoreMark's measurement glue in `run_product.sh` is a best-effort guess**
  at #243's interface (`COREMARK core=... cycles=... iterations=...` lines,
  `COMPARE_COREMARK_CFLAGS`), wrapped to fall back to `not_yet_measured`
  rather than fail the whole run if the guess is wrong. Landing #243 may need
  a small follow-up fix to `measure_coremark()` in `run_product.sh` if the
  real interface differs — the schema, freshness check, and workflow do not
  need a second change either way.


thejefflarson:
Assessed, not merged — no content hold, but two things to settle first and it has had no architect review until now.

**The mechanism is sound and matches how this repo already stamps measurements.** `soc/baseline_sweep.sh` stamps a sweep with its base commit and resolved tool versions and `soc/baseline_summary.py` refuses to subtract two sweeps whose stamps disagree; `soc/compare/product.json` is the same idea applied to the cross-core product, and the problem it closes is real and recorded — that paragraph has gone stale twice, once inside a day.

Specifics I checked and agree with:

- **`--require-news` asks `product_check.py`'s own staleness question instead of inventing a percentage threshold.** Correct, and the header says why: no churn band has ever been derived for the hx8k part this harness places, so a moved-by-how-much gate would be a number nobody swept standing in for one. That is the right refusal.
- **The graded call is not in a pipeline.** The `tee` run is separate from the status-bearing call, citing CLAUDE.md's rule about `errexit` without `pipefail` making a step's status `tee`'s. That rule held the formal job accidentally green for its whole life, so getting it right here matters.
- **`runs-on: little-cpu-runners` bare, not `ci.yml`'s fork-PR ternary** — justified because `schedule` and `workflow_dispatch` carry no fork-PR context, so no untrusted code reaches the pool. That reasoning holds: `schedule` only runs on the default branch and `workflow_dispatch` requires write access. `permissions:` is scoped to `contents: write` + `pull-requests: write` rather than blanket, `actions/checkout` is SHA-pinned, and **no auto-merge** is explicit. Good.
- The `GITHUB_TOKEN` anti-recursion caveat is documented honestly rather than discovered later, and the no-move path is made as visible as the move path.

**1. Latent: the CoreMark arm references a variable that does not exist yet.** The workflow passes

```
--current "coremark:cflags=$(make -s print-COMPARE_COREMARK_CFLAGS)"
```

unconditionally, but `COMPARE_COREMARK_CFLAGS` is introduced by **#243**, which is currently held on content. `print-%` does exist on `main`, so the three Dhrystone calls are fine; this one resolves to an empty string. `run_product.sh:1123` guards its own CoreMark path with `&&`, so the guarding instinct was there — the workflow line did not get it. Please confirm `product_diff.py` treats an empty `--current` for an unmeasured pair as a no-op rather than as news, or guard the line the same way the script does. This will not show up on any PR: it is a weekly cron, so a wrong answer here surfaces as a Monday morning PR nobody asked for, or as silence.

**2. Rebase for #250**, now on `main`: `test/march_test.sh`'s exception format is `<path> <ISA> <count>`. `soc/compare/product.json (any)` is correct as-is — `(any)` carries no count by design — but `soc/compare/run_product.sh (empty)` and `soc/compare/product_write.py (empty)` each need a trailing count, because `(empty)` is a value and not a wildcard.

Neither is a content objection. Settle (1) and rebase, and I will merge on the required set.

PR #247, opened 2026-08-30
Dhrystone runs on all three comparison cores, one image, one simulation

thejefflarson:
Closes JEF-919.

Dhrystone now runs on all three `soc/compare/` cores in one image and one
simulation, not just littlecpu and VexRiscv, closing the gap ADR-0139 left
explicit for Hazard3's iCE40 build.

## The ISA

Verified against the actual pinned sources, not trusted from a prior
summary:

- **littlecpu** — RV32IMAC + Zicsr + Zifencei + Zkt.
- **VexRiscv** (pinned riscv-formal clone) — RV32IC, no M: no `mul`/`div`
  identifier anywhere in the generated Verilog.
- **Hazard3's iCE40 build** — RV32IMA, no C: `fpga_icebreaker.v` states
  `EXTENSION_C(0)`, `EXTENSION_M(1)`, `EXTENSION_A(1)`.

The intersection is plain **RV32I**, so `COMPARE_DHRY_CFLAGS` moves from
`rv32ic` to `rv32i`. Multiply and divide are libgcc calls for every core in
the three-way row, even the two whose netlists carry a hardware multiplier.

## The three-way row, measured on this tree, 400 runs

| | cycles | DMIPS/MHz |
|---|---|---|
| littlecpu | 300,427 | **0.758** |
| VexRiscv | 385,647 | **0.590** |
| Hazard3 (iCE40) | 319,632 | **0.712** |

Every pairwise ratio the ISA permits comes out of this one image (RV32I is
the only ISA all three share): VexRiscv 1.284× littlecpu's cycles, Hazard3
1.064×, Hazard3 0.829× VexRiscv's.

**Hazard3's row discloses a bias the other two do not pay.** Its AHB5
adapter's mandatory write-data wait state (one cycle of `hready` low per
write's address phase) is counted directly from `wr_pending_q`, not
estimated: **9.01% of its measured cycles**. `soc/compare/dhry_dmips.py`
prints it beside the row. Also confirmed: all three cores' writes are
counted the same way (`|mem_wstrb`/`|mem_wstrb_mux`, never `== 4'b1111`),
matching the two checks a security review already made for the two-core
VexRiscv pair.

## The ISA's cost, isolated from the geometry

`soc/compare/dhry_solo_tb.v` runs littlecpu alone at the same enlarged
geometry, built at its native `-march=rv32imac_zicsr_zifencei_zkt` instead
of the shared RV32I:

| | cycles | DMIPS/MHz |
|---|---|---|
| littlecpu, RV32I (shared) | 300,427 | 0.758 |
| littlecpu, native ISA | 301,227 | 0.756 |

RV32I costs this core nothing measurable on Dhrystone — if anything it's
0.27% *faster*, a reproducible compiler-codegen difference, not a
hardware-capability cost. Dhrystone barely multiplies and this core's fetch
machinery already decodes a compressed successor, so C's density win and
M's hardware-multiply win both land on a benchmark that spends almost
nothing on either.

## The block-RAM question a third core raises

Checked before writing the rest of this, not assumed: Hazard3's iCE40 build
is 4 `SB_RAM40_4K` standalone — the same as littlecpu's — so it fits the
harness's 32-block budget the way littlecpu does; VexRiscv's branch
predictor still does not (46 of 32).

## What this does not close

**No absolute DMIPS product is quoted.** `soc/compare/sweep.sh`'s existing
five-seed clock table (ADR-0139) predates ADR-0129's region-wait fault — a
real change to `rtl/memory.v`/`rtl/decoder.v` on the critical path
`soc/compare/bench.S` also exercises — so multiplying fresh cycle counts by
that stale clock would repeat the exact "two trees, one number" mistake
this ADR already named and reversed once for the VexRiscv pair.

A twelve-seed re-sweep of all three cores, paired by seed, is the
documented follow-up (exact command in the ADR). I started one during this
session; host contention (`uptime` load average 20–34 on a 10-core shared
machine, several other concurrent agent worktrees) made completing all
36 placements impractical before its own numbers went stale in turn. Three
of eleven extra littlecpu seeds landed before I stopped waiting on it —
30.90 / 32.69 / 32.37 MHz, consistent with (not a replacement for)
ADR-0139's existing five-seed figures — which I did not fold into the
shipped ADR text, since a partial sweep quoted as if complete is the same
mistake in miniature.

**DECISION NEEDED**: whoever picks up the twelve-seed re-sweep should also
decide whether to fold the resulting product into this ADR or into a new
one; I judged amending ADR-0098 in place (its natural home, matching how a
prior PR amended it for the VexRiscv pair's own twelve-seed re-take)
sufficient and did not take ADR-0150.

## Coordination

- **JEF-918** (product.json stamping, open PR #246): this PR does not touch
  that artifact. Whichever lands second should confirm the schema allows N
  cores per benchmark row, not exactly two.
- **JEF-912** (Hazard3 wait-state / write-counter security finding): not
  duplicated here — this PR's own instance of both checks (on the Dhrystone
  three-way pair) is clean, and disclosed rather than corrected, matching
  how a prior PR handled the identical question for the two-core pair.

## Simplify pass

Four parallel review agents (reuse / simplification / efficiency /
altitude) found real duplication in the first draft: `dhry_solo_tb.v`'s
marker-tracking logic copied `dhry_tb.v`'s inline, and `run_dhrystone.sh`
detected mul/div by pattern-matching `-march=` text (fragile, and it
planted a second `-march=`-shaped literal for `test/march_test.sh`'s own
scanner to trip over). Fixed: `soc/compare/dhry_monitor.v` factors the
per-core marker/verdict tracking into one module instantiated per DUT
(verified bit-identical against the pre-refactor cycle counts, twice); the
Makefile now passes `hardware`/`libgcc` through explicitly instead of
`run_dhrystone.sh` re-deriving it. `dhry_dmips.py`'s pairwise-ratio
printing was also simplified from an `itertools.combinations` all-pairs
model down to reference-core-only, matching the RAM-agreement guarantee
the script actually makes.

## Tests

- `soc/compare/dhry_dmips.py` generalized from a fixed two-core interface to
  an N-core `--cores` list with a reference-core RAM-agreement rule; new
  `test/probe_gates.sh` probes cover the three-way row, the *second*
  non-reference core's RAM divergence (not only the first), the wait-state
  disclosure, `--cores` naming no core, and the solo (no-reference,
  no-ramdiff) case.
- `test/march_test.sh`'s exception table updated: `COMPARE_DHRY_CFLAGS`'s
  `rv32i` folds into the existing `bench.S` entry; the stale `rv32ic`
  entry is removed.

`make probe-gates`: 594 comparisons, 0 red.
`make compare-smoke`: PASS, all three cores agree (unaffected by the
`dhry_monitor.v` refactor — different testbench).
`make compare-dhrystone`: run twice end-to-end (once before, once after
the monitor refactor); cycle counts, ramdiffs and the wait-cycle count are
byte-identical both times.

Manual self-review against soundcheck's OWASP/LLM patterns: no findings —
this is a local build-harness/RTL-simulation change with no network
surface, no user input, no credentials.

## Scope notes

- `soc/compare/geometry_test.sh`'s six-file cross-check is still two-core
  only (littlecpu/VexRiscv), matching the precedent ADR-0139 already set
  when Hazard3 joined the clock-only comparison — extending it to a third
  file is called out there as a small, separable follow-up, and I left it
  that way rather than touching a file another concurrent PR might also be
  mid-edit on.
- No `rtl/` file changed.


thejefflarson:
Update: the twelve-seed clock sweep flagged as a DECISION NEEDED in the PR description finished shortly after I opened this PR (host contention eased). Pushed a follow-up commit (`213bbd0`) that folds it in on the same tree the cycle table was measured on, closing that gap rather than leaving it deferred:

| | worst | median | best |
|---|---|---|---|
| littlecpu | 30.90 MHz | 32.00 MHz | 33.06 MHz |
| VexRiscv | 48.24 MHz | 49.55 MHz | 50.84 MHz |
| Hazard3 (iCE40) | 32.07 MHz | 33.20 MHz | 33.78 MHz |

littlecpu's worst-of-twelve reproduces ADR-0139's own five-seed 30.90 MHz exactly, so the staleness risk the first commit flagged didn't materialize on this path — worth having checked rather than assumed either way.

Every product the ISA permits, worst-of-twelve × the RV32I cycle table:

- VexRiscv leads littlecpu **1.22×** (28.46 against 23.42 DMIPS, 1.21× on medians)
- littlecpu leads Hazard3 **1.03×** (23.42 against 22.83) — within 3%, littlecpu ahead both ways
- VexRiscv leads Hazard3 **1.25×** (28.46 against 22.83, 1.24× on medians)

All three rankings hold reading medians instead of worst placements. `docs/adr/0098`, `CLAUDE.md` and `docs/adr/README.md` updated to match; no code changed in this commit.

The DECISION NEEDED in the original PR description (fold the eventual sweep into ADR-0098 vs. a new ADR-0150) is resolved the same way I proposed: folded into ADR-0098 in place.

thejefflarson:
Assessed, not merged — the content is good and the blocker is ordering.

**This PR depends on #245 and must land after it.** Its `CLAUDE.md` text cites *"a different ISA and image from the two-core RV32IC pair's **1.05–1.16×** above"* — the 1.16× is #245's re-take, which is not on `main` yet. Merged first, that sentence points at a figure the file does not contain.

The dependency is not just prose. Both branches append to the same `ADR-0098` header line (after *"Amended 2026-08-12"*) and both add a `## Amendment, 2026-08-30` section. After #245 lands, this branch's header edit needs to become the **third** amendment in the chain rather than the second, and the two same-dated sections need an explicit order. That is reconciliation of two published measurement records, not a mechanical re-anchor, so I have deliberately not rebased it for you — getting it wrong would silently misattribute which sweep produced which number, which is the exact failure `soc/baseline_summary.py` refuses to make elsewhere in this repo.

Also rebase for **#250**, now on `main`: `test/march_test.sh`'s exception format is `<path> <ISA> <count>`. This PR deletes `Makefile rv32ic`, which is now `Makefile rv32ic 1`, and rewrites the comment above `Makefile rv32i`, now `Makefile rv32i 1`.

What I checked and liked, so it does not get lost in a rebase:

- **The three-way row is measured at RV32I, the only ISA all three cores share**, with each pinned build's subset verified against its actual sources rather than a README — including grepping `VexRiscv.v` for a `mul`/`div` identifier. That is this repo's "an inherited conclusion is not a measurement" rule applied correctly.
- **Hazard3's AHB5 write-data wait state is disclosed as 9.01% of its measured cycles, counted directly and printed beside the row rather than folded in.** That is the right handling, and it is the same class of defect #243 is currently held for — worth noting that this PR demonstrates the disclosure standard #243 has not met.
- Twelve placements a side, and littlecpu's worst reproduces ADR-0139's independent five-seed 30.90 MHz exactly.
- No new mechanism for Hazard3's cycle factor — the existing marker protocol, built for VexRiscv's identical CSR-free gap, needed only wiring. Cheapest thing that works.

No content hold from me. Land #245, rebase, and this is a merge candidate on the required set.

PR #278, opened 2026-09-06
make doctor: verify the toolchain runs, not just that it is on PATH

thejefflarson:
## Problem

Two machine-wide toolchain breakages blocked work this week and were repaired outside the repository, so a fresh machine (or one where the fix silently regresses) hits them again with no warning until mid-sweep:

1. `nextpnr-ice40` broken by a stale boost linkage — the binary answers `--version` with a dynamic-linker error.
2. `icetime`'s chip database not resolving relative to its own binary path (a flattened Homebrew symlink, or a stale OSS CAD Suite cache lacking `chipdb-up5k.txt`) — the binary runs fine, but every device lookup fails, which breaks the timing half of `make soc-timing`.

## What changed

- **`soc/print_toolchain.sh`**: `icetime`'s handling now runs a one-line stub `.asc` through the resolved binary (`icetime -d up5k <stub>`) and requires it to reach `Creating timing netlist` — a positive success marker, not a denylist of one known failure string — before stamping a version. This function is already called by `fit-toolchain` and `soc-timing-toolchain`, both prerequisites of `make fit` / `make soc-timing`, so both now catch a broken chip database **before spending a placement**, with no new call site needed. Case 1 (nextpnr's dynamic-linker error) was already caught by the existing `--version` check — verified this is live and unrelated to icetime's failure mode.
- **`Makefile`**: new `make doctor` target — resolves the RISC-V cross compiler (same candidate loop `soc-rom` uses) and calls `soc/print_toolchain.sh` with it plus `$(SOC_TIMING_TOOLS)` (yosys, nextpnr-ice40, icetime), printing each tool's resolved path and version in one command to run before a placement or a sweep.
- **`test/probe_gates.sh` / `test/PROBES_EXPECTED`**: one stub `icetime` (toggled by `STUB_ICETIME_BADCHIPDB`, matching the file's established env-var-toggle stub pattern) and two probes demonstrating both directions of the new refusal.
- **`CLAUDE.md`**: one line documenting `make doctor`.

## Design decisions

- **`make doctor` is a standalone target, not wired into `make setup`'s critical path.** `make setup` is documented as installing only the RISC-V compiler and svlint; the OSS CAD Suite (yosys/nextpnr/icetime) is a separate toolchain many contributors never install, and CI fetches it fresh at "the latest release" rather than pinning it. Hard-failing `make setup` on a missing FPGA toolchain would break it for every non-FPGA contributor. `make doctor` is the "run before FPGA/timing work" command the ticket also names as an acceptable shape.
- **Not wired into `make test`.** The ticket requires a missing tool not be fatal to targets that don't need it; `make test` never touches nextpnr/icetime.
- **The icetime probe is a positive-success check, not a denylist**, so an unrecognized future breakage (a reworded message, a different failure mode) also gets caught rather than silently reported healthy.

Verified this reproduces exactly the reported failure mode: on this machine, `icetime`'s chip database resolution depends on a real `share/icebox/chipdb-<device>.txt`, and the probe's diagnosis matches what a genuinely broken install prints.

## Testing

- `make test`: 75/75, failure list matches `test/EXPECTED_FAIL` exactly.
- `make probe-gates`: 698/698, both new probes (`control: icetime whose chip database resolves is stamped like any tool`, `icetime that cannot resolve its chip database is refused before a version is stamped`) pass.
- `make doctor`, `make fit-toolchain`, `make soc-timing-toolchain`: manually verified against the real toolchain on this machine and against synthetic good/bad `icetime` stubs.
- Ran `/simplify` (four parallel review angles) on the diff and applied: reused `SOC_TIMING_TOOLS` instead of a second hardcoded tool list in `doctor`, merged two near-duplicate test stubs into one env-var-toggled stub matching the file's existing convention, and switched the chipdb check from a denylist of one known error string to a positive success marker. Declined as out of scope: refactoring the RISC-V-compiler candidate loop, which is already duplicated across ~15 other call sites in this repo predating this change.
- Ran a security self-review (soundcheck pr-review shape): no findings — this is developer-facing build tooling with no untrusted input, network calls, or secrets.

Closes JEF-929


thejefflarson:
Ratifying the decision this PR flagged: **`make doctor` stays off `make setup`.**

`CLAUDE.md` documents `make setup` as "macOS: brew install riscv64-elf-gcc and svlint; Linux: prints the apt line", and it names the OSS CAD Suite separately as "the one tool that floats" — CI takes it at the latest release and no `setup` step installs it. Wiring a hard-failing toolchain probe into `setup` would make the command fail on exactly the machine it exists to serve: a fresh one, which by definition has no yosys, no nextpnr and no icetime yet. The split stands.

`doctor` reaching further than `soc-timing`'s own prerequisite list — it adds the RISC-V compiler those two do not need — is the right shape for a command a human runs on purpose.

Filed as a follow-up rather than held here: `icetime_probe()` runs the PATH-resolved binary with no timeout, so a hung `icetime` wedges `make doctor`, `make fit`, `make soc-timing` and any CI job built on them until the runner's own timeout. Fixing it needs care — `timeout(1)` is GNU coreutils and not on stock macOS, and this repo's primary dev machine is darwin — so it wants the background-PID-and-kill spelling or a `gtimeout` fallback, with its own forced-red probe.

PR #285, opened 2026-09-06
Eliminate the AHB write wait-state bias from the cross-core harness

thejefflarson:
## Update: route 3 reverted after timing review

The architect's review caught a real problem with this PR's original approach: the write-forwarding
fix below cost Hazard3 its own 12 MHz step on up5k. See commit 424bc13 for the full investigation
and revert. **This PR is now a no-op against main plus a documented investigation** — the RTL is
back at its pre-#285 text, and `docs/adr/0146` carries the finding for the future.

### What was measured

`soc/compare/sweep.sh COMPARE_CORES=hazard3 COMPARE_PART=up5k`, hazard3 alone, worst of three seeds
(default, 1, 2):

| spelling | worst MHz |
|---|---|
| baseline (`hready = !wr_pending_q`) | **13.15** |
| full fix: ROM off `haddr` + read-after-write forwarding | 10.92 |
| ROM off `haddr` alone, no forwarding | 10.53 |
| ROM off `haddr` alone, `hready` gated on `is_rom_next` only | 10.63 |

All three route-3 spellings cluster at 10.5-11.5 MHz — a clean miss of the 12 MHz requirement, and a
drop to the 6 MHz step below it (`SB_HFOSC` gives 48/24/12/6). This is **not** the forwarding
comparator's fault: the comparator-free spelling isn't meaningfully faster than the full one, and a
minimal one-term spelling costs the same again. What every spelling shares is that `hready` — the
baseline's `!wr_pending_q`, a bare flip-flop output with zero same-cycle dependency on anything the
core presents — now has to read `haddr` combinationally within the cycle to decide whether the
current transaction may proceed, since "selective" is exactly a claim about what the current
transaction is. That feeds back into `hazard3_cpu_1port`'s own address-phase logic on a path with
too little slack above 12 MHz (~8-13% at baseline) to absorb any same-cycle `haddr`-dependence,
however cheap the comparison itself.

The cycle-count win was real (Dhrystone's disclosed wait share fell 9.01% -> 0.95%, CoreMark's 1.98%
-> 0.30%, both RAM comparisons still bit-identical, `make compare-smoke` still green) but the
published DMIPS/CoreMark-at-clock figures depend on the very clock step it costs. Route 2's
disclosure stands.

### What shipped in the revert commit (424bc13)

- `soc/compare/bench_hazard3.v`, `soc/compare/dhry_tb.v`, `soc/compare/coremark_tb.v` reverted
  byte-for-byte to their pre-#285 text (verified against `c080e67`).
- `docs/adr/0146-*.md` gets a new amendment section (route 3 attempted and declined) replacing the
  earlier "removed, not disclosed" one — the original dated measurement stays untouched.
- `docs/adr/README.md`'s ADR-0146 row and `CLAUDE.md`'s cross-core bullet updated to match.
- `make test` 75/75, `make probe-gates` 714 (unchanged), `make compare-smoke` green,
  `make compare-dhrystone` reproduces the exact pre-#285 disclosed numbers
  (319632 cycles, 28805 wait_cycles, 9.01%).

---

## Original PR body (superseded by the revert above, kept for history)

Route 3 for JEF-912: instead of disclosing Hazard3's AHB5 write wait-state bias (route 2, PR #243), this removes almost all of it.

`soc/compare/bench_hazard3.v`'s one-entry write-drain buffer used to route ROM reads through the SAME address mux the RAM write-drain used, so **every transaction following a store — including an instruction fetch, the common case — paid the drain's cost**, even though the ROM block RAM and the RAM's SPRAM are two separate physical ports with nothing to contend over. This was found, mid-review, to cost Hazard3 its own 12 MHz step on up5k -- see the update at the top of this description.

Closes JEF-912.


thejefflarson:
Requesting changes: this fixes a real cycle bias and introduces a larger clock bias pointing the same way. Both are artifacts of this adapter, which is our wrapper around their core, so both get published as facts about Hazard3.

**What #289 measured on a tree rebased onto this branch**, twelve up5k seeds, one toolchain:

| | before this PR | after this PR |
|---|---|---|
| hazard3 worst / median / best | 12.56 / 12.97 / 13.18 MHz | **10.80 / 10.94 / 11.54** |
| step reached | 12 | **6** |
| hazard3 DMIPS | 8.54 | **4.47** |
| littlecpu's lead | 1.09x | **2.10x** |

Hazard3's own DMIPS/MHz *improves* here, 0.712 to 0.745 — the cycle fix works. It falls off the 12 MHz step because the forwarding comparator lands on `hready`'s combinational path, and on this part the next step down is 6, so a small timing regression halves the published figure.

Net: we traded a 9.01% cycle bias in our favour for a ~2x clock bias in our favour. That is worse than what it replaced, and it is the fourth time in this workstream that a harness artifact has flattered this core. A 2.10x win over Hazard3 is not a fact about Hazard3.

**What is right in here and must survive:** decoupling ROM reads from the write-drain mux. `rom_index` moving from `mem_addr_mux[ROM_BITS+1:2]` to `haddr[ROM_BITS+1:2]` is the finding — ROM and RAM are separate physical ports and a fetch after a store never needed to wait. That is almost certainly where most of the 9.01% went, because a fetch is the common case after a store.

**What to reconsider:** the same-word read forwarding. It puts a 30-bit comparison (`haddr[31:2] == wr_addr_q[31:2]`) plus a strobe test inside `hready`, and `hready` gates the core's own address register. One hypothesis worth measuring first, because it is nearly free to try: drop the forward entirely and let a RAM read during a drain simply stall —

```verilog
assign hready = !(wr_pending_q && want_read && !is_rom_next);
```

no comparator, no `fwd_valid_q`/`fwd_data_q`. That is the original behaviour plus the ROM decoupling. Measure what it costs in cycles against the full version and whether it closes 12 MHz at twelve to sixteen paired seeds. If the ROM fix carries most of the benefit, the comparator is buying a little CPI at the price of the opponent's clock step, which is a bad trade at this quantisation.

If neither spelling closes 12 MHz, that is a real and publishable answer: route 3 is partly achievable, the ROM half lands, the forwarding half does not, and JEF-912's disclosure stays for the residual. Unfinished beats wrong.

**Acceptance for this PR:** all three cores reach the 12 MHz step on up5k at twelve to sixteen paired seeds, worst/median/spread quoted; `make compare-smoke` green across all three; the cycle numbers re-measured on whatever spelling ships. #289 is based on this branch and carries the ADR and CLAUDE.md corrections, so the two need to land together.

thejefflarson:
Closing per architect direction — this was route 3's docs-only revert; ADR-0146's route-3 decline content is being carried into PR #289 instead, and route 4 starts fresh from main in a new PR.

PR #286, opened 2026-09-06
Split the hazard column into its three causes, as a graded identity

thejefflarson:
Closes JEF-933

## What changed

`rtl/decoder.v`'s `hazard_rs1`/`hazard_rs2` collapse three different situations
into one bucket: the producer still in `out` with no result anywhere (A), the
producer in the executor with its result not yet unpacked -- a load, an AMO,
`lr.w`, `sc.w` (B), and a ready result forwarding has no path to because the
consumer sits outside `rs1_fwd_eligible`/`rs2_fwd_eligible` (C). Only C is a
candidate for anything, and it had never been sized against this tree --
previously only scratch instrumentation with no grader behind it.

`test/cxxrtl.cc` now classifies whichever of `hazard_rs1`/`hazard_rs2` gets
charged, using only `out_match_rsN` and `rsN_fwd_eligible`: `ex_match` is
implied whenever `out_match` is false (the only other way `hazard_rsN` can be
true), and `eligible` implies `!executor_out.rd_ready` whenever `ex_match`
holds, so `rd_ready` itself is never read. **No RTL changed and no new stall
reason was declared** -- all the signals this reads were already cxxrtl debug
items, and CLAUDE.md's six-declaration-sites rule for stall reasons does not
fire here.

`test/stall_report.py` checks `hzA + hzB + hzC == hazard` as an identity, the
same shape as the eight-column check it already runs against `cycles`, and
reports `hzCcsr`, C's slice behind a CSR register-form read.

**`hzCcsr` measures nonzero on the `.S` suite and zero on Dhrystone** -- the
ticket's own RTL argument for "provably zero" does not hold in general
(`serialize` narrows the window but does not close it): `test/asm/csr.S`'s own
vectors feed a CSR from a value the immediately preceding instruction just
produced (`csrrw a1, mscratch, a0`), which is exactly the C shape. Per the
ticket's "reported separately or asserted zero" allowance, and its own
"report the new numbers, don't force the old ones" instruction, this is
reported rather than hard-asserted.

`test/probe_gates.sh` mis-charges `hzB` by one cycle to demonstrate the
identity check goes red on a dropped classification arm; the label is in
`test/PROBES_EXPECTED`, sorted.

## Numbers (this tree, this toolchain)

- `make dhrystone`: HAZARD 317,207 cycles splits **167,844 / 6,007 / 143,356**
  (A/B/C) -- reproduces the ticket's scratch-instrumentation numbers exactly.
  `hzCcsr` = 0.
- `make cycles` (`.S` suite): HAZARD 12,121 cycles splits **5,739 / 317 /
  6,065**. `hzCcsr` = 121 (all in `csr.S`/`csrset.S`/`hpm.S`/`zicsr.S`, the
  CSR-forwarding-adjacent vectors).

## Testing

- `make sim && make cycles` -- prints the A/B/C split under HAZARD; identity
  holds.
- `make dhrystone` -- same, reproduces the ticket's quoted numbers exactly.
- `make probe-gates` -- 742+ graded comparisons green, including the new
  mis-charge probe (`hazard's three causes losing a cycle between them is red`).
- `make test` -- 75/75 `.S`/`.c` suite passed, baselines match, full green.
- `make lint` -- svlint clean both passes (no RTL touched).
- `/soundcheck:pr-review` -- no Critical/High findings (pure test
  instrumentation reading simulator debug items; no untrusted input, no
  network, no secrets).
- `/simplify` -- reviewed by hand given the diff's size (4 files, ~120 lines);
  no changes needed.

## Scope notes

- Instrumentation only, as the ticket requires: no `make fit`, no seed sweep,
  no netlist digest, no F/G re-measurement.
- `hzCcsr`'s nonzero suite measurement is a real finding against the ticket's
  own RTL reasoning; flagging it here rather than silently forcing zero or
  hard-failing the build on a true (if narrow) hazard.


thejefflarson:
Architect review: merging.

DECISION on the flagged item (per the ticket's 'reported separately or asserted zero' allowance): no ADR, no CLAUDE.md amendment. The 'provably zero' claim lived only in the ticket, never in CLAUDE.md or an existing ADR, so there's nothing on record to correct. This is diagnostic instrumentation (a cxxrtl debug counter + a stall_report.py identity check), not a design commitment/ceiling/ratchet, and it gates nothing beyond hzA+hzB+hzC==hazard. The repo's own precedent for a measured-but-unproven fact (exec_tb.v's ORACLE BROKEN, sanitize_monitor.py's site-count check) is to record it in code next to the mechanism, which test/stall_report.py's new docstring already does: the fact (121 cycles on .S, 0 on Dhrystone), the mechanism (csr.S's vectors feed a CSR from the immediately preceding instruction's result), and the caveat (serialize narrows, does not close, the window). Revisit with an ADR the day something starts assuming hzCcsr==0 as a gate.

Verified independently rather than trusted: zero RTL changes (diff touches only test/PROBES_EXPECTED, test/cxxrtl.cc, test/probe_gates.sh, test/stall_report.py); all six signals read (hazard_rs1/2, out_match_rs1/2, rs1_fwd_eligible/2, instr_csr_access) already exist unchanged in rtl/decoder.v on main. The hazard bucket stays one OR'd stall reason -- none of CLAUDE.md's six new-stall-reason declaration sites are touched, because this isn't declaring one. No PROBES_EXPECTED collision with the other in-flight PRs, no ADR/invariant numbers leaked into the new code comments.

PR #295, opened 2026-09-07
Thread ROM_WORDS through littlesoc so a wider ROM can be measured and CoreMark can run on the iCESugar-Pro

thejefflarson:
## Summary

Every published CoreMark figure describes hardware this project does not ship: `make coremark` links against `test/testbench.v`'s simulated 16 KB ROM, double the up5k's shipping 8. The iCESugar-Pro (ECP5 LFE5U-25F) has 20 spare `DP16KD` blocks over the 36 `littlesoc` already uses, and no up5k-style 8 KB ceiling, so it looked like the one route to a silicon CoreMark figure.

**The blocker, fixed first:** `rtl/littlesoc.v` hard-coded `ROM_WORDS=2048` twice — once into `littlecpu`'s `LS_TEXT_WORDS`, once into `imemory`'s `ROM_WORDS` — and no Makefile recipe threaded `SOC_ROM_WORDS` to either. `make ecp5-timing SOC_ROM_WORDS=4096` generated a 16 KB hex and then silently synthesised the unchanged 8 KB design. One `ROM_WORDS` parameter now drives both instantiations, and `SOC_SYNTH`, `ecp5.json` and `icesugar.json` each `chparam` it before `hierarchy`/`synth_*` — the same order `COMPARE_READ`'s callers already use for `COMPARE_ROM_WORDS` — consolidated into one `SOC_ROM_CHPARAM` Make variable so the three call sites can't drift from each other.

**What it measures, at 16 KB:**
- `DP16KD` 36 → **40**/56 (was silently unmeasured before the fix)
- Fmax (200 MHz constraint, nextpnr's own estimate): 35.11 → 34.78 MHz (0.94%, a null)
- `icesugar.bit` at the wider ROM still **PASSes** nextpnr's own 25 MHz check with margin
- CoreMark's built image is 14,236 of 16,384 bytes — **it fits the iCESugar-Pro**, where it does not the up5k (ADR-0135: 32 block RAMs against 26 free there)

**New targets:** `make coremark-rom` builds CoreMark against `test/bench/coremark.lds` with `COREMARK_CFLAGS` unchanged (the route's whole advantage), mirroring `dhrystone-rom`'s shape. `make icesugar-coremark` chains it through `icesugar-prog` and a UART read, mirroring `icesugar-dhrystone`.

**Verified against cxxrtl on the exact board binary**, the way ADR-0163 verified Dhrystone: same `COREMARK_CFLAGS` + `-DCOREMARK_UART`, run under `test/bench/run_coremark.sh ./sim`. PASS, self-check clean against EEMBC's published CRCs, **2.203 CoreMark/MHz**, 45,389,189 cycles — matching the already-published simulated figure (ADR-0154) to three decimals.

**What I did not do:** no iCESugar-Pro was available in this environment. `icesugar-prog` and the UART read that follows it need real hardware and were not run. Full details, including what did and didn't get verified, are in ADR-0165.

## Decisions made

- **Thread `ROM_WORDS` as one module parameter** rather than adding a second elaboration-time consistency check, since a single source of truth makes the two instantiations structurally unable to disagree (the class of defect the ticket names).
- **`icesugar.json` needs its own `chparam`** even though `icesugar_pro_top` never sets `ROM_WORDS` itself, because the parameter lives on `littlesoc`, the submodule it instantiates at whatever default `chparam` leaves it — missing this on the first pass built a 16 KB image against an unchanged 8 KB synthesis, caught by reading the DP16KD count back rather than trusting the build to fail.
- **Did not touch `SOC_EXPECT_EBR`/`ECP5_EXPECT_DP16KD` defaults** — they stay the 8 KB numbers for the default (unmeasured-override) build; the 16 KB numbers are reported in the ADR and are overridable on the command line the same way `SOC_ROM_WORDS` already is.
- **Kept the CoreMark pin/membership check as a third inline copy** in the Makefile (matching the existing precedent of `test/bench/run_coremark.sh` and `soc/compare/run_coremark_compare.sh` each carrying their own), rather than extracting a shared script that would touch those two pre-existing files outside this diff's stated scope. Its diagnostics now match the existing copies' quality (itemized missing/unlisted files) after a `/simplify` pass.

## Testing

- `make test`: 75/75, matching `test/EXPECTED_FAIL` exactly, both ways — run twice (before and after the `/simplify` pass)
- `make probe-gates`: 718 graded comparisons, all green, matching main's count
- `make window-test`: 44/44 elaborations, both frontends
- `make lint`: clean, both RVFI passes
- `make elaborate-strict`: clean
- `make soc-timing`: 13.17 MHz, RATCHET OK (confirms the ice40 `chparam` addition is a no-op at the default `SOC_ROM_WORDS`)
- `make ecp5-timing` at default (8 KB) and `SOC_ROM_WORDS=4096` (16 KB), both measured and reported above
- `make coremark-rom` and `make icesugar.bit ICESUGAR_ROM=noop-rom` at `SOC_ROM_WORDS=4096`: both build and place successfully
- cxxrtl run of the exact board binary (same `COREMARK_CFLAGS` + `-DCOREMARK_UART`): PASS, 2.203 CoreMark/MHz

**Not tested (needs the board):** `make icesugar-prog` and `make icesugar-read`/the `icesugar-coremark` UART report. A human with the iCESugar-Pro plugged in needs to run `make icesugar-coremark` and confirm the board prints the same self-check and CoreMark/MHz figure cxxrtl reports here.

Closes JEF-967.


thejefflarson:
**Integration notes (architect), for the record on a merged PR.**

Three things changed between review and merge; all three are recorded in ADR-0165 rather than only here.

**1. `ecp5-timing` was red for a real reason and is now green.** The refactor into `SOC_ROM_CHPARAM` wrapped two recipe-level `yosys -p` scripts over a backslash-newline **inside their single quotes**, which is not a shell continuation — both characters reached yosys, which read the lone `\` as a command and stopped with `ERROR: No such command: \`. Every ECP5 flow was dead. `make soc-timing` stayed green throughout because `SOC_SYNTH` is a make variable and make joins those, so the ice40 half hid the ECP5 half. Both scripts are one line now, with the reason as a tripwire above `SOC_ROM_CHPARAM`. The ADR's own measurement reproduces on the fixed spelling: `DP16KD` 36 at the default and **40 at `SOC_ROM_WORDS=4096`**.

**2. The chparam is emitted only when a caller overrode `SOC_ROM_WORDS`, because it was not free.** Measured against merged `main` with a byte-identical ROM image and the same tools, a chparam re-stating the unchanged 2048 moves the up5k SoC netlist **6289 → 6319 mapped cells** — yosys derives a parameterised copy of `littlesoc` and stops sharing across the hierarchy. Nothing in the tree would have caught it: `FIT_MAX_LC` ratchets the core alone, the EBR/SPRAM censuses count memories rather than logic, and `SOC_MIN_MHZ` is graded at one CI seed. This is ADR-0088's own rule — a parameter tied off to today's value is not free of the mapper — arriving in a new place. `$(origin SOC_ROM_WORDS)` answers "did someone ask for a different size" with no second copy of the number, and `test/memmap_test.sh` already grades the Makefile's `SOC_ROM_WORDS` equal to the RTL's default, so the shipping build is provably the one where the chparam is redundant. Emitted only on override, the shipping netlist is **cell-identical to `main` in all eleven types**.

**3. `coremark-rom` collided with #296's target of the same name**, at the same anchor, with a different linker script, different flags, and — in that one — no `PINNED.sha256` check before compiling the vendored sources. Make is last-wins for a redefined recipe and first-wins for a `?=` default beside it, so the merge would have paired one route's flags with the other's linker script, and the flags string is compiled into the image for EEMBC's disclosure line. This target is `coremark-rom-ecp5` with `COREMARK_ECP5_*` variables; #296's is `coremark-rom-up5k`. Both take `coremark-pin-check` as a prerequisite, running one script — `test/bench/coremark_pin_check.sh` — rather than a third hand-copy, because the route that flashes hardware must not be the one whose check was forgotten.

New `test/makefile_target_test.sh` grades the collision class off make's own `overriding recipe` warning, inside `make test`, with five forced-red probes.

PR #297, opened 2026-09-07
CANDIDATE: the learned successor-pair table, rebuilt and fully verified

thejefflarson:
**This is a CANDIDATE, not a landing.** It carries the RTL ADR-0113 declined, plus the verification that decline was re-taken without. Everything below passed; nothing is red. The decision to spend 144 placed cells and two block RAMs is yours.

Every number here was taken on **this branch against `origin/main` at `45f0b9b`**, one toolchain on both arms: OSS CAD Suite — Yosys 0.68+48 (`ff5817c34-dirty`), nextpnr-ice40, icetime (oss-cad-suite 20260811). Both benchmark halves and both timing arms are from this tree and this session.

## What the ADR-0113 amendment said it did not run — all of it, all green

| item | result |
|---|---|
| `make -C formal remeasure-fg` | **PASS. F = 6, G = 6 — both reproduce, unchanged.** Flip points 7 (`hang`) and 6 (`liveness`, at both trig 10 and trig 15). No `#derive` line moves, no generated depth moves. |
| `make -C formal check` | PASS — 86 checks, 86 pass, 0 fail; `EXPECTED_FAIL` exact, `EXPECTED_CHECKS` exact at 86 |
| `components_decoder` | PASS — successful proof by k-induction (basecase and induction both) |
| `components_executor` | PASS — successful proof by k-induction (engine_1, bitwuzla) |
| `components_accessor` | PASS — successful proof by k-induction |
| `components_pcloop` | PASS — successful proof by k-induction |
| `components_traps` | PASS — successful proof by k-induction |
| `components_busarbiter` | PASS — successful proof by k-induction |
| `complete` | PASS |
| `complete_cover` | PASS (the anti-vacuity control) |
| `imemcheck` / `dmemcheck` / `cover` | PASS |
| `nonperturbation` | PASS — gold and gate netlists structurally identical at 42 158 cells |
| `make cosim-suite` | PASS — 69/75 agreed, divergence list exact against `test/COSIM_EXPECTED_FAIL` |
| `make test` | PASS — 75/75, `test/EXPECTED_FAIL` exact, all probe gates green |
| `make probe-gates` | PASS (inside `make test`) |
| `make lint` | PASS — svlint clean in both passes |
| `make elaborate-strict` | PASS |
| `make mutation-check` | PASS — 11 mutations, each caught by exactly its paired detectors |

**F and G were the load-bearing one and they did not move.** The table changes what `operand_stall` sees, but not how long anything takes to retire: it removes operand-fetch cycles rather than adding a stall reason or lengthening a stage. Nothing downstream of `formal/checks.cfg` needs editing.

## Three landing debts the amendment did not name

1. **No formal source list read `rtl/pairtable.v`**, so no task that instantiates `rtl/littlecpu.v` could elaborate at all. Added to `formal/checks.cfg`, the five `.sby` lists, `formal/Makefile`'s `RTL_SOURCES` and `formal/check-nonperturbation.py`.
2. **`rtl/littlecpu.v` declared the pair nets after the instance that reads them.** yosys resolves that silently; iverilog refuses it with 56 elaboration errors. `make window-test` is the first thing that sees it, and the amendment's "`make test` is green except one probe" never reached it — probe-gates fails earlier and `make` stops. Declarations moved above the instance and out of the `RISCV_FORMAL` ifdef they had landed inside. **Netlist-neutral**: the SoC re-placed at the identical 5076 `ICESTORM_LC` / 22 `SB_RAM40_4K` / 75.93 ns before the rebase.
3. **`ECP5_EXPECT_DP16KD` gates and had to move 36 → 37.** The table is one `DP16KD` on ECP5 against two `SB_RAM40_4K` on ice40. With it bumped, all four ECP5 checks pass: DP16KD 37, TRELLIS_DPR16X4 32, MULT18X18D 4, and no block RAM reset driven by logic. ECP5 Fmax publishes at 35.95 MHz against the base's 33.87 (one seed each, no band derived for that part — not a claim).

## The mutation-coverage ruling, and why

`rtl/pairtable.v  unpaired  check`.

**A mutation of this file is architecturally undetectable by construction, and that is a fact about the design rather than a gap in the graders.** Everything the table produces is a *guess*, and `operand_stall` lets nothing issue until the pair the register file holds is the pair the issuing instruction reads. A wrong guess costs a cycle and can change no architectural value.

Measured rather than assumed, two mutations run:

- **Deleting the tag comparison outright** — `hit = entry[10]` in place of the tag equality, the one term that makes an entry belong to the address reading it — leaves **all 75 suite programs passing with the same 22 081 retires**, `EXPECTED_FAIL` still exact. The only thing that moves is `make cycles`'s operand column: 608 → 1448 cycles.
- **Always-hit with an inverted pair** did go red, as `uart.S BELOW-FLOOR retires`. That is `test/run_tests.sh`'s 5000-cycle budget running out (the operand column went 608 → 11 364), not a value comparison. A detector built on that would be grading slowness, and would break the first time `uart.S`'s margin moved.

So `check` is named for what it *grades*, not for going red: `formal/wrapper.v` instantiates `rtl/littlecpu.v`, which instantiates this file; the generated per-instruction checks compare `rvfi_rs1_rdata`/`rs2_rdata` against the spec model on every retire, and `reg_ch0` grades the register file under the write-through bypass this guess feeds. Those are what say the table cannot corrupt a retire whatever it answers. The reasoning is written into `test/MUTATION_COVERAGE` beside the line, per that file's own "measure or read before writing an entry".

`test/probe_gates.sh`'s control asserts the manifest count, so it moves 19 → 20 in the same commit.

## The measurements, re-taken on this tree

`origin/main` moved under this work — a tree-wide comment rewrite (#291) landed mid-run. This branch is rebased onto it and **everything above was re-run after the rebase**. Deleting comments changes yosys's generated cell names, which changes ABC factoring and nextpnr placement, so the netlist moved and the amendment's sweep no longer described this tree. **A fresh sixteen-seed paired sweep was owed and is here**, `default`…`15`, paired by seed, `SOC_MIN_MHZ=0` on the base arm so a failing placement does not stop it.

| | base `45f0b9b` | pair table |
|---|---|---|
| worst of sixteen | 83.38 ns / **11.99 MHz** | 80.11 ns / **12.48 MHz** |
| median | 80.445 ns / 12.43 MHz | 78.030 ns / 12.82 MHz |
| best | 78.10 ns / 12.80 MHz | 76.22 ns / 13.12 MHz |
| spread | 6.8% | 5.1% |
| **under 12.00 MHz** | **1 of 16** | **0 of 16** |
| `make fit` `ICESTORM_LC` | 4063 | 4157 (ratchet 4219) |
| SoC placed `ICESTORM_LC` | 4920 / 5280 (93.2%) | 5064 / 5280 (95.9%) |
| `SOC_EXPECT_EBR` | 20 of the part's 30 | 22 of the part's 30 |
| Dhrystone, 2000 runs | 1 506 772 cycles, CPI 1.59, 0.777 DMIPS/MHz | 1 370 424, CPI 1.44, **0.853** (−9.05%) |
| Dhrystone operand column | 238 562 cycles | 99 131 cycles |
| CoreMark | 45 389 189 cycles, 2.203 CoreMark/MHz | 40 897 881, **2.445** (−9.90%) |
| `.S` suite | 38 746 cycles, 21 911 retires, CPI 1.77 | 38 476, 22 081, CPI 1.74 (−0.70%) |
| suite operand column | 1 218 cycles | 608 cycles |

Per-seed: median **−3.05%**, **14 faster / 2 slower**. Median period −3.00%, worst placement −3.92%. **Read −3.00% of median as a null** — it is inside the ~3.6% edit-churn band, and the honest statement remains *the table does not cost period*.

**What is not inside a band is the tail, and the sign has changed since ADR-0113.** On today's `main` the base arm reads **11.99 MHz at seed 9, one of sixteen under the board clock**, and the candidate is **0 of 16**. That is worth reading twice: on this tree the arm that misses the requirement at its worst placement is the one *without* the table. (`make soc-timing`'s default seed on base reads 12.12 MHz, so CI is green either way — only seed 9 trips.)

**The product, both factors from this tree and this toolchain.** At each distribution's worst placement, 0.777 DMIPS/MHz × 11.99 MHz = 9.32 DMIPS against 0.853 × 12.48 = **10.65, +14.3%**; at the median, 9.66 against 10.93, **+13.2%**.

## The two things the amendment asserted but did not test

**The +2 `SB_RAM40_4K` claim holds**, verified against `SOC_EXPECT_EBR` and the real census: base 20, candidate 22, both `as declared` by `soc/cell_census.py`. **Remaining headroom is 8 of the part's 30**, against the base's 10.

**The competition with a larger fetch window is real for exactly one shape, and that shape is already declined.**

- **16 KB (32 `SB_RAM40_4K`) is unreachable either way.** Available is the free blocks plus the 16 the current ROM would release: 26 without the table, **24 with it**. Short by 6 before, short by 8 now. The table does not close a door that was open.
- **ADR-0145's 12 KB window (24 blocks, `SOC_EXPECT_EBR` 28) is where the competition bites.** ADR-0145 recorded it as fitting "with two block RAMs to spare"; with the table it is **30 of 30, exactly full, zero spare**. Not blocked — but its entire margin is gone.
- **There is no live demand on the up5k.** ADR-0145 already declined the 12 KB window on period, and `test/testbench.v`'s `ROM_WORDS = 4096` (16 KB) exists only so CoreMark can simulate — the SoC is `SOC_ROM_WORDS = 2048`. Were CoreMark ever made to fit 8 KB, the last reason to want one would go too.
- **And #295, which merged while this was being measured, moves the live demand off the part the table competes on.** It threads `ROM_WORDS` through `rtl/littlesoc.v` and measures 16 KB at DP16KD 36 → 40 of 56 on the iCESugar-Pro, where it fits and the `icesugar.bit` still clears nextpnr's 25 MHz check. **A wider fetch window is live on the ECP5, and that is the part with 56 block RAMs, not 30.** This table is one `DP16KD` there: 40 + 1 = 41 of 56 for a 16 KB window with the table, comfortable. So the up5k competition the amendment named is real only against a 12 KB shape already declined on period, and the place a wider window actually runs has room for both.

## Is it landable as it stands

**Yes, on the evidence.** Every gate this repo has passes, the two ratchets pass with margin, F and G are unmoved, all four verification legs are green, the worst of sixteen placements clears the board clock where the base's does not, and the three landing debts are paid in the commit. What remains is not a verification question but a spending one: 144 placed cells, 2.7 points of `ICESTORM_LC` occupancy (93.2% → 95.9%), and two of the eight remaining block RAMs, against +14.3% of the DMIPS product at the worst placement and +9.9% of CoreMark.

Two things a reviewer should weigh that no gate covers. **95.9% `ICESTORM_LC` occupancy** is the highest this SoC has run at; ADR-0121 measured that occupancy does not set the tail on this part, so it is a headroom question rather than a period one, but it is a small headroom. And **no ADR is written here** — this branch deliberately touches no `docs/`, because the ADR-0113 amendment lives on `thejefflarson/jef-942-…` and a second amendment from here would conflict with it. Landing this owes that amendment an update carrying these numbers.

**This branch is based on `45f0b9b`, and `main` moved twice more while these measurements ran** (#294 and #295). Nothing in either touches this diff's files, so a rebase should be mechanical — but every number above was taken against `45f0b9b`, and the sixteen-seed sweep in particular would be owed again if the rebase moves the netlist the way #291's comment cut did.

Not run, and why: `make suite-board`, `make dhrystone-board`, `make icesugar-*` (need the hardware); `make dual-smoke` / `make dual-ecp5-timing` (the dual top is a fourth design and the table is not wired into it); `make compare-*` (the cross-core product is stale on its own check against this tree already, and re-stamping it is a separate ticket's).


thejefflarson:
Declined — see the ADR-0113 amendment. Nothing measured here is retracted: both halves of the original decline are gone, every gate passes, F and G never moved. The objection is where the gain lands. Dhrystone falls 9.05% and CoreMark 9.90% while the suite falls 0.70%, which is a successor-pair table doing exactly what it does — paying where control flow repeats. 8 KB of text is the whole program on this machine, what fits is mostly straight-line, and nobody has measured the table there. Spending 144 cells and two of eight remaining block RAMs to reach 95.9% occupancy is the wrong currency while the divider, an interrupt controller and the flash data path are all still deferred.

The branch stays as the record of what was built. What would reopen it is `make cycles` on a program shaped like what this core actually runs.

PR #308, opened 2026-09-08
Pin the shipping placement, and grade SOC_MIN_MHZ against it

thejefflarson:
Closes JEF-993. Closes JEF-996.

`SOC_MIN_MHZ` now grades the **pinned shipping configuration** rather than the worst of an unseeded sweep. The requirement itself is unchanged at 12.0 MHz — the board crystal, whose next `SB_HFOSC` step down is 6 — only what is measured against it changes.

## Why

Three re-rolls of *identical RTL semantics* spanned a worst-of-sixteen of 12.44 / 11.99 / 12.28 MHz — about 3.7% against a 3.5% clearance, with one draw of three already failing. Yosys names generated cells after `file:line`; `abc9_ops.cc` topologically sorts them by that name string before handing ABC's order-sensitive `&dch`/`&if` their input boxes. So a comment moves the mapped netlist with nothing changing meaning. Open upstream as [yosys#3713](https://github.com/YosysHQ/yosys/issues/3713) since 2023, and deliberately exploited by [yosys#3277](https://github.com/YosysHQ/yosys/pull/3277)'s `rename -scramble-name`.

## What ships

- `soc/pin.json` — netlist digest, seed, measured MHz, toolchain stamp, and **the distribution the pin was chosen from**.
- `make soc-timing` with no `SOC_SEED` override places at the pinned seed and grades that placement. A digest mismatch fails as a distinct **RE-PIN NEEDED**, not as a timing miss — the two point a reader at different fixes.
- `make soc-seed-search` (off `make test` and off CI, like `make fit`) searches **high-entropy** seeds — sha256 of a counter, never 1..N — and writes a pin.
- Margin must clear 5% over `SOC_MIN_MHZ` (12.6 MHz) so a pin survives ordinary OSS CAD Suite drift.

Pinned: **seed 125781539 at 12.61 MHz, 5.08% margin**.

## On seeds 1..N

Seeds 1..16 are provably a 4-dimensional subspace of nextpnr-ice40's 64-dimensional xorshift state (`state(3) == state(1) XOR state(2)` over GF(2), with only five warm-up rounds). Whether that correlates *placements* is unmeasured and the decision does not lean on it. Empirically it did not change the picture: twelve high-entropy seeds read 11.91–12.61 MHz with 1 of 12 under 12.0, against the lattice's 11.99–12.80 with 1 of 16 under. **The worst of the very sweep that picked this pin is itself under 12.0** — the argument reproducing on its own data.

## Notes for review

- The pinned figure is a property of the artifact that ships, **not** an estimate of typical Fmax. Best-of-N is a biased order statistic, legitimate only because that exact configuration is what gets built.
- A pin is valid **per netlist**; the digest is what invalidates it.
- Amends ADR-0066 rather than rewriting it.
- **ADR renumbered 0170 → 0171**, since #304 claims 0170.
- `test/PROBES_EXPECTED` gains 10 lines here. Four sibling PRs (#304, #305, #306, #307) each grew that file from a different base, so it needs a hand-merge — union the added lines and re-sort, never take one side.
- Two commits in this branch's history are WIP blobs rescued from stalled agents and committed unreviewed; the final diff was verified against `main` to contain only this ticket's scope, with no `rtl/` changes.


thejefflarson:
**HELD at integration — `soc-timing`, a required check, is red, and structurally so.**

`soc/pin.json` — `RE-PIN NEEDED`, before nextpnr ran:

```
pinned:  sha256:cd83a72f...67f81dd2
current: sha256:105196cd...1b48b40ea
```

The pin was written with Homebrew `Yosys 0.68+post (c12172fbae)`. CI's `setup-oss-cad-suite`
supplied `Yosys 0.68+201 (7c26ff16c-dirty)` from the 20260908 suite. Two yosys builds, two
mappings, two digests. The mechanism worked exactly as designed and named the failure
correctly; the problem is that it can only ever be satisfied by the one toolchain that wrote
the pin.

**Why the margin does not cover this.** ADR-0171's decision 4 argues the 5% floor makes a pin
"survive ordinary toolchain drift... the pinned OSS CAD Suite floats". The margin protects the
*frequency* against drift. The gate is the *digest*, and the digest moves on any mapping
change at all — ADR-0170, in this same batch, proves the mapping moves on a comment. So the
margin and the gate are not measuring the same thing, and the one that fails first is the one
with no slack in it. CLAUDE.md names the OSS CAD Suite as "the one tool that floats"; this
change makes a required check a function of it.

This is not a flake and it is not fixable by re-pinning from a developer's machine: a pin taken
against today's suite goes red on the next release. Nothing else in the PR is in question —
`test`, `fit`, `ecp5-timing`, `lint`, `cosim`, `nonperturbation` are all green, `make test` and
`make probe-gates` (777 comparisons with this PR's ten) both pass locally, the ADR is careful
and the probes have real red directions.

**What would unblock it**, in rough order of how well each preserves what the ADR is for:

1. Gate the digest check on the recorded toolchain matching. A pin measured under a different
   yosys is not evidence about this netlist either way, so the honest reading is "no applicable
   pin" — fall back to the unseeded placement the job graded before, and say so in the output.
   Keeps the pin load-bearing where it applies and stops it gating where it cannot.
2. Pin the OSS CAD Suite for the `soc-timing` job specifically. Buys a real gate at the cost of
   making a floating tool a second pinned one, which is a repo-wide decision, not this ticket's.
3. Re-pin from CI (`make soc-seed-search` in a workflow). Twelve placements per re-pin and it
   still breaks on every suite release; the weakest of the three.

Whichever is taken, ADR-0171 needs a paragraph separating digest drift from frequency drift —
decision 4 currently conflates them, and that conflation is what let this ship.

PR #326, opened 2026-09-10
nanocpu: build the RV32E oracle patch, its diff-grader and the ill_e check

thejefflarson:
Closes JEF-984.

## What this does

Builds the RV32E verification infrastructure ahead of nanocpu's reshape to E (the reshape itself is a later ticket), per `docs/ideas/nanocpu-a-verified-core-on-a-2x2-tile.md` decision 9:

- **`nano/formal/rvfi_insn_check.sv`** — a fork of the pinned `checks/rvfi_insn_check.sv` differing by exactly one `ifdef RISCV_FORMAL_E` block, bounding the spec model's own **decoded** register addresses (`spec_rs1_addr`/`spec_rs2_addr`/`spec_rd_addr`), never raw instruction bits — bounding raw bits would silently narrow `lui`/`auipc`/`jal`/`csrr*i`/`fence`'s immediate space, since those encodings don't have a register in that field position at all.
- **`nano/formal/check-rvfi-insn-check.py`** — a diff-grader modelled on `formal/check-genchecks.py`: undoes the one documented edit and requires byte equality with the pin. `make -C nano/formal check-rvfi-insn-check`, wired into the `monitor-freshness` CI job.
- **`nano/formal/ill_e.sv`** — a hand-written check for the RV32E register-naming restriction (an rv32i-legal encoding naming x16-x31 must trap, write no register, write no memory), checked against a small reference implementation (`ill_e_top`) rather than `nano.v`, which is still the unreshaped 32-register donor with no CSR/`mcause` layer (ADR-0167). Nine cover goals, one per base-ISA opcode class.
- **`nano/formal/ill-e-probe.py`** — mutates the assertion's three clauses (`trap`, `rd_addr`, `mem_write`) and the cover-reachability gate independently, requiring each to fail at its own pinned line. Wired as a prerequisite of `ill_e_cover` only (not `ill_e` alone), matching `complete_cover`'s own asymmetry against `complete` so CI spends the solver once rather than twice.
- **Pin-bump path** — `formal/bump-riscv-formal-pin.sh` now also checks `rvfi_insn_check.sv` against the freshly-cloned upstream file at a new pin and reports STALE/in-sync in the issue body; `formal/propose-pin-bump.sh` and the pin-bump workflow gained a comment naming the coupling.
- **ADR-0174** records the exact diff, why the assumption sits where it does, the two graders, and states plainly that this patch remains the one place in nanocpu where a green result can mean "stopped asking."

## Decisions made (flagging for review)

- **`ill_e` checks a reference model, not `nano.v`.** `nano.v` is still RV32IMC (32 registers, no CSR/trap layer per ADR-0167), so a property requiring "traps with cause 2" cannot be checked against it today — there's no `mcause` signal, and 32-register `nano.v` legitimately retires every x16-31 encoding without trapping. The reference (`ill_e_top`) proves the *property itself* is sound and non-vacuous; swapping in the reshaped `nano.v` (once it has real x16-31 trapping and a CSR layer) and adding the cause-2 comparison is deferred to the reshape ticket. This is stated in both the file and the ADR, not silently deferred.
- **`test/comment_density_test.py` gained an `EXCLUDE_FILES` entry** for `rvfi_insn_check.sv`, mirroring the existing `formal/genchecks-local.py` exemption: it's a byte-diff-graded fork of a licensed upstream file, and the mandatory unmodified license banner alone puts a 212-line file over a 5% budget. `docs/comment-budget.md` documents this.

## Testing

- `make probe-gates` — green, **853 graded comparisons** (837 on main + 16 new), every failure path executed.
- `make adr-numbering-test` — green, 169 ADRs.
- `make comment-density-test` — green, 348 files scanned.
- `make tracked-ignored-test` — green.
- `make pin-bump-test` — green, 13 checks.
- `make -C nano/formal check-rvfi-insn-check` — reports a match against the pin.
- `make -C nano/formal ill_e` and `ill_e_cover` — both PASS (real `sby`/boolector runs).
- `nano/formal/ill-e-probe.py` run directly against the real tree — shipping reference passes all assertions and reaches all 9 cover goals; each of the four mutations (`no-trap`, `no-rd-clear`, `no-mem-clear`, `no-illegal`) fails for its own stated reason.
- Manual security self-review of the diff (no Critical/High findings — pure file-diff tooling and subprocess calls with list args, no `shell=True`, no untrusted input); `/simplify`-style 4-agent parallel review applied (dropped a redundant parameter, fixed a double-solver-invocation in CI, and closed a soundness gap where two of `ill_e`'s three assertion clauses had no independent forced-red direction).

Not run: `make -C formal check` / the generated riscv-formal check suite (untouched by this ticket — `nano.v`'s ISA is unchanged) and anything needing a real board.


thejefflarson:
## HELD — not merged. Green CI, but the two controls the ADR names as load-bearing are not both graded.

#324 and #325 are merged; this one stays open. I re-reviewed the diff independently rather than
taking the earlier list on trust, and it changed one item's premise and raised another's severity.
Severity-ordered, `file:line — problem — fix`.

---

### 1. `test/probe_gates.sh` — the drift-grader has no red direction, and the ADR says it is one of only two things holding the property up

`nano/formal/check-rvfi-insn-check.py` appears **zero** times in `test/probe_gates.sh`; all sixteen
new labels belong to `ill-e-probe.py`. ADR-0174's own Consequence section is what makes this
blocking rather than a nit:

> the only thing standing between "assumes correctly" and "assumes away the property" is the
> diff-grader's byte equality against the pin, plus `ill_e`'s own proof that the property it
> states is non-vacuous.

One of those two controls has never once been shown to fail. That is the exact defect class
CLAUDE.md names — "A grader that cannot fail is not a grader" — over the file the ADR says the
whole patch rests on.

**Fix:** a forced-red probe per failure arm of `check-rvfi-insn-check.py` — a fork with drift
outside the documented block, a fork with the block deleted, and a missing/unreadable pin copy —
each required to fail for its own reason, with labels in `test/PROBES_EXPECTED`.

### 2. `nano/formal/ill_e.sv:28-32` — the 27 class/field memberships are graded by nothing

`uses_rd`/`uses_rs1`/`uses_rs2` are the entire content of the restriction, and the nine cover goals
at `:52-60` cannot see an error in them: each asks only that `e_illegal && class_X` be reachable,
and `e_illegal` at `:34` is a three-way OR. Drop `class_op` from `uses_rs2` and `add x1, x1, x16`
stops being E-illegal while `cover (live && e_illegal && class_op)` is still reached through
`rd[4]` or `rs1[4]`. Every probe reports green.

**Fix:** make the cover goals per (class, field) rather than per class — reach each `class_X`
through each field it is claimed to use — or add a mutation per membership to `ill-e-probe.py`.

### 3. `nano/formal/rvfi_insn_check.sv:172-174` vs `ill_e.sv:12` — the assumption is global, its compensating check is nine 32-bit classes, and nothing couples them

Correcting the earlier framing: `RISCV_FORMAL_E` is **defined nowhere in the tree** — not in
`nano/formal/checks.cfg`'s `[defines]`, not on any command line — so the `assume` is inert today
and nothing is currently removed from BMC's reach. ADR-0174:159 is honest that wiring it up is the
reshape's job. So this is not a live hole, and I am not holding on the claim that a reshaped E core
passes CI while retiring `csrrw x20, mstatus, x1`.

What I am holding on is that **nothing in the tree will go red on the day someone defines the
macro.** `assume(spec_rs1_addr < 16 && spec_rs2_addr < 16 && spec_rd_addr < 16)` sits in the file
every generated per-instruction check reads, so it bounds SYSTEM, MISC-MEM and every compressed
encoding that names a full 5-bit register. `ill_e.sv` gates on `is32` at `:12` and has no
`class_system`, no `class_miscmem` and no compressed class. The coupling — "this macro may only be
defined while `ill_e`'s class list covers the generated set's encodings" — is stated in prose and
enforced by nothing, which is how it will land silently in the reshape.

**Fix:** make it fail closed. Either widen `ill_e` to the encodings the assumption bounds before
the macro can be defined, or add a grader that refuses a tree in which `RISCV_FORMAL_E` is defined
while `ill_e`'s class list is narrower than the generated check set — so the reshape trips it
rather than inherits it.

### 4. `docs/adr/0174-...md:90-91` — the ADR states an equivalence that is false, and it conceals exactly the gap in item 3

> the nine base-ISA opcode classes (`LOAD`, `OP-IMM`, `AUIPC`, `STORE`, `OP`, `LUI`, `BRANCH`,
> `JALR`, `JAL`, the same list `complete.sv`'s own cover goals already use)

`formal/complete.sv` has **twelve** cover goals: those nine plus `RVC quadrant 0/1/2`. `ill_e`'s
list is a strict subset that drops precisely the three compressed quadrants — the encodings item 3
is about. Under CLAUDE.md the ADR is the record a future reader relies on; a false claim of
equivalence there is what makes the reshape's author believe the class list is already complete.

**Fix:** say what the list actually is and why the RVC quadrants are out of scope for this ticket,
and state the precondition for defining `RISCV_FORMAL_E`.

---

### Should-fix, not blocking

- `ill-e-probe.py`'s anti-vacuity control accepts a mutant that leaves **one** of nine goals
  unreached; its sibling `complete-cover-probe.py` demands all. Match the sibling.
- Two probe arms guarding the empty-set / log-phrasing failure have no red direction.
- `ill_e` has no probe prerequisite while `ill_e_cover` does — a control runnable separately from
  the thing it controls eventually is not run, which the probe's own header argues.
- `formal/bump-riscv-formal-pin.sh`'s new drift check reads the clone's **working tree** rather
  than the resolved `$UPSTREAM_SHA`, so an upstream push between `ls-remote` and `clone` yields a
  false "still in sync with the new pin".
- The grader report reaches a GitHub issue body with no size cap and no fence escaping, after
  `git push` has already run.

### Not at issue

The `spec_*_addr`-versus-raw-bits distinction in `rvfi_insn_check.sv`'s header is right, and worth
keeping as written — bounding `instr[19:15]`/`instr[24:20]` would silently narrow the immediate
space on `lui`/`auipc`/`jal`/`csrr*i`/`fence`. The `ci.yml` change adds only `run:` steps to two
existing jobs, so the fork-PR ternary keeping untrusted code off the self-hosted pool is untouched,
and no graded command sits in a pipeline. `PROBES_EXPECTED` is hand-inserted in correct `LC_ALL=C`
position with the header byte-identical.

Note that main has moved twice under this branch (#325 `6415657`, #324 `457d492`), and both touched
`test/PROBES_EXPECTED` and `test/probe_gates.sh`. Rebase before the next round; the resolution on
that file is always keep both sides' labels, sort under `LC_ALL=C`, never dedupe.

thejefflarson:
Pushed a response to the architect review (force-pushed after rebasing onto main @ 457d492, which had #324/#325 land while this was in flight).

## Blocking 1 — the assumption's global reach vs ill_e's coverage
Widened `ill_e.sv` from 9 to 12 classes: added `MISC-MEM` (decoded, deliberately kept out of the OR-list since its `rd`/`rs1` bits are reserved, not registers) and `SYSTEM`'s two CSR sub-shapes (`csrrw/s/c` reads both `rd`/`rs1`; `csrrwi/si/ci` reads `rd` only — `rs1`'s field position is a 5-bit uimm there). Added 3 compressed classes (CR/CI/CSS) reading a full 5-bit register field, matched field-for-field against the pinned `insn_c_*.v` spec models. This is option 1 from the follow-up correction (the assumption is inert today — `RISCV_FORMAL_E` is defined nowhere in the tree — but the widening is comprehensive against everything the pin's 70 checks + `rvfi_insn_check.sv`'s routing can reach today, so it's the right call regardless of the correction's timing). ADR-0174 records why the wiring tripwire doesn't *also* need a class-coverage check on top of this.

## Blocking 2 — the false "same list as complete.sv" claim
Fixed. ADR-0174 now states the real relationship: 9 of `ill_e`'s classes happen to name the same 9 base opcodes `complete.sv` covers (coincidence of overlap, not a shared source), and `complete.sv` covers 3 RVC quadrants generically where `ill_e` covers 3 specific full-register-field compressed classes instead.

## Blocking 3 — the drift-grader had no forced-red probe
Added to `test/probe_gates.sh`: a control (fork matches pin) and a hand-drifted-fork case requiring `has DRIFTED` with the actual diff printed.

## Blocking 4 — the 27 (class, field) memberships graded by nothing
Restated `e_illegal` as 23 individually named, isolated terms (`ill_op_rs2` = "OP reads rs2", isolated against its sibling fields) instead of `uses_rd`/`uses_rs1`/`uses_rs2`. `ill-e-probe.py` now reads the OR-list out of the file by regex and adds a `drop-<term>` mutation per membership (23 runs), each requiring *exactly* its own isolated cover goal lost — closing the exact gap in your worked example (deleting `class_op` from `uses_rs2` now makes `ill_op_rs2`'s own cover goal go unreached, and only that one).

## Should-fix items
- `tie-low`'s check: measured (at both `depth 2` and `complete_cover.sby`'s `depth 100`) that sby's cover engine reports only ONE of 23 independently-gated unreached goals per run, not all of them the way `complete-cover-probe.py`'s single-shared-gate stalled-bus case does — so `tie-low` requires "at least one real one," and the `drop-<term>` loop is what's precise. Documented in both the probe's docstring and the ADR.
- Added stub coverage for the two previously-untested branches (`len(reached) != len(goals)`, and the analogous wrong-site check on the `tie-low`/`drop-<term>` arms).
- `ill_e` and `ill_e_cover` both depend on the probe now (a hand-written reference proves nothing about itself without it, unlike `complete`); collapsed into one CI step (`make -C nano/formal ill_e ill_e_cover`) so the shared PHONY prerequisite still runs once, not twice.
- `bump-riscv-formal-pin.sh` reads the upstream fork with `git show "$UPSTREAM_SHA:..."` now, not the clone's working tree (which can have moved between `ls-remote` and `clone`).
- The drift report is capped at 300 lines and tilde-fenced (not backtick-fenced — the diffed file is SystemVerilog, whose own preprocessor directives are backtick-prefixed).

## Tautology
ADR-0174 now says plainly: the three assertion clauses (`trap`, `rd_addr`, `mem_write`) are tautologies by construction (`assign trap = e_illegal` under `if (e_illegal) assert(trap)`), the mutations against them prove the *probe* works, not that the property is falsifiable against `ill_e_top`, and a green `ill_e` means the restriction as enumerated is internally consistent and every membership is independently reachable — never that `nano.v` has it. Added `test/ill_e_wiring_test.py`, on `make test`'s path: red if `RISCV_FORMAL_E` ever goes live in `checks.cfg` while `ill_e.sv` still checks the reference model instead of the real core. No-op today (both halves false).

## Testing
- `make probe-gates`: **866** graded comparisons (841 on main + 25 added), every failure path executed, green.
- `make adr-numbering-test`, `make comment-density-test` (349 files), `make tracked-ignored-test`, `make retired-term-test`: all green.
- `make test` (the full suite, rebased onto 457d492): green end to end, 75/75 `.S`/`.c` programs pass, `Failure list matches test/EXPECTED_FAIL exactly`.
- `make -C nano/formal ill_e ill_e_cover` and `nano/formal/ill-e-probe.py` run directly against real `sby`/boolector: shipping reference passes all 3 assertion clauses and reaches all 24 cover goals; all 23 `drop-<term>` mutants and 3 assertion mutants fail at exactly their own pinned line; `tie-low` fails and names a real membership goal.
- Manually verified the attack from your worked example is now caught: dropped `ill_op_rs2` from the OR-list by hand, confirmed the assert side still passes vacuously (as expected — that's the disclosed tautology) while the cover side goes red and names exactly `ill_op_rs2`'s own line, nothing else.

PR #343, opened 2026-09-13
Fix nano's real divider and add its differential oracle

thejefflarson:
## Summary

nano's non-`ALTOPS` divide state (`nano/nano.v`) had three independent bugs, none of them ever exercised by riscv-formal's generated checks (which run under `RISCV_FORMAL_ALTOPS` and substitute a cheap function of the operands for the real divider) and left baselined `divide.S MONITOR-ERROR 105` by ADR-0180 instead of fixed:

1. The shift-subtract comparison was backwards (`x <= y`, not `x >= y`), corrupting the remainder on the very first loop iteration.
2. The iteration count was 65 against the 32 a divisor pre-shifted by 31 actually needs.
3. `DIV`/`REM` never converted a negative operand to magnitude before dividing.

`nano/asm/divide.S` itself carried a fourth, unrelated bug: its `rem -20, 6` vector expected `2`, but RISC-V's remainder takes the dividend's sign, so the correct value is `-2` (-20 = 6*(-3) + (-2)).

## What changed

- `nano/nano.v`: fixed the comparison, the iteration count, and added magnitude conversion for signed `DIV`/`REM`; `DIV`'s sign restore gates on `rs2 != 0` rather than adding a second arm for the divide-by-zero constant.
- `nano/asm/divide.S`: fixed the `rem` test vector; `nano/asm/EXPECTED_FAIL` is now empty; `OBSERVED_FLOOR`'s `divide.S` line moves 5 → 35 (a full passing run).
- New: `nano/tb/nano_exec_cxxrtl.cc` + `nano/tb/nano_exec_tb.v`, a randomized differential oracle for nano's real multiply and divide, in the shape of `test/exec_tb.v`: hand-computed literals self-tested before any RTL vector runs, then 2,000 random vectors per op plus twelve directed corners (`INT_MIN / -1`, divide/remainder by zero — including by zero with a negative dividend — and both operand-sign combinations). nano has no separate executor module and `nano.v` is not iverilog-elaborable, so the oracle drives the real `riscv` state machine by poking `regs`/`instr`/`cpu_state` directly through cxxrtl's `debug_items`, walking one instruction from `decode_instr` through retirement.
- New: `nano/tb/nano_exec_probe.sh` (+ `nano_exec_run.sh` helper) — the forced-red prerequisite: reintroduces each of the three RTL bugs into a scratch copy of `nano.v` and requires the oracle to report a mismatch. All three are caught.
- `nano/tb.mk` + `Makefile`: wires `nano-exec-sim`/`nano-exec-probe`/`nano-exec-test` in, joining `make test`.
- `nano/nano.mk`: `NANO_MAX_UM2` moves 60759 → 61412 (measured 61411.4 µm², rounded up per ADR-0178's convention).
- `docs/adr/0181-...md`: records the defect, the fix, and what does *not* move (F/G reproduce unchanged at 12/10, since every nano formal check runs under `ALTOPS` or is too shallow to reach the real divide loop's completion).

## Testing

- `make nano-test`: 6/6 programs pass, `divide.S` included, `EXPECTED_FAIL` empty.
- `make nano-exec-test`: forced-red probe catches all three reintroduced bugs; oracle then passes 0 mismatches over 12 directed + 2,000×8 randomized vectors.
- `make nano-area`: 61411.4 / 61412.0 µm² — OK.
- `make -C nano/formal remeasure-fg`: F=12, G=10, both reproduce.
- `make -C nano/formal check`: 82 checks, matches `EXPECTED_FAIL`/`EXPECTED_CHECKS` exactly (unchanged from before this PR).
- `make -C nano/formal all`: exit 0.
- `make elaborate-strict`: exit 0.
- `make test` (full littlecpu suite + 903 probe-gates comparisons + nano-test + nano-exec-test): exit 0, run twice clean (once before, once after a `/simplify` cleanup pass).

## Scope notes

- Ran `/simplify` on the changed files: applied the altitude fix (removed a redundant special-case literal from `DIV`'s divide-by-zero handling by gating the existing sign-restore predicate on `rs2 != 0` instead — verified equivalent for every input), factored a repeated `is_div || is_rem` condition, merged duplicated mismatch-reporting code in the C++ oracle, and aligned `nano_exec_run.sh`'s compiler flags with the Makefile's to close a warning-detection gap between the two build paths. Skipped: routing the probe's rebuild through `make` instead of a standalone shell recipe (would risk colliding with `make test`'s own `nano-exec-sim` build artifacts), and expressing the three mutations as `test/mutations/*.patch` files (would pull nano into littlecpu's `rtl/`-scoped mutation-coverage manifest, out of scope for this fix).
- This is the only ticket editing `nano/nano.v` in flight (per coordination note); `nano/formal/checks.cfg`'s first line and `test/asm` were left untouched.

Closes JEF-1015


thejefflarson:
Addressed the review's grader-gap findings (all Medium severity, none Critical/High) in e50b58d:

1. **rd not poisoned / colliding directed expectations** — `run_op()` now seeds `regs.curr[rd] = ~expected` before every vector (directed and random), skipped only when `rd` aliases `rs1`/`rs2` (where the stale operand itself serves as the poison). Added directed vectors with `rd == rs1` (`div x1, x1, x2`) and `rs1 == rs2` (`div x4, x1, x1`).
2. **Multiply half never shown red** — added the `mulhsu-sign-extends-rs2` mutation to `nano_exec_probe.sh` (MULHSU sign-extending rs2 instead of zero-extending it, mirroring MULH's own bug class) and four new self-tested reference literals verified by hand: `mul(-1,-1)=1`, `mulhsu(-1,-1)=-1`, `mulhsu(INT_MIN,INT_MIN)=0xc0000000`, `mulh(INT_MIN,INT_MIN)=0x40000000`. All four mutations (3 divide + 1 multiply) are now caught.
3. **Uniform random operands miss corners** — `random_operand()` mixes in 0/1/INT_MIN/INT_MAX and small magnitudes (`rng() >> (rng() % 32)`) alongside full-uniform draws; added directed multiply rows at the same corners.
4. **Wire poke fragility + no shape validation** — `instr`/`cpu_state` pokes now set both cxxrtl's `curr` and `next` storage (a wire's two separate arrays), not just `curr`. Every debug item lookup now goes through a `must_find` + `shape_ok` pair (size==1, type, width, and for `regs` also depth/zero_at), so a renamed or reshaped signal fails by name instead of silently misreading. Verified both failure modes by hand (a rename produces `error: 'dut cpu_state' not found`; the poke fix verified against the passing full run).
5. **Two build paths for one binary** — `nano-exec-test` now runs `nano/tb/nano_exec_run.sh nano/nano.v` directly; the duplicate `nano/tb/nano_exec_rtl.cc`/`nano-exec-sim` Makefile rules and their `.gitignore` lines are gone.
6. **-Wno-unused-parameter too broad** — removed; only the `#include` of the generated cxxrtl code is now wrapped in `#pragma GCC diagnostic push/ignored "-Wunused-parameter"/pop`, documented as CLAUDE.md's third allowlisted warning exception.
7. **Unquoted paths** — `nano_exec_run.sh`'s `yosys -p` script now quotes each path the way `nano/synth_script.sh` does; `tb.mk`'s surviving `nano-sim` recipe's `-isystem` is quoted too.

Re-ran and confirmed green: `make nano-exec-test` (19 directed + 2000×8 random vectors, 0 mismatches; all 4 forced-red mutations caught), `make nano-test` (6/6), `make elaborate-strict`, `make nano-area` (unaffected, 61411.4/61412.0), and the full `make test` (75/75 littlecpu programs, 903/903 probe-gates comparisons, exit 0).

Not touched: nothing armed or merged.

thejefflarson:
## Architect review: READY, pending required CI on e7ec5ff

I read the diff, re-ran the gates in the engineer's worktree, and pushed one small follow-up commit (e7ec5ff: docs plus one comment line, no RTL change).

### Verified independently
- **`divide.S`'s rem vector is right against the spec, not just against the core.** The M extension says a nonzero REM result takes the dividend's sign: −20 = 6·(−3) + (−2), so the answer is −2. I hand-checked the other four vectors too (20/6 = 3, −20/6 = −3, divu = 3, remu = 2).
- **The RTL divider is correct.** It is a restoring shift-subtract: the divisor starts pre-shifted to `|b| << 31`, so 32 compare-and-shift steps yield the quotient in `mul_div_store[31:0]` and the remainder in `mul_div_x`. Negating INT_MIN gives 0x80000000, which is its correct unsigned magnitude. REM by zero returns the dividend and REM of INT_MIN/−1 returns 0, both with no special case. DIV's `rs2 != 0` gate is exactly what keeps the all-ones by-zero result from being negated.
- **The probe is on `make test`'s path and follows an accepted shape.** It is a prerequisite of `nano-exec-test`, the same shape as the zkt probes that gate `components_decoder` and `components_executor`, so I accept it as is. It fails a mutation whose sed matches nothing in `nano.v`, and it counts only a `MISMATCH` as caught; a TRAP or HANG does not count. Re-run: all 4 mutations caught; the real core shows 0 mismatches over 19 directed vectors plus 2,000 random draws for each of the 8 operations; 10.7 s wall clock.
- **The C++ reference models are free of undefined behaviour.** Division by zero and INT_MIN/−1 are handled before any `/` or `%`. Every product fits in `int64_t`. Casting a `uint32_t` above INT32_MAX to `int32_t` is implementation-defined in C++17 (not undefined) and wraps modulo 2³² on clang. I hand-checked all 22 self-test literals.
- **Numbering, density, and ticket IDs are clean.** ADR-0181 has a unique number on this tree and one README row. No ticket IDs appear in the code, comments, ADR or commit messages. Every touched file is at or under 5% comment lines.
- **`make nano-area` reproduces the new ratchet:** 61,411.4 of 61,412.0 µm².

1. **Medium, fixed in e7ec5ff: ADR-0181 and its README row had stale counts.** After round 2 they still said "twelve directed corners" and "three mutations, all three caught"; the true counts are 19 directed vectors and 4 mutations, the fourth being the MULHSU sign mutation.
2. **Medium, fixed in e7ec5ff: ADR-0181 omitted its effect on ADR-0179's freeze-line projection.** The fix adds 653 µm² locally. ADR-0179's formula is (local area + 11,500 µm² for the two unbuilt layers) × 0.915. That moves the projection from 66,116 to about 66,714 µm², 708 over the 66,006 line instead of 110. The ADR now says this, and that ADR-0179's pre-committed cut order is what answers it; a correct divider is not traded for area.
3. **Low, fixed in e7ec5ff: the new warning pragma had no reason at its site.** `nano/tb/nano_exec_cxxrtl.cc` suppresses `-Wunused-parameter` for the generated include, and the CLAUDE.md rule this PR adds requires each exception to be "documented where it is allowlisted". One comment line now gives the reason.
4. **Low, not fixed, needs a follow-up ticket: yosys warnings on nano's simulation paths go ungraded.** `nano/tb/nano_exec_run.sh` writes yosys's output to a log and never checks it for `Warning:`, and `nano/tb.mk`'s `nano_rtl.cc` rule does the same. Neither nano testbench is part of `elaborate-strict`. This predates the PR, so it is not a hold reason.

### Coordination
- **#345 also adds an ADR numbered 0181** (`0181-nanocpus-benchmarks-trapped-on-an-uninitialized-gp.md`). #343 merges first, so #345 must renumber to 0182 before it can land. Otherwise `adr-numbering-test` goes red, and `docs/adr/README.md` will also conflict textually. #345 and #346 both edit `nano/tb.mk`, which this PR also edits.

### Gates re-run after e7ec5ff
`nano-exec-test` (probe included), `comment-density-test`, `adr-numbering-test`, `retired-term-test`, `band-source-test`: all exit 0.

PR #347, opened 2026-09-13
Re-take the cross-core product, fresh, on both parts

thejefflarson:
Closes JEF-935.

## Second round: the stamp itself is dropped from this PR

Security/design review found that the first version of this PR re-took the
measurement and committed `soc/compare/product.json` with `base` set to a
commit on this PR's own branch. That does not survive review: this repo
squash-merges and deletes the source branch on every merge, so the instant
this PR merges, `base` names a commit no checkout can resolve. `product.json`
is now unchanged from `main`; the real stamp lands from `main` via
`.github/workflows/compare-product-schedule.yml`'s scheduled dispatch after
this merges (a post-merge step, not run from this branch), which is also
this ticket's `workflow_dispatch` acceptance criterion.

That review also found four more defects the first stamp would have carried
right alongside it, plus several hardening items -- all fixed here (see
`docs/adr/0183-...md` for the full narrative):

- `product_diff.py`'s `is_news()` had product_check.py's own self-watching
  bug, one level up -- fixed by threading the `artifact` exclusion through.
- An orphaned `base` (the same squash-merge problem, one snapshot later)
  made `--require-news` refuse instead of recover; it now reports news.
- up5k's product multiplied by each core's own placed Fmax instead of the
  12 MHz step CLAUDE.md's own rule requires (1.990x where cycles alone give
  1.14x) -- fixed with a new `--step-mhz` flag on `product_write.py`.
- The schedule workflow's pair-derivation could fail open through a process
  substitution `set -e` cannot see -- captured into a variable instead.
- Both Python scripts validate `base` as a 40-character SHA and use
  `--end-of-options`; both wrap `main()` so an uncaught bug can't collide
  with "stale"/"no news"; `measure_coremark()` propagates a write failure
  instead of falling back to `not_yet_measured` over an already-measured
  pair; ECP5 pairs stamp `ecp5_part`/`ecp5_target_mhz`/the Trellis digest;
  the schedule workflow gets `persist-credentials: false`, `GH_TOKEN`
  confined to its publish step with `gh auth setup-git`, and a main-only
  job guard, graded by a new `test/compare_product_schedule_token_test.py`.

CLAUDE.md no longer claims a fresh stamp. It states that the harness fix
lands here and the stamp from main, and corrects six absolute DMIPS/CoreMark
figures that were off by a rounding-methodology bug independent of which
session's clocks were used (multiply at full precision, round once):
littlecpu 9.39 DMIPS/Hazard3 18.03 CoreMark at up5k, VexRiscv 47.42 DMIPS/
littlecpu 73.89/VexRiscv 123.91/Hazard3 73.45 CoreMark at ECP5. It also now
says Hazard3's Dhrystone row is not stamped by `run_product.sh` at all (only
littlecpu and VexRiscv are in that pair).

ADR renumbered 0181 -> 0183 (0181/0182 are claimed by other open PRs).
Rebasing onto main after #343 merged (which claimed the real 0181) hit the
exact `docs/adr/README.md` union-merge duplicate-row problem the coordinator
flagged from another session's experience -- fixed in a follow-up commit,
and `grep -oE '^\| \[[0-9]{4}\]' docs/adr/README.md | sort | uniq -d` now
prints nothing.

## First round (original PR body, still accurate for the harness work)

`soc/compare/run_product.sh` now sweeps littlecpu, VexRiscv and Hazard3's
clocks on both up5k and ECP5 (previously up5k only), and CoreMark's pair
gains a second column -- VexRiscv's, alongside Hazard3's. An ECP5 number is
its own pair (`dhrystone_ecp5`/`coremark_ecp5`) rather than a schema change,
so it is never merged with or averaged against up5k's (ADR-0160, ADR-0171).

## Testing

- `make test`: 75/75 passed, failure list matches `test/EXPECTED_FAIL`
  exactly -- run repeatedly across this round's changes, most recently on
  the rebased tree after the README fix.
- `test/adr_numbering_test.sh` and the manual duplicate-row grep above, both
  clean.
- Every new graded comparison has a forced-red probe in `test/probe_gates.sh`
  with a hand-sorted `test/PROBES_EXPECTED` line: malformed `base` (both
  scripts), a reserved `--field` name, `--field`/`--step-mhz` computing and
  stamping correctly, `--require-news`'s own artifact not counting as its
  own news (both directions), an orphaned base recovering as news, and all
  seven directions of the new workflow token test.
- Verified the committed tree (not just the working tree) is what actually
  passes the comment-density budget, after catching and fixing a
  stage/working-tree mismatch from an earlier `git add -A` during this round.

## Scope notes

- The scheduled workflow's `workflow_dispatch` run is a post-merge step, not
  run from this branch (it has `contents: write` and opens PRs).
- `test/adr_numbering_test.sh` does not catch two rows claiming the same
  number when the file list itself has no collision (each row's own linked
  file exists) -- only a manual grep catches it today. Worth a follow-up in
  its own right, flagged by the coordinator as something another session
  also hit; not fixed here since it's outside this ticket's scope.
- Out of scope for this round, filed as follow-ups per the coordinator:
  replacing the commit `base` with a content digest, watching
  `test/bench`/`Makefile`/`formal/pin.mk` and comparing the tools block, the
  dirty-flag check-then-use race, splitting the workflow into measure/publish
  jobs, `run_product.sh`'s seed/`CM_ITERATIONS`/`eval` hardening, and
  `setup-oss-cad-suite`'s expression injection.


thejefflarson:
## Architect review (integration): READY once CI is green on the fix-up commit

I reviewed the diff against origin/main, traced the post-merge dispatch step by step, and pushed one fix-up commit. Severity-ordered:

### Fixed on the branch (fix-up commit f51f25a)

1. **High, pre-existing, and it would have failed the post-merge dispatch in its first minute.** `Makefile` `compare.$(COMPARE_CORE).json` and `compare_ecp5.$(COMPARE_CORE).json` continued their `yosys -p` scripts with a backslash-newline *inside* the single quotes. The only earlier pool run of this workflow (run 34111542512, 2026-09-07) died 39 s in on `ERROR: No such command: \` at exactly that line. Reproduced here: macOS `/usr/bin/make` (GNU Make 3.81) hides the backslash from yosys, so every local run, the engineer's twelve-seed sweep included, passed; GNU Make 4.4.1 passes both characters through, and yosys given them directly fails the same way. Fix: both scripts on one line, the rule CLAUDE.md already states for `SOC_ROM_CHPARAM`. Verified under `gmake` 4.4.1: littlecpu up5k (step gate PASS, 13.01 MHz), littlecpu ECP5 (34.11 MHz) and VexRiscv up5k all synthesise, place and grade.
2. **Medium.** `timeout-minutes: 150`, with a comment saying the run was "measured at just under 70 minutes on this pool". The only measurement was about 60 min on a developer Mac; the workflow has never reached its sweep on the pool. That Mac runs `make ecp5-timing` in 31 s against the pool CI job's 72 s (2.3x), which puts the re-take near 140 min on the pool. Raised to 300, with the comment corrected.
3. **Medium.** `soc/compare/product_write.py --step-mhz` credited a core the step without checking it: a hand run with littlecpu at 11.1 MHz got 12 MHz and rc 0. It now refuses when any core's worst placement is under the step, so the invariant no longer rests on the caller having run `step_gate.py`. New probe `--step-mhz refuses a core whose worst placement is under the step` (`test/PROBES_EXPECTED` line added); the existing `--step-mhz replaces ...` probe is its green control.
4. **Low.** ADR-0183 and its index row said the ECP5 provenance fields are "checked". `product_check.py --current` *can* compare them, but no caller passes `--current` for `ecp5_part`/`ecp5_target_mhz`, and nothing compares the `tools` block the Trellis digest sits in. The wording is corrected, and an "At integration" section records items 1-3.

5. **Rebase.** Main requires branches to be up to date (`strict: true`), and #345 merged ADR-0182 after this branch was cut, so I rebased onto main. The union merge driver on `docs/adr/README.md` doubled the 0181 and 0183 rows and put 0182 above 0181. The index is now main's file plus the one corrected 0183 row: rows 0180 to 0183 once each, in order, and `adr-numbering-test` passes on 178 ADRs. The engineer's commit that removed the earlier duplicate became empty and was dropped.

### Verified, no change needed

- **First dispatch against main's current stamp** (base 122ef7b, `dirty: yes`, hx8k-era `ram_words: 512`, no ECP5 pairs). I extracted the workflow's diff step verbatim and ran it against main's `product.json` plus a four-pair stamp written by this branch's `product_write.py`. The old schema reads cleanly; the plain diff prints; `--require-news` says **news**; the step writes `moved=true`. Re-diffing that fresh stamp against the same tree gives `moved=false` ("no news"), and an orphaned base gives news rather than a refusal.
- **`--step-mhz` numbers.** Fed the branch's recorded cycle factors, the code gives VexRiscv/littlecpu **1.1449** on Dhrystone at up5k (CLAUDE.md 1.14x), CoreMark VexRiscv 1.0146 and Hazard3 0.6511 (CLAUDE.md 0.986x and 1.54x inverted). The old Fmax product read 1.990x.
- **In `run_product.sh`'s path, `step_gate.py` gates before 12 is used.** `make compare-timing COMPARE_PART=up5k` runs `step_gate.py`, which exits nonzero under the step, so `sweep_clock` stops the whole run. A core under the step aborts the re-take; it is never credited 12 MHz. (The message says "failed to place", which misnames a step-gate failure. Cosmetic.)
- **Token test.** `test/compare_product_schedule_token_test.py` has six forced-red probes plus a control, each red for its own reason; `pin_bump_token_test.py` still passes.
- **CLAUDE.md** agrees with ADR-0183 and with itself: 9.39/10.75/10.80 DMIPS and 27.70/28.10/18.03 CoreMark at 12 MHz, and ECP5 25.06/47.42/43.99 DMIPS and 73.89/123.91/73.45 CoreMark, all reproduce from the recorded factors, as do the 1.14x, 1.15x, 1.54x, 1.89x, 1.76x, 1.68x and 1.01x ratios. It claims no fresh stamp exists.
- **ADR index**: no duplicate rows; 0183 is free (0182 is claimed by open #345). Comment density passes. No ticket IDs in the diff or commit messages. The one CI pipeline (`product_diff.py | tee`) runs under `set -euo pipefail`, and the graded `--require-news` call is not piped.

### Cannot be determined without running it on the pool

- Whether littlecpu clears 12 MHz at **all twelve** up5k seeds under the pool's yosys (0.69+24, not the 0.68+48 the branch measured with). CLAUDE.md's worst is 12.40 MHz, a 3% margin; one seed under 12 aborts the whole re-take by design.
- The actual wall time (estimated ~140 min, timeout now 300) and memory headroom for ECP5 placements in a 3 Gi pod.

### Follow-ups, not blocking

- **A PR opened with `GITHUB_TOKEN` triggers no workflows.** The stamp PR the dispatch opens will show every required check as "Expected", with no run behind it. A person has to kick CI on it (for example close and reopen it, or push to it with their own token). The pin-bump workflow has the same property.
- `board.json` in the Makefile has the same backslash-inside-quotes shape (off this path; it only works on macOS make).
- CLAUDE.md's one-line rule is scoped to scripts naming `SOC_ROM_CHPARAM`; the defect recurred in scripts that don't. Worth generalising to every multi-command `yosys -p` script.
- A leftover `git stash` entry from this branch (`stash@{0}: verify-committed-tree-jef935`, 2026-09-12 23:30 -0700). It holds pre-rebase copies of `Makefile`, `test/PROBES_EXPECTED` and `test/probe_gates.sh`, all subsumed by the branch. I left it in place because the stash stack is shared.

PR #346, opened 2026-09-13
Let nano run the portable subset of littlecpu's .S suite

thejefflarson:
## Summary

nano's simulation harness (#341) ran six hand-written nano programs rather than littlecpu's 75-program suite, because 73 of the 75 use x29 as scratch through the shared test macros (`test/asm/test_macros.h`, `test/asm/riscv_test.h`). nano is RV32E: any instruction naming x16-x31 traps, so nearly the whole suite trapped on nano before this change.

- **The shared macros now use x9, not x29**, as their comparison scratch register. x9 was picked by exhaustively checking every `TEST_CASE`/`TEST_CASE_D32`/`TEST_AMO` invocation across `test/asm` for a register that is never itself used as a destination (`testreg`) — the only such register in x0-x15. One direct collision was found and fixed: `test/asm/rvc.S` used `s1` (= x9) as a destination in four `RVC_TEST_CASE` calls, renamed to `a1`.
- **`test/macro_register_test.sh`** (new, joins `make test`) greps both shared headers for any register outside x0-x15 (by number or ABI alias), so a future edit that reaches for x16-x31 goes red immediately. Forced-red probes cover upstream's own x29 reappearing, an ABI alias (a7 = x17), and that the second header is scanned too.
- **`nano/asm/run_nano_tests.sh`** (the existing nano-suite runner) now also runs littlecpu's own `test/asm` suite directly, restricted to the portable subset: a `FLOOR` line whose second field is `EXCLUDED` names a program nano's core (RV32E, M, no CSR/traps layer, no A, no timer, no UART) cannot run at all, and the runner skips it rather than attempting it. Two new optional arguments (`lds-path`, `cycles`) let the same script serve both nano's own 6-program suite (unchanged call site) and this new 50-program run against `test/asm` (which has no `nano.lds` and needs a larger cycle budget than nano's own suite).
- **`nano/asm/LITTLECPU_FLOOR`** carries both the manifest and the floor/exclusion, the way `test/OBSERVED_FLOOR` already carries manifest + floor — one file, checked against `test/asm`'s real file list by the existing `test/check_suite_shape.sh`, rather than a second file that could drift out of sync. 24 of the 74 `.S` programs (plus the suite's one `.c` program) are `EXCLUDED` by name for needing a CSR (Zicsr/Zifencei included), the A extension, the machine timer or the UART.
[...]

PR #351, opened 2026-09-13
test/run_tests.sh: fail closed on a non-numeric retire count

thejefflarson:
## Problem

`test/run_tests.sh` reads the simulator's retire counts and compares them against `test/OBSERVED_FLOOR` using `[ "$retires" -lt "$floor" ]` inside `elif` conditions. If either operand is not a valid integer, `[` exits with status 2 ("integer expression expected"), bash's `elif` treats that as false, and `set -e` never fires inside a condition — so the run stays `PASS` with neither floor applied. A 20+ digit floor passes the old `^[0-9]+$` check but can overflow bash's 64-bit integers the same way.

Found during the security review of PR #346, which fixes the identical pattern in `nano/asm/run_nano_tests.sh`. #346 is still open; neither PR depends on the other.

## Fix

- Both the sim's observed counts and `OBSERVED_FLOOR`'s numeric fields are now validated against `^[0-9]{1,10}$` before any comparison; a count that fails this becomes `NO-COUNTS`, a floor line that fails it is rejected the same way the existing malformed-floor check already was (just with the digit cap added).
- The surviving comparisons are reversed to `! [ "$x" -ge "$floor" ]`, so a `[` parse error (exit 2) lands on the failing (`BELOW-FLOOR`) branch instead of silently reading as success.
- Mirrors the shape of #346's fix to `nano/asm/run_nano_tests.sh` (not yet merged), adapted for this runner's two-column floor (`retires`, `spec-checked`). Both runners use the same `^[0-9]{1,10}$` count check and the same fail-closed comparison once both land.

## Testing

- Added a `test/probe_gates.sh` probe with a non-numeric stub retire count and a non-numeric stub spec-checked count (`STUB_SIM_RETIRES=x STUB_SIM_SPEC=x`), required red with `NO-COUNTS`, plus its `test/PROBES_EXPECTED` line (LC_ALL=C sorted).
- `make probe-gates`: all 906 graded comparisons pass, including the new probe.
- `make test`: exit 0. littlecpu's `.S`/`.c` suite is unchanged: `75/75 passed`, `Failure list matches test/EXPECTED_FAIL exactly`. nano's suite: `6/6 passed`.
- `python3 test/comment_density_test.py .`: all 384 scanned files still at or under the 5% budget.

## Scope notes

- Did not touch `nano/asm/run_nano_tests.sh` (#346 fixes it) or the floor-file shape check in `test/check_suite_shape.sh` (unrelated — it only validates program names, not counts).
- `test/OBSERVED_FLOOR`'s existing entries are all well under 10 digits, so the new cap changes no currently-passing behavior.

Closes JEF-1023.


thejefflarson:
## Architect review: READY, pending CI on f55b6dc

**The fix is correct.** Under the old code, `[ x -lt 10 ]` exits 2, `elif` reads that as false, and the run stays PASS. I measured this on bash 5.3: `[ 99999999999999999999 -ge 1 ]` exits 2, so the reversed `! [ x -ge floor ]` is what makes an unparseable operand fail closed. The two 10-digit caps (`^[0-9]{1,10}$`, at most 10 digits) sit in front of that comparison and turn it into a named status: NO-COUNTS for a count, or the malformed-floor error for a floor. ADR-0081's `.c` floor cap of 64 is unchanged at `test/run_tests.sh:52-54`, and its probe ("a C floor copied out of the measured table is rejected") still runs.

**Pushed by the architect (f55b6dc, plus a reworded a81f9dd):**

1. `test/probe_gates.sh:535` — the one probe set both stub counts to `x` at once, so deleting either half of the NO-COUNTS test left it red, and neither half had a red direction of its own. I split it into one probe per count.
2. `test/run_tests.sh:42,197,198` — the three 10-digit caps had no red direction: reverting `{1,10}` to `+` turned no probe red. I added one probe each for a 20-digit floor, retire count and spec-checked count.
3. Commit a81f9dd — the message said the pattern "mirrors the same fix already shipped" in #346. #346 is still open. This repo squashes with `COMMIT_MESSAGES`, so that sentence would have landed on main. I reworded it. The PR body is corrected the same way.

**Mutation evidence.** Each run reverted only the named arm and ran the full `make probe-gates`:

| Mutation | Probes red | Probes ok |
|---|---|---|
| floor cap and retire-count cap to `+`, spec half of NO-COUNTS deleted | exactly: floor too long, retire count too long, non-numeric spec-checked | 906 |
| retires half of NO-COUNTS deleted | exactly: non-numeric retire count, retire count too long | 907 |
| spec-checked cap to `+` | exactly: spec-checked count too long | 909 |

The reversed comparisons have no probe of their own: both caps sit in front of them, so no non-integer can reach them. That is defense in depth, and the reason is recorded in f55b6dc's message.

**Checks:**
- All five probes sit above `actual_labels=`.
- The body of `test/PROBES_EXPECTED` passes `LC_ALL=C sort -c`.
- `test/comment_density_test.py`: 384 files, all at or under 5%. The PR adds no comments.
- No ticket IDs in the diff or the commit messages.
- CI's awk supports the `{1,10}` interval: #346's `test` job passed on the same runner pool with the same awk regex on `make test`'s path.

**Local run on f55b6dc:** `make test` exits 0. That is 910 graded comparisons, 75/75 with the failure list matching `test/EXPECTED_FAIL`, and nano 6/6.

**Relation to #346.** Neither PR depends on the other; #351 touches only littlecpu's runner. Once both land, the two runners read the same way: the same `grep -qE '^[0-9]{1,10}$'` count check, the same 10-digit floor cap in awk, and an equivalent fail-closed comparison (`! [ x -ge f ]` here, `if [ x -ge f ]; then :; else BELOW-FLOOR` there). **The second PR to merge will hit a one-line textual conflict in `test/PROBES_EXPECTED`**, because both insert directly after "a non-numeric floor is named rather than compared arithmetically". To resolve it, keep both lines. In `LC_ALL=C` order, #351's "...retire count is NO-COUNTS, not a silent PASS" and "...spec-checked count is NO-COUNTS too" sort before #346's "...retire count is graded NO-COUNTS..." (`N` sorts before `g`). The labels differ, so the probe multiset has no collision.

**Merge state.** GitHub reports MERGEABLE but BEHIND main by b2739c6 (#344). The merge with main is clean (`git merge-tree`). If branch protection requires an up-to-date branch, update the branch before merging.

PR #352, opened 2026-09-13
nano formal: delete the dead SYSTEM exclusion, gate both memchecks on mem_instr

thejefflarson:
## What

Two small fixes to nano's formal harness, both cheap:

1. **Delete the ungraded SYSTEM exclusion.** `nano/formal/complete.sv` skipped
   its completeness assertion for any SYSTEM-opcode retire, with no comment,
   manifest entry, or anything grading the constant `7'b1110011` it matched
   on. nano has no CSR/trap layer yet, so every SYSTEM instruction already
   traps and the `!rvfi_trap` guard on the line above excuses it before the
   opcode test is reached — the guard was dead. Deleted it outright and
   re-ran `make -C nano/formal complete`: it still PASSES, which is the
   measurement that says the exclusion was excusing nothing on this core.

   **A later ticket that adds CSRs reopens this.** Once SYSTEM instructions
   can retire non-trapping, whoever lands that must either keep `complete`
   green as-is or declare the exclusion properly — an `// EXCLUDE` marker, a
   `nano/formal/COMPLETE_EXCLUSIONS` baseline, and a `complete-exclusions`
   prerequisite running a `check-complete-exclusions.py`, the way the main
   core's `formal/COMPLETE_EXCLUSIONS` does.

2. **Gate both memchecks' assumes on `mem_instr`.** `dmemcheck.sv`'s memory-
   shadow assume ran on every bus transaction, including instruction
   fetches, so whenever the solver-chosen watched data address happened to
   equal a fetch address, the assume forced the fetched word to equal the
   data shadow — pruning any trace where the core fetches from an address
   it also loads from and the two differ. `imemcheck.sv`'s mirror bug: its
   rand_const-halfword assume ran on data accesses too, so it also bound
   `mem_rdata` when the core loaded from the watched instruction address.

   Added `&& !mem_instr` to `dmemcheck.sv`'s assume and `&& mem_instr` to
   `imemcheck.sv`'s. Both changes only ADMIT traces the old, over-broad
   assume used to prune — neither narrows an assertion — so neither can
   introduce a false failure, and both checks still PASS after the change.

   Documented `imemcheck.sv`'s one remaining, pre-existing limitation in a
   two-line comment rather than porting the main core's `shadow_stored`
   gating: nano's `imemcheck.sv` wires the vanilla
   `riscv-formal/checks/rvfi_imem_check.sv` checker directly, whose internal
   `assert` is unconditional (unlike the main core's fully custom, hand-
   written checker, which owns its assert and can gate it). Porting
   `shadow_stored` here would mean forking that checker file too — the same
   scale of change as `nano/formal/rvfi_insn_check.sv` already is for a
   different reason — which is a bigger change than "gate the assume", the
   ticket's own proposal, and not needed to satisfy the acceptance criteria
   (comment OR port).

## Why

`docs/adr/` and `CLAUDE.md`'s four-habits rule: "a grader that cannot fail is
not a grader." Both memcheck weakenings were verified to still catch a real
defect (see Testing below), not just re-run for PASS.

## Testing

- `make -C nano/formal complete` — PASS with the guard deleted (measures the
  exclusion was dead).
- `make -C nano/formal complete_cover` (and its `complete-cover-probe`
  prerequisite) — PASS; the anti-vacuity control (a stalled-bus mutant) still
  makes every cover goal unreachable, unaffected by the `complete.sv` edit.
- `make -C nano/formal dmemcheck` — PASS with `!mem_instr` added.
- `make -C nano/formal imemcheck` — PASS with `mem_instr` added.
- `make -C nano/formal check ill_e ill_e_cover check-rvfi-insn-check` — all
  PASS / matching baseline, unaffected by this change.
- **Manual soundness check** (not committed, ad hoc): built a copy of
  `nano.v` with `rvfi_mem_rdata <= mem_rdata;` mutated to `<= ~mem_rdata;`
  and ran `dmemcheck` against it — FAILS, as expected. Same for
  `rvfi_insn <= instr;` mutated to `<= ~instr;` against `imemcheck` — FAILS.
  Confirms neither weakened assume turned its check into one that cannot
  fail.
- `make test` — green: 75/75 `.S`/`.c` programs pass, `nano-test` and
  `nano-startup-test` match their baselines, `probe-gates` (25 comparisons)
  all green, `comment-density-test` passes (384 files, all ≤5%).

## Scope / risk

- No RTL changed (`nano/nano.v` untouched) — this PR only touches
  `nano/formal/{complete,dmemcheck,imemcheck}.sv`.
- Neither memcheck gained a new forced-red probe in `test/probe_gates.sh`:
  the main core's own `formal/imemcheck.sv`/`dmemcheck.sv` targets have none
  either, and the ticket's own verification bar for this class of change was
  "re-run and confirm PASS" plus (per a repo lesson) a deliberately-broken
  core showing FAIL, both done manually above.
- DECISION NEEDED (flagged, not blocking): none — the ticket's own two
  options (delete-and-remeasure, or comment-vs-port) both resolved cleanly
  against the current tree.

Closes JEF-991.


thejefflarson:
## Architect review — READY (not armed; the main loop owns the merge slot)

I read the diff, re-ran the gates, and pushed one small follow-up commit (below). Nothing blocks.

### Findings, severity-ordered (`file:line — problem — fix`)

1. **Low, fixed on the branch.** `nano/formal/complete.sv:64` — deleting the SYSTEM guard creates a dependency on the upcoming CSR/trap layer, and the only record of it was this PR's description and commit message. Once `csrrw` or `mret` retires without trapping, `complete` goes red: the pinned riscv-formal has no SYSTEM spec model, so `spec_valid` is low for every SYSTEM encoding. The risk is that whoever lands that layer restores the bare opcode guard to get green, which is exactly the ungraded exclusion this PR removed. **Fix pushed:** a two-sentence tripwire comment above the assertion pointing at the main core's graded `EXCLUDE` shape, plus a dated amendment to ADR-0168 (the ADR that owns nano's `complete.sv`) recording the dependency and its required resolution.
2. **Informational, the PR text only.** The Testing section's "probe-gates (25 comparisons)" is wrong. `make probe-gates` on this branch reports **905 graded comparisons, every failure path executed**, exit 0, both before and after my commit. The 25 is almost certainly `test/mutation_probe.sh`'s own summary line ("N comparisons, each forced red for its own reason…"), which `make test` also prints. No subset was run; the label was misattributed.
3. **Informational, follow-up and not a hold.** No committed forced-red direction exists for nano's `dmemcheck`/`imemcheck`. The same is true of littlecpu's pair: `formal/Makefile` gives neither a probe prerequisite, `test/MUTATION_DETECTORS` names neither, and `test/probe_gates.sh` mentions them only as `HARNESS` entries in the interrupt-tie-off fixture. That confirms the engineer's claim.

### Decision on the red direction (recorded in the ADR-0168 amendment)

A committed probe is **not owed by this change**. Both edits add a conjunct to an assume's guard: `!mem_instr` in dmemcheck, `mem_instr` in imemcheck. The new condition implies the old one, so each assume constrains a subset of the cycles it used to, and the traces the solver may explore are a superset of the old set. Any counterexample the old harness could find, the new one can still find. A change that only removes environment constraints cannot remove a red direction. It can only add false reds, and both checks still pass. The hand demonstration (inverting `rvfi_mem_rdata` and `rvfi_insn` in a scratch `nano.v`) is good evidence on the day, but it is not a gate. The standing gap, no committed red direction for any of the four memchecks across both cores, is older than this PR. It should be its own ticket, shaped like `nano/formal/complete-cover-probe.py`.

### Verified independently

- The gates are correct for nano's bus. `nano.v` sets `mem_instr <= 1` only in `fetch_instr` and `<= 0` when it issues a load or store, so the split falls exactly between fetch and data. dmemcheck's shadow update stays ungated, and that is harmless because a fetch drives `mem_wstrb` to zero.
- The imemcheck comment is accurate: `imem_data` is a trace-constant `rand_const` inside the unforked `rvfi_imem_check.sv`, so a store into text goes unmodelled. Documenting that instead of forking the checker is the right call.
- `nano/formal/checks.cfg` is untouched; line 1 is still `#insn-check rvfi_insn_check.sv`.
- Comment density: all 384 files at or under 5%. `complete.sv` is 2 comment lines of 88 and `imemcheck.sv` 2 of 64.
- There are no ticket IDs in code, comments, the ADR, or either commit message.
- After my commit: `make -C nano/formal complete` PASS, `complete_cover` PASS with its forced-red `complete-cover-probe` prerequisite, `make probe-gates` 905/905, and `adr-numbering-test`, `retired-term-test` and `band-source-test` PASS.
- The branch is one commit behind main (#344, which touches no nano file) and GitHub reports it MERGEABLE.

PR #353, opened 2026-09-13
Give nano.v an iverilog leg: it has no four-state simulation

thejefflarson:
## Summary

`nano/nano.v` didn't elaborate under iverilog: decode flags and the register array were
used before they were declared, and its RVFI outputs are typed `wire` by the shared
`RVFI_OUTPUTS` macro while `nano.v` drives them from an `always_ff` block. This gives
nano's simulation harness a second, four-state elaboration frontend — the one thing
cxxrtl structurally cannot provide — and the chip-top work needs it to check pad
tri-states.

**No functional change to `nano.v`**: every bare `logic` declaration moves ahead of the
assigns that read it (nothing about what drives what changes), and the RVFI outputs are
now driven through a local `` `RVFI_SHADOW(name) `` macro that declares a `logic` shadow
register per output (sized by `$bits(name)`, so it tracks `` `RVFI_OUTPUTS `` rather than
restating its arithmetic) and continuously assigns each out to its still-macro-declared
port. `` `RVFI_OUTPUTS `` itself is untouched.

- `make nano-area` (built with no `RISCV_FORMAL`, so it covers the declaration reorder
  alone): 61411.4 of 61412.0 µm² both before and after, byte-identical `yosys stat`.
- `make -C nano/formal check` (the real evidence for the `RVFI_SHADOW` block, since it's
  the only build that compiles it): 82 checks, 80 pass, the same 2 known-fail entries.

## What's new

- `nano/tb/nano_icarus.vvp`, `nano/tb/nano_testbench.v`'s general ICARUS runner,
  `nano/tb/nano_sim_icarus.sh` (wraps it behind `nano-sim`'s own CLI/exit codes), and
  `nano/tb/nano_dual_leg_test.sh` (runs both legs' suites **in parallel**, grades each
  against `nano/asm/EXPECTED_FAIL`/`OBSERVED_FLOOR`, then diffs their per-program result
  tables and requires an exact match). `make nano-test` now runs this; all six programs
  agree, digit for digit, on both legs.
- `nano/tb/nano_x_probe.sh` forces the iverilog leg's X-detection both ways, on **both**
  the load path (an unzeroed word) and the store path (`nano.v` never resets
  `regs[1]`-`regs[15]`, so storing one unwritten is a real X on the shipping harness — no
  mutation needed to prove that path). Not hermetic; wired as a prerequisite of
  `make nano-test`, the same standing `nano/formal/ill-e-probe.py` has for
  `make -C nano/formal ill_e`.
- 27 hermetic probes in `test/probe_gates.sh` covering all of the above against stubbed
  tools; `test/PROBES_EXPECTED` updated to match.
- ADR-0185 records the change (renumbered twice as other in-flight PRs claimed 0183 and
  0184 first).

## Revision history on this PR

**Round 1** (initial): the core RTL fix + dual-leg harness + X-detection probe, as
described above.

**Round 2**, after `/soundcheck` review found 7 Low-severity issues in this PR's own new
code and a correctness note on the ADR:
1. `nano_dual_leg_test.sh`: fixed a status-comparison bug (`-ne 0` read an empty status
   file as success; now a string compare) and a `set -e` interaction where a no-match
   `grep` in `extract_table` aborted the script before the "could not parse" guard could
   ever run (now wrapped in `{ ... || true; }`). Added a probe for the no-match case.
2. **Widened the X-detection check**: now covers `pc_wdata`, `rs1`/`rs2_rdata`,
   `mem_addr`, and `mem_wdata` masked by the write strobe (previously just
   `insn`/`pc_rdata`/`rd_wdata`), and validity itself is read as `1'b1` **or** `1'bx`
   rather than a plain truthy `&&` (so an unknown validity signal doesn't silently skip
   the check). Confirmed `make nano-test`'s six programs stay clean under the wider
   check — none of them read a register before writing it. Added a second probe pair
   (`store_clean.S`/`store_x.S`) proving the store-path term specifically.
3. Added 8 probes exercising `nano_sim_icarus.sh`'s own verdict-to-exit mapping directly
   (PASS/FAIL/TIMEOUT/monitor-error/trap/X-reached/zero-retires/unrecognized), via a stub
   `vvp` that plays back a fixed transcript per case.
4. Gave an X-reaching-a-retire its own exit code (**7**), distinct from a real monitor
   error's 4 — previously it collapsed into a bare `MONITOR-ERROR` with the diagnostic
   detail lost once the run log's temp dir was cleaned up. `run_nano_tests.sh`'s `case`
   gained a matching `X-REACHED` arm.
5. Two GNU-only grep patterns (`\|` alternation, `\S`) replaced with portable
   `-E 'a|b'` / `[^[:space:]]` — BSD grep reads `\|`/`\S` literally.
6. Fixed an off-by-one in the iverilog leg's cycle budget: `nano_cxxrtl.cc` spends its
   first `--cycles` edge in reset, but the ICARUS loop was taking the reset edge *outside*
   the counted loop and then running the full budget again, giving it one extra
   post-reset edge for the same `--cycles`/`+CYCLES=` value. Now the reset-deassert edge
   is folded into cycle 0 of the counted loop, matching `nano_cxxrtl.cc`'s own timing.
7. Corrected the ADR: `make nano-area` never compiles the `` `ifdef RISCV_FORMAL `` block
   at all (no macro defined), so it's evidence for the declaration reorder only, not the
   RVFI_SHADOW change — dropped a wrong citation of littlecpu's own component proofs and
   named the real evidence (the nano formal run + the dual-leg agreement) instead. Also
   corrected a factual claim that nano "never" implements `mcycle`/`minstret`: ADR-0177
   keeps those counters as a feature, and a later ticket wires them up as real CSRs,
   which is exactly why the shadow-register/macro-driven port list was chosen over
   hand-declaring ports the way `rtl/littlecpu.v` does — it needs no changes when that
   lands, only eight more `` `RVFI_SHADOW(...) `` lines in that later ticket.

All fixes re-verified: `make test`, `make nano-test`, `make probe-gates`,
`make -C nano/formal check`, and `make nano-area` all green after each round, rebased
cleanly onto `main` as it moved under this PR (ADR-0183 → 0184 → landed; renumbered to
0185 to stay clear).

## Decisions made (no open questions)

- **RVFI outputs stay macro-driven** rather than hand-declared like `rtl/littlecpu.v`
  does — see point 7 above.
- **`nano-test` itself changed shape** (now dual-leg) rather than adding a parallel
  `nano-test-icarus` target, since grading both legs against one shared baseline plus an
  explicit diff is a strictly stronger property than two independent baseline passes.
- **`nano-dhrystone`/`nano-coremark` are untouched** — out of scope; the iverilog runner
  only implements what `nano-test` needs.

Closes JEF-1017.


thejefflarson:
## Architect review: READY (not armed; the main loop arms)

Reviewed head 66439ee, plus one docs-only commit I pushed, **46acba2**, which touches only `docs/adr/0185-nanov-gets-an-iverilog-leg.md` and its `docs/adr/README.md` row.

### No functional change: proven, not sampled

I ran a yosys equivalence check of `nano/nano.v` at origin/main (e212bb4) against this branch: `proc; memory_map; opt_clean; equiv_make; equiv_struct; equiv_induct -seq 1; equiv_status -assert`.

| Build | Result |
|---|---|
| `NANO_RISCV_FORMAL_MACROS` defined (compiles `RVFI_SHADOW`) | 1,673 of 1,673 `$equiv` cells proven |
| no defines (the `nano-area` build) | 1,289 of 1,289 proven |
| red control: gate with `rvfi_trap_q <= 0` instead of `<= trap` | 1 unproven (`rvfi_trap`), `equiv_status -assert` fails |

The proof covers reset and time 0: neither file gives an RVFI register an initial value, and neither adds or removes a reset arm, so both designs start from the same free state. The formal run and the two-leg agreement were consistent with this but only sampled it. This check covers every state.

1. **Medium, fixed in 46acba2.** In `docs/adr/0185-nanov-gets-an-iverilog-leg.md`, the X-check paragraph called a program that reads a register before writing it a *false positive* the six programs happen to avoid, while `store_x.S` uses exactly that event as its red direction. It now states the event as the check's intent. Nano never resets x1-x15, and cxxrtl reads them as zero, so a verdict can depend on state the hardware does not define, and only the iverilog leg can report that. The first C program whose prologue spills an unwritten callee-saved register will trip it. The remedy then is a startup file that writes x1-x15, not a narrower check.
2. **Medium, fixed in 46acba2.** In the same ADR, "The RVFI ports" said the mcycle/minstret ticket needs only eight more `RVFI_SHADOW` lines. `RVFI_OUTPUTS` declares `rvfi_csr_mcycle_*` and `rvfi_csr_minstret_*` only under `RISCV_FORMAL_CSR_MCYCLE`/`_MINSTRET` (see `rvfi_macros.vh`). genchecks defines those macros in the generated sby files, and `NANO_RISCV_FORMAL_MACROS` does not. Unguarded, `$bits(rvfi_csr_mcycle_rdata)` names a port that does not exist, and both sim builds stop elaborating. The ADR now says each group of four goes under its own guard and that each port is 64 bits wide.
3. **Low, pre-existing, not a hold.** At `nano/nano.v:259`, `rs2_valid` is true for `ebreak`, whose rs2 field is 1, so nano reports x1 as `rvfi_rs2_addr`/`rvfi_rs2_rdata` on an `ebreak`. riscv-formal has no spec model for SYSTEM, so no check sees it. On the iverilog leg, an `ebreak` with x1 unwritten reports X-REACHED (exit 7) before the trap (exit 5). Fix: drop `is_ecall`/`is_ebreak` from `rs2_valid`, in the CSR/traps ticket.
4. **Low, not a hold.** `nano/tb/nano_testbench.v:194-195` and `nano/tb/nano_sim_icarus.sh:17`: every iverilog suite run dumps the whole testbench to `./nano_testbench.vcd` in the repo root. The file is gitignored and overwritten per program. Meanwhile the wrapper accepts `--vcd` and throws it away. Fix: gate `$dumpfile`/`$dumpvars` on a `+VCD=` plusarg, and have the wrapper pass `--vcd` through.

### Other checks

- **Cost to `make test`**, on one Apple-silicon Mac against origin/main at e212bb4:

  | Target | main | this branch |
  |---|---|---|
  | `make nano-test`, warm | 0.55 s | 1.86 s |
  | `make nano-test`, cold | 19.6 s | 18.6 s |
  | `make probe-gates` | 959 comparisons in 2:45 | 980 in 2:40 |

  The cold figure is dominated by the cxxrtl compile, and both the cold and probe-gates differences are run-to-run noise. The real cost is about +1.3 s per run, because the X probe's two iverilog builds and four runs repeat on every call. This is now recorded in the ADR.
- **ADR index**: 179 ADR files, each with exactly one README row, and no duplicate number. 0185 is absent from origin/main; 0184 stays with #350.
- **Comment density**: all 397 files are at or under 5%. The diff carries no ticket IDs, and the code carries no ADR or invariant numbers.
- **Re-run after my commit**: `adr-numbering-test`, `retired-term-test`, `band-source-test` and the comment-density test are all green.
- **CI**: at 66439ee, all 33 checks passed, with `auto-merge` skipped. It is re-running on 46acba2.

PR #354, opened 2026-09-14
Fix compare-product-schedule's publish step and exercise it on make test's path

thejefflarson:
## What

The publish step of `.github/workflows/compare-product-schedule.yml` ran
`git commit -m "Refresh the cross-core product stamp" -F /tmp/product-diff.md`,
which git refuses (`fatal: options '-m' and '-F' cannot be used together`).
Nothing on `make test`'s path had ever executed this step, and every earlier
run of the workflow had died before reaching it, so the first run that
completed the whole 58-minute measurement lost the result at the last step.

- **Fix**: build the commit message from one file (title, blank line, the
  measured diff) and pass `-F` alone.
- **Test**: `test/compare_product_schedule_publish_test.py` extracts the
  publish step's `run:` script out of the workflow YAML (the same
  step-chunking `test/compare_product_schedule_token_test.py` already uses)
  and executes it as a real shell script against a throwaway git repository,
  with `gh` and `git push` stubbed on `PATH`. It checks: the branch name
  (`compare-product/refresh-YYYYMMDD`), the `git config` identity, the
  commit's subject and body, `gh pr create`'s `--title`/`--head`/
  `--body-file`, and the `GITHUB_STEP_SUMMARY` write. Wired into `make test`,
  with a forced-red probe in `test/probe_gates.sh` (the old `-m ... -F` line,
  restored, required to fail for the reason it actually failed on the pool).
- **Artifact upload**: added `actions/upload-artifact` (pinned by SHA, same
  convention as `ci.yml`) right after `soc/compare/product.json` and the diff
  are written, `if: always()`, `if-no-files-found: warn` (best-effort -- a
  job that failed before `make compare-product` ran leaves nothing to
  upload). So a failure in any step after that point no longer throws the
  hour of measurement away with the runner.

## What else this touched

`Makefile` and `CLAUDE.md`'s `make test` enumeration gained the new
`compare-product-schedule-publish-test` target. `test/PROBES_EXPECTED`
gained the four new probe labels in `LC_ALL=C` order (verified: the body
sorts identically before and after).

## What's still unexercised end-to-end

The ticket asked which steps have never run for real. From the one
completed run (34777290727): "Diff against the committed stamp" ran fine
on the pool. "Upload the freshly measured stamp" is brand new and hasn't
run on the pool yet. "No move -- say so and stop" has never run on the
pool either (that run found news, so the `moved == 'true'` branch is the
only one that's executed there), and this PR doesn't add a test for it --
it's a much smaller diagnostic step with no external side effect to stub,
unlike the publish step this ticket is about.

## Alternative considered and not taken

A `/simplify` pass surfaced that this workflow's ~30-line publish script
lives inline in the YAML, the same shape of problem
`.github/workflows/riscv-formal-pin-bump.yml` solves by delegating to a
tracked script (`formal/bump-riscv-formal-pin.sh`) that can be tested by
direct invocation, with no YAML re-parsing. That's a more general fix, but
it's a bigger structural change than this ticket's acceptance criteria ask
for (which specify extracting and executing the `run:` block, matching
`test/compare_product_schedule_token_test.py`'s existing approach).
Flagging here as a **DECISION NEEDED** for a follow-up: extract
`compare-product-schedule.yml`'s publish logic into a standalone script the
way the pin-bump workflow already does, which would let this new test
invoke it directly and delete the YAML-extraction machinery.

## Testing

- `make test` (full suite, including `probe-gates`): green, twice, before
  and after a `/simplify` pass. `963 graded comparisons, every failure path
  executed.` includes the new group's 4 probes (control, missing-workflow,
  renamed-step, and the forced-red old `-m`/`-F` line reproducing the
  actual `fatal: options '-m' and '-F' cannot be used together`).
- `/soundcheck:pr-review`: no Critical/High findings.
- `/simplify`: applied -- dropped an unnecessary `if/else` in the new
  `probe_gates.sh` fixture (confirmed `sed ''` is a no-op passthrough) and
  a `real()` git helper now returns `stdout` so four post-run git
  inspections reuse it instead of hand-rolling `subprocess.run` each time.
  Not applied: the triplicated `steps()` YAML-chunker (each
  `*_token_test.py`/`*_publish_test.py` file is deliberately
  self-contained, since `probe_gates.sh`'s fixtures copy exactly one script
  into an isolated directory) and merging the near-duplicate
  `cpst_fixture`/`cpsp_fixture` helpers (matches this file's existing local
  convention of one fixture helper per probe group, e.g. `pbt_fixture` next
  to `cpst_fixture`).

Closes JEF-1031


thejefflarson:
## Architect review: READY

The fix is correct: the step now builds one message file and calls `git commit -F` alone. The failed run's log (34777290727) shows `gh auth setup-git`, `git config` and `git checkout -b` all succeeding on the pool, then git refusing `-m` with `-F` and exiting 128. The new test reproduces that refusal from the old line.

### What I pushed to this branch

Two commits on top of 94eaa36:

- **6a43b22**: two changes.
  - **Hermetic temp files.** The test wrote and deleted the shared `/tmp/product-diff.md`, `/tmp/commit-message.md` and `/tmp/pr-body.md`. Two worktrees running `make test` at once could delete each other's files, and a developer's own copies were removed. The test now rewrites the step's `/tmp/` prefix to its own temporary directory.
  - **Per-run branch name.** The refresh branch was `compare-product/refresh-YYYYMMDD`. A second dispatch on the same UTC day, after a first had pushed, would be refused at `git push` and lose the re-take again. The name now ends in `-$GITHUB_RUN_ID`, read from the environment rather than written as a `${{ }}` expression.
- **The security audit's two Low findings**:
  - **Git isolation.** Both subprocess calls now share one environment: every `GIT_*` key dropped, `GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_NOSYSTEM=1`, and `HOME` set to the temporary directory. I exported `GIT_DIR`, `GIT_WORK_TREE` and `GIT_INDEX_FILE` at a scratch repository and ran both versions. The previous test rewrote that repository's `.git/config`; the new one leaves it byte-identical.
  - **The push target is graded.** The single push must be exactly `push --quiet origin <PR head>`. `gh pr create` must receive `--base main`. The commit must touch only `soc/compare/product.json`; the fixture now carries a stray tracked edit, so a `git add -A` has something to sweep in.
  - **Four new probes**, each red for its own reason: a push to `"$branch:main"`, `--base develop`, `git add -A`, and the branch without the run-id suffix. They sit in the same group above `actual_labels=`, and their `PROBES_EXPECTED` lines are inserted in C order.

`make test` on 9ed990c, the pushed head, is green: 967 graded comparisons, 4 more than the engineer's 963. All eight probes in the publish group read `ok`, and the suite is 75/75, matching `test/EXPECTED_FAIL` exactly.

### 1. What the stub test proves, and what it cannot

It **proves**:
- The bytes of the step's `run:` block execute under `bash` with the step's own `set -euo pipefail`.
- Real git `config`, `checkout -b`, `add` and `commit` work against a scratch repository.
- The commit's subject and body are right, the push goes exactly to `origin <branch>`, and the PR's head, base, title and body file are right.
- The step summary is written.
- `date -u +%Y%m%d` and `+%Y-%m-%d` are POSIX formats, identical on GNU and BSD `date`.
- I also checked that `git commit -F` keeps the `## <pair>` headings `product_diff.py` emits. With no editor, git only trims whitespace; it strips `#` lines only when an editor opens.

It **cannot prove**:
- **The real `git push`** through `gh`'s credential helper with the Actions token. `gh auth setup-git` itself is proven on the pool by the failed run.
- **The real `gh pr create` call.** The permissions say it should work: the job has `pull-requests: write`, and the repository reports `can_approve_pull_request_reviews: true`, which is the "Allow GitHub Actions to create and approve pull requests" setting.
- **`product_diff.py`'s real output** in the message. The fixture diff is one line.
- **YAML semantics.** The extractor is deliberately YAML-naive: a `${{ }}` expression added to the block would reach bash unsubstituted. It would go red, but for the wrong reason.
- **Whether the PR it opens can merge. This is the important one, and it predates this PR.** ADR-0069 measured that GitHub fires neither `pull_request` nor `push` for a PR opened by the Actions token, so the PR gets zero check runs. ADR-0069 says it outright: "`gh workflow run` is not a way to satisfy branch protection, and neither is a bot-opened PR." The stamp PR this step opens will show its required checks as expected and never reporting. See the second decision below.

### 2. Steps that have never run end to end

- **"No move -- say so and stop".** It was skipped in both runs: 34111542512 died in `make compare-product`, and 34777290727 had `moved=true`. It only echoes, `cat`s and `date`s into the step summary and holds no credential. If it fails, a no-news run goes red and loses nothing, because the upload has already run. **Acceptable.**
- **The upload step** is new, so it has never run. Two cosmetic points, neither blocking:
  - It is `if: always()`, so after a failed `make compare-product` it uploads the checked-out old `product.json` under an artifact named for the fresh stamp.
  - It mixes `/tmp/...` with a workspace path, and upload-artifact roots the artifact at their common ancestor, `/`. The files therefore land nested under `tmp/` and the workspace path.
  **Acceptable.**

### 3. The upload step's pin and retention

`actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1` matches `ci.yml` lines 645 and 761 byte for byte. Retention is 14 days, against `ci.yml`'s 1 day for intermediate artifacts. For a weekly workflow, 14 days keeps the artifact alive past the next run. **Sensible.**

### 4. Housekeeping

- `test/comment_density_test.py` passes: 392 files, all at or under 5%.
- `test/probes_header_test.py` passes.
- The `PROBES_EXPECTED` body passes `LC_ALL=C sort -c`.
- The label "a missing schedule workflow file is an error, not an empty pass" appears twice on purpose. `docs/manifests/probes-expected.md` allows exactly this: the manifest is compared as a multiset, and the same message is checked against a different script.
- No ticket IDs appear in any file or commit message.

### Decisions recorded

**Script extraction (DECISION NEEDED): yes, the publish logic moves into a tracked script, in a follow-up and not in this PR.** The pin-bump workflow already follows this pattern, per ADR-0176: tracked `decide`/`regenerate`/`publish` scripts, with the credential set per step.
- **Why not here:** this PR unblocks the re-dispatch, and its extractor already executes exactly what the runner executes, which is what this failure needed.
- **What extraction removes:** the YAML-naive extractor, and the third copy of the step-chunker (`test/pin_bump_token_test.py`, `test/compare_product_schedule_token_test.py`, and this test).
- **What it keeps:** `compare_product_schedule_token_test.py` still confines `GH_TOKEN`, because the token stays on the step that calls the script.

**Who opens the stamp PR: follow ADR-0069, in the same follow-up.** ADR-0069 is the repo's recorded decision on this exact mechanism: a bot-opened PR cannot satisfy branch protection, so a bot proposes and a human opens the PR. The follow-up should switch this step to ADR-0069's shape: push the branch, then open an issue with a compare link. It should not reach for a PAT, which ADR-0069 rejected because the repository is public. Until then, the stamp PR from the next dispatch needs one action from a human account, closing and reopening it or pushing a commit to it, before its checks run. The reopen route is unmeasured here; what ADR-0069 measured is that a PR opened under a human account gets its checks.

PR #357, opened 2026-09-14
nano's QSPI timing model: the loop buffer is the lever, FIFO depth alone is not

thejefflarson:
Closes JEF-1030.

## What this does

Adds `nano/tb/nano_qspi_memory.v`, a behavioural timing model of nano's QSPI flash (code) and
PSRAM (data), standing in for `nano_memory.v`'s zero-wait model on the same bus. Selected by
a `NANO_QSPI_TIMING` build macro and a new `nano-qspi-sim`/`nano-qspi-timing` Makefile target
(`NANO_QSPI_PREFETCH_DEPTH`, `NANO_QSPI_LOOP_WINDOW`, `NANO_QSPI_PREAMBLE_CYCLES` parameterize
it). `nano.v` is not touched. `nano/bench/run_qspi_timing.sh` sweeps eight configurations
(zero-wait control, no-overlap, FIFO depth 2/4, FIFO+loop-buffer 8/16, and a QPI-preamble
variant) against Dhrystone and CoreMark, reported by `nano/bench/qspi_timing_report.py`.

Full derivation, modelling assumptions, the bug the model surfaced (and fixed, in the model
only), and the measured table are in `docs/adr/0186-...md`.

## Headline result

FIFO depth alone barely moves the needle (depth 2 -> 4: +1.4% CoreMark/MHz). The loop buffer
is the real lever the brief predicted: 8 parcels buys +14.6%/+22.1% (Dhrystone/CoreMark)
DMIPS-or-CoreMark/MHz over the depth-2 baseline, 16 parcels a smaller further increment. QPI's
shorter re-entry preamble adds ~3.5% more on top of the best FIFO+loop configuration. The
brief's ~2 MIPS estimate is reached only once a loop buffer is present; the measured
no-overlap figure alone is 1.34-1.69 MIPS at an assumed 64 MHz.

## Testing

- `make test` — full suite, green (littlecpu's 75/75 `.S`/`.c` programs, nano's 6/6, the new
  cxxrtl/iverilog dual-leg comparison, `probe-gates`, every repo-scanning `*-test` target).
- `make nano-qspi-timing` run end to end for all eight configurations; both benchmarks pass
  their own self-check in every one, and the accounting identity (every cycle in exactly one
  of four buckets) holds in every run.
- The zero-wait control reproduces ADR-0182's own figures to the cycle (2,526.2 cycles/run,
  0.225 DMIPS/MHz; 1,848,480.0 cycles/iteration, 0.541 CoreMark/MHz), confirming the default
  `nano-sim`/`nano-test`/`nano-dhrystone`/`nano-coremark` path is byte-for-byte unchanged.
- New probes in `test/probe_gates.sh` (`nano/bench/qspi_timing_report.py` group) force the
  accounting-identity check red both ways, with `fixture_anchor` calls tying the hand-typed
  log fixture to the real printf format so it goes stale if that format ever changes.
- `/soundcheck:pr-review`-equivalent self-review: no Critical/High findings — this is pure
  simulation tooling with no untrusted-input surface (CLI args and log files this same
  toolchain produces).
- `/simplify`: removed a redundant `xfer_data` latch (replaced by the already-stable
  `mem_instr` input) and a duplicated `mem_valid && mem_instr && !already_aimed` boolean
  (factored into `redirect_now`); re-verified after with `make test` and a spot Dhrystone run
  reproducing identical numbers.

## Scope notes

- **A design bug in the timing model itself, not in `nano.v`**, was found and fixed while
  deriving it: a redirect's `mem_ready` check read a live, per-cycle recomputation of "already
  aimed here" rather than a latched decision, and a newly triggered redirect did not clear the
  flash's stale `arrived_valid`/`arrived_index` until its own preamble completed — so a
  just-abandoned stream's leftover state could satisfy a new redirect before its real preamble
  ever ran, deadlocking Dhrystone at cycle 178,749. Full writeup and the fix in the ADR.
- Two named assumptions about a front end that doesn't exist yet (peeking the low parcel's
  length bit early; a loop-buffer hit's background resync running independently of the main
  FIFO) — both stated in the ADR, as the ticket asked.
- The comparison against littlecpu's own figures belongs in this PR description, not the repo,
  per CLAUDE.md and the brief: **nano's best measured configuration here (2.26 DMIPS, 7.96
  CoreMark at an assumed 64 MHz) remains far below littlecpu's up5k figures (9.3 DMIPS-class)**
  — the loop buffer helps, substantially, but doesn't change nano's fundamental character as a
  QSPI-bandwidth-bound core.
- DECISION NEEDED (flagged for the architect, not blocking): the ADR recommends the 8-parcel
  loop buffer (~5.2k um² TT-flow) as the configuration most worth building if area is still
  contested, since it captures roughly three-quarters of the measured gain at half the area of
  the 16-parcel version — but which to actually build is a design call for the front-end
  ticket, not resolved here.


thejefflarson:
## Architect review: HOLD

The model's no-overlap and FIFO fetch timing and its load cost check out, both by hand and against TinyQV's controller. The loop-buffer rows do not. Most of their gain comes from serving parcels that were never fetched, and the model gives real resident loops much less benefit than a real buffer would. The headline "the loop buffer is the lever" is therefore unsupported, and so is "the brief's ~2 MIPS is reached only with a loop buffer". Stores are also overcharged. Nothing was pushed to the branch.

### How I checked

I wrote micro programs (straight-line `.S`, rv32emc, `nano/asm/nano.lds`) and ran them against `nano-qspi-sim` builds at depth 0, FIFO 2 and FIFO 2 + loop 8. Each cost below is the cycle difference between 64 and 32 repetitions, divided by 32, so the prologue cancels. Every figure is bus cycles, where a zero-wait access is 1.

| Program | Hand-derived | Model | Verdict |
|---|---|---|---|
| compressed ALU op, depth 0 | 8 fetch + 4 execute = 12 | 12.0 | correct |
| 32-bit ALU op, depth 0 | 16 + 4 = 20 | 20.0 | correct |
| compressed ALU op, FIFO 2 | fetch-bound, 8 | 8.0 | correct |
| 32-bit ALU op, FIFO 2 | 16 | 17.0 | one-cycle full-FIFO bubble; minor |
| `c.lw` + `c.addi`, depth 0 | 8 + 2 + 45 (PSRAM) + 2 + (25 + 8 + 1) + 4 = 95 | 95.0 | correct |
| `c.sw` + `c.addi`, depth 0 | 94 (no `reg_write`) | 94.0 | correct |
| `c.lw` + `c.addi`, **no branches at all**, FIFO 2 + loop 8 | 91 (same as FIFO 2; there is no loop) | **83.0** | phantom hits |
| 2-instruction loop inside one aligned 16-byte block, FIFO 2 + loop 8 | about 12 with the body resident | **38.0** | resident body not served |
| 4-instruction, 8-byte loop in one block, FIFO 2 + loop 8 | about 24 | **54.0** (FIFO 2 alone: 62) | resident body not served |

A load's penalty over a sequential stream is 44 + 25 + 1 = 70 clocks, against the brief's ~68. A redirect's first parcel costs 25 + 8 + 1 = 34 bus cycles, which matches TinyQV's flash path at 34–36. The deadlock fix therefore does not make fetches too cheap. The cheap fetches come from `loop_hit`.

1. **`nano/tb/nano_qspi_memory.v:73-75` (High).** `loop_hit` only asks whether the target is in the same aligned block as the last completed fetch. It never asks whether the target's parcels arrived. After every load or store the stream is closed, so the next sequential instruction is a "redirect" in the load's own block, and it "hits" in 2 cycles without ever being fetched. The straight-line row above shows this: a program with no loops gets 8 cycles cheaper per load. **Size on the benchmarks:** I built a variant that allows a hit only on a *backward* target (`&& mem_addr < window_base`), which excludes the post-data-access and forward cases. Dhrystone FIFO 2 + loop 8 then reads **21,890.6 cycles/run, identical to FIFO 2 alone**, so the published +14.6% is entirely phantom. At loop 16, Dhrystone's gain falls from 23.8% to 5.8% and CoreMark's from 19.1% to 2.3%. CoreMark at loop 8 falls from 18.1% to 2.2%. These percentages are cycle reductions against the FIFO 2 row. **Fix:** give the buffer one valid bit per parcel, set when a parcel streams in under the current tag and cleared on retag. A fetch hits only if every parcel of the target is valid.
2. **`nano/tb/nano_qspi_memory.v:67-69, 86, 146-154` (High, the opposite direction).** The buffer is consulted only when `!already_aimed`. A hit aims the flash at the hit's successor (`preamble_target <= target_index + target_len`). The successor is then "already aimed", so it skips the buffer and waits out the rest of a 25-cycle preamble. The model therefore serves one instruction per redirect from the buffer, never a resident loop body. That is why the 2-instruction loop costs 38 cycles instead of about 12. **Fix:** consult the buffer on every fetch, sequential included, and restart the flash only at the first parcel that is not resident.
3. **`nano/tb/nano_qspi_memory.v:142-145, 151` (Medium, pessimistic).** A data access closes the stream, and the next fetch's redirect clears `arrived_valid`, so parcels already in the FIFO are discarded. By trace this is usually small: at 8 clocks per parcel and only 2 non-bus cycles between a load's fetch and its data access, the FIFO rarely holds anything beyond the load. After a divide or multiply it does. I did not measure it. The model also waits for the core's next fetch before starting the re-preamble, even though a controller knows the sequential address when the PSRAM releases chip select. That costs about 2 clocks per data access. **Fix:** keep arrived parcels across a data access, and start the flash re-preamble at PSRAM chip-select release, aimed at the FIFO's tail.
4. **`nano/tb/nano_qspi_memory.v:81, 145` (Medium, pessimistic): stores are charged the same as loads (44).** See the PSRAM section below: a word store is about 33 clocks. A variant charging stores 32 moves Dhrystone FIFO 2 from 21,890.6 to 20,990.3 cycles/run (−4.1%) and FIFO 2 + loop 16 by −5.4%. CoreMark moves −1.9% and −2.4%. **Fix:** give stores and loads separate parameters: 33 for a store, 44 for a load (use 43 if the TT mux latency is configured at its minimum).
5. **`docs/adr/0186-…md:50-60, 147-179`.** The loop-buffer description, the "8 parcels captures three-quarters of the gain" recommendation and the "~2 MIPS only with a loop buffer" conclusion all rest on findings 1 and 2. They have to be re-derived after the rework, not edited.
6. **`docs/adr/README.md:311` (housekeeping).** The 0186 row sits above 0185; move it below. `adr_numbering_test.sh` passes because it does not check order.

### PSRAM cost against TinyQV's controller (the biggest bucket)

I could not open the APS6404L datasheet offline. Instead I read TinyQV's `cpu/qspi_ctrl.v` and its README; TinyQV runs on this same Pmod (APS6404L-3SQR-SN) at a 64 MHz core with SCK at 32 MHz. Its README says to use the PSRAM in QPI mode with Fast Read `0Bh` ("12 cycles preamble for read, 20 cycles total for a 32-bit read", with a 66 MHz limit) and Write `02h` ("no delay cycles… 16 cycles total for a 32-bit write"). Those counts are SCK cycles. Counting its state machine in core clocks at SCK = clk/2 gives:

| Access | Select | Command (2 nibbles) | Address (6) | Dummy (4) | Data (8 nibbles) | Input latency | **Total** | Model |
|---|---|---|---|---|---|---|---|---|
| word read, `0Bh` in QPI | 1 | 4 | 12 | 8 | 16 | 1–3 (TT mux, configurable) | **42–44** | 44 |
| word write, `02h` in QPI | 1 | 4 | 12 | none | 16 | none | **33** | 44 |
| flash redirect to first parcel, `EBh` continuous | 1 | none | 12 | 4 mode + 8 dummy | 8 per parcel | 1–3 | **34–36** | 34 |

- **Loads: the model's 44 is right**, within 0–2 clocks (5% or less). It matches `0Bh` in QPI with TinyQV's latency setting at its maximum. With `EBh` and 6 wait cycles, which is probably what the brief's 44 assumed, a read would be about 46–48, 2–4 clocks above the model. Record the command in the ADR, because the number depends on it. The PSRAM must also be switched into QPI mode (`35h`) at boot, as TinyQV requires.
- **Stores: the model is 11 clocks too high per store, about 33%.** On the benchmarks that is 2–5% of total cycles (finding 4), and it is pessimistic.
- Two things the model does not charge are small: TinyQV's two-clock lockout before re-selecting the same RAM, which rarely applies because a fetch sits between data accesses, and the one-clock `stop_txn` that ends the flash stream.
- **Net: the PSRAM bucket is about right for loads and overstated for stores.** Fixing stores moves every row by a few percent and changes no conclusion.

### Graders the rework should add

The two defects above are exactly what a both-ways test catches, and the programs already exist:
- a straight-line `lw`/`addi` program must take **identical** cycles with and without the loop buffer (the current model fails this at 91 vs 83 per pair);
- a 2-instruction loop inside one aligned block must run near its execute bound (about 12 cycles per iteration) with the buffer, and at the FIFO cost without it.

### The DECISION NEEDED on buffer size (my view, recorded here rather than in an ADR)

This data cannot choose a size. Dhrystone's 8-parcel gain is 0.0% once phantom hits are removed, and the backward-only 16-parcel gains (5.8% Dhrystone, 2.3% CoreMark) still carry finding 2's under-count. The front-end ticket should decide after a model that serves resident bodies has been run at 8 and 16 parcels. It should also try a "last N parcels streamed" window next to the aligned block, since a loop that straddles an aligned boundary never hits an aligned buffer.

### Verified and fine

- There are no ticket IDs in the diff. `test/comment_density_test.py` passes (401 files, all 5% or under). The probe group sits at `test/probe_gates.sh:1170`, above `actual_labels=` at 8021, and both labels are in `test/PROBES_EXPECTED`.
- The default zero-wait path behaves identically. The generated cxxrtl differs from `main`'s only by source line numbers and one extra value copy from `(* keep *)` on `mem_valid`, so it is not byte-for-byte identical. Moving that `keep` under `` `ifdef NANO_QSPI_TIMING `` would make it so; that is a nit.
- I reproduced the FIFO 2 + loop 8 Dhrystone row exactly: 19,129.0 cycles/run. The accounting identity and its probes are sound.

thejefflarson:
## Architect re-review of 56b1855: READY after one fix, which I pushed

The rework fixes what the first review held: the phantom hits, the hit that answered a cycle late, the store charge, and the accounting identity. The 16-parcel hang, though, came from a third defect that 56b1855 still carried, and **it biased the 8-parcel rows too**. I root-caused it, fixed it, added graders that are forced red, and re-ran the sweep. The table in ADR-0186 is re-measured.

1. **`nano/tb/nano_qspi_memory.v:75-81, 177, 228` (High), now fixed. A loop-buffer hit moved the flash queue's head.** `expect_index` served as both the core's next parcel and the head of the flash's queue. A hit moved it without moving the stream, which went wrong three ways:
   - **Forward: a deadlock.** After a hit to a parcel beyond the stream, `(next_to_produce - expect_index) < PREFETCH_DEPTH`, an unsigned comparison, wrapped. The stream stopped for good, and the next fetch the buffer did not hold waited forever. This is the 16-parcel CoreMark hang. I traced both shapes to the same stuck state: the fetch aimed, `produce_room` = 0, and the stream behind the core. In the CAM the core is at parcel 1,372 while the next parcel to stream is 1,369. In the tagged block it is 912 against 907.
   - **Backward: a phantom.** After a backward hit, a later fetch the buffer did not hold matched `expect_index`, and the stream's last-arrived parcel lay beyond it. The model then served parcels the flash never streamed in its current run. An instrumented copy of 56b1855 counted 2,579 of these on Dhrystone and 10,158 on CoreMark, in the 8-parcel tagged block.
   - **A clobber.** At `:177` the preamble's completion rewrote `expect_index`, undoing hits taken during the resync. Each clobber could charge a redirect that was not owed; there were 7,697 on Dhrystone, in the 8-parcel tagged block.

   **Fix:** a separate `fifo_head`, which advances only when a delivered instruction starts at it. `produce_lead` is now signed, and `PREFETCH_DEPTH=1` is refused at elaboration, because at that depth a 32-bit instruction's second parcel could never be queued.
2. **The same model had no grader for "served from the flash".** I added `stream_fault`, computed from the run's start and the last-arrived parcel, independently of the aim logic. Every `nano-qspi-sim` exits 7 on it, benchmark runs included. `nano-qspi-loop-test` gains a third invariant: a loop with a load and a block-straddling 32-bit `addi` must run to PASS in both shapes. The probe gains two mutations:
   - the queue head following every hit, which is caught by the stream-fault exit;
   - an unsigned lead, which is caught by `TIMEOUT`.

   All four probe mutations are caught, each for its own reason.
3. **`docs/adr/README.md` (housekeeping), fixed.** The 0186 row sat after 0187, and its text still described the first draft: 400 runs, four buckets, and numbers since superseded. It is now rewritten and placed after 0185.

### What moved

No-overlap, FIFO 2 and FIFO 4 are cycle-identical before and after on both benchmarks.

| Row | Dhrystone cycles/run, 56b1855 → fixed | CoreMark cycles/iteration, 56b1855 → fixed |
|---|---|---|
| FIFO 2 + tagged 8 | 17,010.5 → 16,290.5 (−4.2%) | 8,265,043.8 → 8,208,735.2 (−0.7%) |
| FIFO 2 + CAM 8 | 16,939.5 → 16,940.5 (+0.006%) | 8,528,141.4 → 8,530,037.4 (+0.022%) |
| FIFO 2 + tagged 16 | hung | 15,387.5 / 8,117,862.2 |
| FIFO 2 + CAM 16 | hung | 14,645.5 / 8,046,371.6 |

The two 16-parcel rows give the fixed model's Dhrystone and CoreMark figures together, since 56b1855 had no figures for either.

The phantom and the clobber pull in opposite directions. The old model's claim that "neither shape dominates" does not survive: at 8 parcels the tagged block leads on both benchmarks, by 4.0% and 3.9% of cycles, and at 16 the CAM leads on both, by 5.1% and 0.9%. QPI now lands on the 16-parcel CAM: 2.57 DMIPS and 8.20 CoreMark at 64 MHz.

### Re-derived by hand

These are marginal cycles per iteration, taken between 200 and 400 iterations.

| Loop | Figure |
|---|---|
| Two-instruction loop, resident in either shape | 10 cycles, exactly its zero-wait cost |
| Straddling loop with a load, CAM 8 (resident there) | 75 = zero-wait 31 + `PSRAM_LOAD_CYCLES` 44, with 0 preamble and 0 parcel wait |
| Same loop, tagged 8 | 152.5, against 157 for FIFO 2 alone; 1.5 hits per iteration, as the block re-tags at the straddler |
| Same loop, 56b1855's tagged 8 | 104, the phantom |

### Checked and fine

- No ticket IDs anywhere in the diff.
- `test/comment_density_test.py` passes on all 413 files.
- `adr_numbering_test.sh` passes.
- ADR-0186 stays in nano's own terms, with no comparison to another core.
- The zero-wait control reproduces ADR-0182 to the cycle.
- `make nano-qspi-loop-test` passes, with all four probe mutations caught, in 4 min 41 s on `make test`'s path, up from 2 min 51 s.
- The full `make test` passes locally.

PR #359, opened 2026-09-14
nano-tt-area-selfhosted: fail fast on congestion, fix the PDK cache and summary bugs, measure a real 4x2/3x2 answer

thejefflarson:
Fixes the nano-tt-area-selfhosted instrument so a dispatch produces a real answer
(completed routing, or a named stop cause read from the uploaded reports) instead of
either a wasted multi-hour detailed-routing run or a silently-lost failure with no
evidence. Also folds in the two defects already filed against the same file.

## Instrument changes

- **Fail fast**: new `disallow_congestion` dispatch input sets `GRT_ALLOW_CONGESTION`
  to 0, so an overflowing global route stops at Global Routing with its own overflow
  report instead of being carried into detailed routing for ~1h45m before the job
  timeout cancels it (what the 4x2 run cited in the ticket did). Tiny Tapeout's own
  template default (allow it) is unchanged when the input is left at its default.
- **Keep the evidence on any outcome**: the run-directory collection step now walks
  every step's `reports/` and `*.log` files (previously only `synthesis-stats.txt`),
  so a failed or cancelled run still uploads the global-routing congestion report and
  whatever detailed-routing reports exist.
- **PDK cache split** (Closes JEF-1033): the combined `actions/cache` step only saved
  on job success, and this instrument's runs usually fail, so the PDK was re-downloaded
  every run (one run died on a DNS failure mid-fetch). Split into
  `actions/cache/restore` + `actions/cache/save`, the save under `if: always()` right
  after the harden step that resolves the PDK, both SHA-pinned to the same commit the
  combined action was already pinned to.
- **Summary mode fixed** (Closes JEF-1032): the job summary echoed
  `${{ inputs.stop_after_synthesis || 'true' }}`, and `||` cannot tell the string
  `"false"` from absent, so it always printed `true`. Resolved once into
  `$NANO_TT_STOP_AFTER_SYNTHESIS` and read from there.
- **Job summary mirrored to stdout**: a step summary is invisible to
  `gh run view --log`, and the fail-fast dispatches confirmed `/usr/bin/time` is not on
  this runner image, so the summary's cgroup memory figure was the only wall-time/memory
  data any run produced and there was no way to read it back short of the Actions UI.
  Each summary line is now echoed to stdout too.
- `timeout-minutes` raised 120 → 150 for the case that does proceed past Global Routing.

## Config change

`PL_TARGET_DENSITY_PCT` raised from the template's 60 to 70 in `src/config.json`, the
value `MichaelBell/tt10-tinyQV` — an accepted multi-tile RISC-V Tiny Tapeout
submission — ships in its own `src/config.json`, and the one setting LibreLane's own
template comment names as the remedy for "Global Placement fails with error GPL-0302"
(this design's congestion is the same class of problem, global placement leaving
routing demand above what the floorplan's tracks can carry). **Measured effect: none
on the outcome.** GPL's own routability optimiser already inflates cells well past our
target when trying to cut congestion — its log shows it hit the internal ceiling
(target density 1.0243 against a hard max of 0.99) and reverted to the best iteration
it found — so a placement-density knob was never the lever; deferring to that loop's
own record of what it already tried is worth keeping as the citation for why raising it
further would not help either.

`PNR_SDC_FILE`/`SIGNOFF_SDC_FILE`: Tiny Tapeout's own template (`ttsky-verilog-template`,
current `main`) ships neither, so the generic OpenROAD fallback SDC this instrument has
always used (confirmed in both runs' logs: `'PNR_SDC_FILE' is not defined. Using generic
fallback SDC for OpenROAD PnR steps.`) is upstream's own behaviour, not a workflow gap.
No change made.

## Results

One dispatch each, `AREA 0`, `disallow_congestion=true` (the reports below are also
consistent across two further 4x2 dispatches taken while debugging the memory/wall-time
capture — the congestion numbers are byte-identical run to run, confirming the flow is
deterministic here):

| tile | utilization | GRT congestion | DRT violations | setup / hold | wall time | peak memory | run |
|---|---|---|---|---|---|---|---|
| 4x2 | 63.638% (core 677.12 × 220.32 µm) | **stopped at Global Routing**: `[GRT-0116] Global routing finished with congestion.` Final report: met1 108.26%, met2 115.92% (worst layer), met3 104.93%, met4 79.09%, **total 105.77% usage**, max H/V/total overflow 27/30/**15718**. GPL's own pre-route overflow ranged 14.19–18.65% across its 7 routability iterations, reverting to iteration 6's density (weighted congestion 1.2466) after 1.0243 exceeded the 0.99 ceiling | not reached (GRT never completes) | setup: 360 endpoints violating → 76 remain after repair, final WNS **−1.442 ns**, TNS **−49.3 ns**; hold: fully repaired, final WNS 0.100 ns, TNS 0.000 (0 violations) after 827 buffers | 958 s (~16.0 min; harden step alone 10.0 min) | 3,221,188,608 B (~3.22 GB) | [34811085686](https://github.com/thejefflarson/little-cpu/actions/runs/34811085686) |
| 3x2 | 85.723% (core 503.24 × 220.32 µm) | not reached | not reached | setup: 249 endpoints violating → 31 remain after repair, final WNS **−1.245 ns**, TNS **−26.4 ns**; hold: 727 endpoints found violating, repair triggered the fatal placement failure before finishing | 646 s (~10.8 min) | 3,221,192,704 B (~3.22 GB) | [34810266409](https://github.com/thejefflarson/little-cpu/actions/runs/34810266409) |

**3x2's named cause is placement, not routing**: `[ERROR DPL-0036] Detailed placement
failed` in the post-CTS resizer (`rsz_timing_postcts.tcl:61`), legalization failing on
45 instances while repairing hold violations — this happens regardless of
`GRT_ALLOW_CONGESTION`, since it is well before Global Routing runs at all. At 85.7%
utilization against a fixed die (`FP_SIZING: absolute`, sized by tile count), no
placement-density setting changes that the design does not fit.

**4x2 is the more interesting case: everything through CTS and resizing succeeds**, and
the flow reaches Global Routing on a legally-placed, timed netlist — it fails purely on
routing demand exceeding the fixed die's tracks (105.77% average, worse on met1–met3),
consistent with the 1,082-terminal `clk` net GPL's own log already flags. Two earlier
allow-congestion runs (the ticket's own citation, and this branch's 4x2 run before this
fix landed) show what "allowed through" costs: 155,430 Shorts by the third
optimisation iteration, cancelled by the job's own timeout rather than converging — so
raising `timeout-minutes` further would not have produced a routed 4x2 either;
`disallow_congestion` reaches the same conclusion in about 16 minutes instead of a
timeout at 120.

**Neither tile size has produced a routed 4x2/3x2 layout under `AREA 0`.** Both named
causes are read from the uploaded artifacts linked above (`error.log`,
`39-openroad-globalrouting/openroad-globalrouting.log`, and
`37-openroad-resizertimingpostcts/openroad-resizertimingpostcts.log`).

## DECISION NEEDED

Neither the 3x2 nor the 4x2 tile routes at `AREA 0` with the one config change this
repo's own Tiny Tapeout precedent supports (`PL_TARGET_DENSITY_PCT`). 4x2's failure is
specifically a routing-resource one (105.77% average demand, 1,082-terminal clock net),
not an area or placement-density one, so the candidates worth measuring next are RTL-side
(clock-tree/fanout structure, register count) or a wider tile (`6x2`/`8x2`, outside this
ticket's config-only scope) — not another LibreLane knob in `src/config.json`. This
instrument now answers the question cheaply (about 10–16 minutes per dispatch) so that
next measurement is inexpensive to take.

## Tests

- `make test` (full suite, including `make probe-gates` and the 5%-per-file comment
  density check): green.
- New `test/nano_tt_area_workflow_test.py`, wired into `make nano-tt-area-workflow-test`
  and `make test`, grades two shapes of the workflow: the summary step must read
  `$NANO_TT_STOP_AFTER_SYNTHESIS` rather than re-derive the mode from `inputs.`
  directly, and the PDK cache must be a `restore` + `save` (`if: always()`) pair, not a
  combined step. Forced red directions for both, plus a missing-workflow-file case, are
  wired into `test/probe_gates.sh` with labels in `test/PROBES_EXPECTED`.
- Four real dispatches on this branch (linked above, plus two intermediate 4x2 runs
  taken while fixing the memory/wall-time capture) are the acceptance evidence itself.

Closes JEF-1034
Closes JEF-1032
Closes JEF-1033


thejefflarson:
Reviewed. The security pass is clean (no Critical or High findings). I reverted `PL_TARGET_DENSITY_PCT` to the template's 60 in 1fcf728, because 70 measured as having no effect. That keeps the 64 MHz clock as the only change from Tiny Tapeout's template, and it matches the earlier runs on main.

One reading correction: the runs' `memory.peak` values (3,221,188,608 and 3,221,192,704 bytes) sit within about 37 KB of the pods' 3 GiB limit (3,221,225,472). They include page cache, so they show the pod was at its ceiling, not how much the tools actually needed. The logs show no OOM kill.

PR #363, opened 2026-09-14
Spike: measure fetch-ahead-with-discard on up5k and ECP5, decline on both

thejefflarson:
Closes JEF-1039.

**Updated after review** (see commit `de3698f`): the review correctly found that K and
K+BTFN's own fetch address still depended on decode combinationally — confirmed by
reading the synthesized ECP5 critical path directly, which runs through decode's `out`
register, `next_pc`'s own adder and `fetch_pc_next` before reaching the ROM address.
K/BTFN's verdict (declined, both parts) is unaffected — cycles alone already declined
them on up5k, and ECP5's product was negative before this was found — but the ADR is
rewritten so it doesn't read as covering a genuinely register-only fetch address, which
was never measured in the first cut.

## What this is

A spike, not a shipped change: **`rtl/` is untouched.** It rebuilds ADR-0078's
registered-fetch-address "K" (predict-not-taken, one-cycle discard) on the tree
ADR-0154 and ADR-0158 have since moved, re-measures it at twelve seeds per part
instead of ADR-0078's eight up5k-only, adds a static backward-taken predictor
(K+BTFN), and takes the ECP5 reading ADR-0087 explicitly never took. After review,
it also attempts a **third variant**, decoupled fetch: a genuinely register-only
fetch address behind a new word-granular queue (`rtl/fetchqueue.v`).

Full writeup: `docs/adr/0188-fetch-ahead-with-discard-is-measured-and-declined-on-both-parts.md`.

## Verdict

**K and K+BTFN: declined on both parts**, same as before. On up5k the clock is a step
function and neither reaches 24 MHz, so the comparison is cycles alone: K costs
2.87–10.81% more cycles depending on workload, K+BTFN roughly halves that. On ECP5 the
clock gain is small (a near-total wash at the worst of twelve placements) and the
product is negative on every workload.

**Decoupled fetch: inconclusive, not declined.** Two real synchronization bugs were
found and fixed in the queue controller (documented in the ADR); a third remains
unresolved. The suite does not pass, so there is no cycle count, no product, no
comparison against K/BTFN or base — this is explicitly *not* a result. What is
measured: it elaborates and synthesizes cleanly on both parts; up5k does not fit at
this queue depth (5861/5280 LC, 111%); ECP5 places at four seeds (not twelve — time,
not policy) reading 37.30–40.99 MHz, higher than base/K/K+BTFN's own ECP5 numbers
(33.01–35.44 MHz), and its critical path now runs **entirely inside
`rtl/fetchqueue.v`'s own pop/occupancy logic**, confirmed by direct inspection of the
synthesized path at all four seeds — the loop that used to close through the ROM's
address decode now closes through the queue's own bookkeeping instead. This is a
structural signal that the mechanism is what review asked for, not a verdict on
whether it's worth shipping.

## How it's reproducible

`soc/fetch_ahead/prototype.patch` (K/BTFN) and `soc/fetch_ahead/prototype-decoupled.patch`
(decoupled fetch) are tracked unified diffs against `rtl/`, applied by
`soc/fetch_ahead/apply.sh`/`apply-decoupled.sh` onto a copy of the tree — patches
rather than an anchor-based generator like `soc/depth/variants.py`, because both reach
too far into `next_pc`/`pc` timing for a handful of checked anchors to carry safely.
`soc/fetch_ahead/sweep.sh {up5k|ecp5} [seeds...]` places variants and reuses
`soc/depth/row.py`'s report readers; `soc/fetch_ahead/cycles.py` patches
`test/cxxrtl.cc` at checked anchors to add `kill`/`redirect` buckets. Nothing here is
on `make test`'s path, matching `soc/depth/`'s own standing.

## What was found and fixed along the way

**K+BTFN**: the first cut of the mispredict signal was "did `next_pc` take a named
redirect arm," correct under plain K and wrong under BTFN, where a *correctly* guessed
backward branch can make the guess equal the real target — the old definition
discarded that correctly-fetched word anyway. Fixed by comparing the guess against the
resolved target directly, spelled so K's own netlist is untouched bit-for-bit.

**Decoupled fetch**: `imemory` answers whatever address `fetch_pc` presents *every*
cycle with no request/response handshake, so any cycle `fetch_pc` is held (a flush's
one settling cycle, a retry) generates a second, genuine response for an address
already accounted for. Two instances fixed (a stale pre-flush response pushed as the
redirect target's own data; a held-for-two-cycles address requested from memory twice
with both answers separately tracked); a third, related instance remains open — see
the ADR's own section for the exact symptom and where the next session should look.

## Testing

- K/BTFN: `.S`/`.c` suite 75/75 PASS, both variants; Sail cosim 69/75 AGREE, matching
  `test/COSIM_EXPECTED_FAIL` exactly.
- Decoupled fetch: elaborates and synthesizes cleanly on both parts; suite does
  **not** pass (1/75) — functionally unverified, reported as such throughout.
- `test/decoder_tb.v` was **not** updated for K/BTFN and fails as expected (24
  mismatches, all `next_pc`/`pc` timing assertions this patch changes on purpose) —
  the same shape ADR-0078 found and fixed with a small, targeted bench edit; not
  spent here since the suite and Sail cosim are this ticket's correctness bar.
- Formal was **not run** for either variant — the ADR gives a "first read" of what
  would need to change, not a re-closed proof.
- `make test` on the actual checkout: green, `rtl/` byte-identical to `main`.
- `/soundcheck:pr-review`: no Critical/High findings.

## Trade-offs / what I didn't do

- Decoupled fetch's third bug was not found within this session's time budget — the
  ADR names the exact symptom (a wrong word at a flush target reached shortly after
  an earlier flush) and where to look next.
- Only four ECP5 seeds for decoupled fetch, not twelve — meaningful as a structural
  signal, not as the go/no-go reading this repo's own methodology requires.
- up5k was not placed for decoupled fetch at all (doesn't fit); a shallower queue is
  the first thing to try if this is picked back up.
- Formal is a first read, not a proof, for K/BTFN; not attempted at all for decoupled
  fetch, which does not yet have a correct simulation to build a proof against.


thejefflarson:
**Architect review of #363. READY once CI is green on dbb9ca4.** No merge and no auto-merge arming were done here.

`rtl/`, `formal/` and `test/` are untouched, and 0188 is free on main (0184, 0186 and 0187 are there). The index merges cleanly and adr-numbering, comment-density, march, tracked-ignored and retired-term all pass. Both patches apply at fuzz 0 to this branch and to main. The up5k count for the 8-word queue reproduces exactly (5861 LC).

Fixed in dbb9ca4, in severity order:
1. **ADR-0188, Decision and Verdict: a median number was quoted as the worst-placement verdict.** "−0.25% (K+BTFN, the suite)" is the median-placement product; at the worst placement it is **−2.73%**. The sentence claiming this was "closer than anything ADR-0078 or ADR-0087 measured" was false, because ADR-0087 recorded a ×1.004 cell. Also corrected: K's cycle range is 3.33–10.81%, not 2.87 (2.87 is BTFN's), and the up5k product is the inverse of the cycle ratio (−2.79% to −9.75%), not the negated cycle cost.
2. **ADR-0188, decoupled section: the claims ran ahead of the evidence.** "The bug is in the queue's data path" was unsupported. "33.01–35.44 MHz across their twelve seeds" was wrong: base reaches 36.41 MHz and btfn 36.51. "Entirely inside `riscv.fetchqueue`" contradicted the path the ADR itself lists, which crosses the register file, `next_pc` and `fetch_stall`. Every decoupled number now says it comes from four seeds of a design that failed 74 of 75 programs. The comparison is against base's own seeds 0–3 (33.87, 34.22, 33.01 and 36.41 MHz).
3. **Title, file name and index row read as "fetch-ahead declined" in general.** They are renamed to scope the decline to decode-coupled K and K+BTFN.
4. **F/G paragraph misstated ADR-0078.** ADR-0078 found F and G unchanged, not lengthened. The K median period move (−2.98%) is inside the 3.6% edit-churn band, so it is a null; BTFN's (−4.03%) clears it. BTFN's `fit` count (4283) is over `FIT_MAX_LC` (4219).
5. **Reproducibility.** `apply.sh` used fuzzy `patch`, not the `git apply` the ADR described, and ran `rm -rf` on any path it was given. It now applies at `--fuzz=0` and refuses to replace a directory it did not make. `prototype.patch` lacked the `uart.S` floor, so the staged K tree read 74/75; it now carries the floor, and K reads 75/75 as staged. `sweep.sh` restated the SoC source list, the defect ADR-0087's amendment fixed, and had no decoupled variant; it now reads each tree's Makefile and places `decoupled` on request. `cycles.py` defaulted CoreMark to 400 iterations against the table's 100, and could not simulate BTFN; it now defaults to 100 and has `--btfn`.

Found at review and recorded in the ADR, not carried in the patch: the decoupled controller has two more defects.
- A flush does not cancel the response already in flight, which produces the back-to-back symptom.
- `committed` ignores a request issued this cycle, so the 8-word ring overruns its head.

In a scratch tree the suite reads 1/75 with neither fixed, 1/75 with the first alone, 50/75 with the second alone and 73/75 with both. The two remaining failures are `selfmod.S`, because nothing flushes the queue on `fence.i`, and `uart.S`'s floor. A 4-word queue packs 5493/5280 LC and does not fit the up5k either.

Not re-run here: Sail co-sim (69/75), the BTFN suite, and the decoupled ECP5 placements, which were not kept.

PR #364, opened 2026-09-15
nano: a latch-array register file for a small Tiny Tapeout tile, selected by NANO_LATCH_RF

thejefflarson:
Closes JEF-1035

## Question

Does a latch-array register file bring nanocpu's Tiny Tapeout area/routing down to a
closing tile? ADR-0184 ranked the register file (and mul/div) as the two largest
contributors to today's flip-flop core failing to route on a small tile, and set the
work order this ticket's step follows: latch register file, then shared-register
mul/div, then a one-read-port register file.

## What this PR ships

- `nano/nano.v` gains a `NANO_LATCH_RF`-selected register file: sixteen
  **transparent-LOW** latches, gated by a **registered** enable/data pair
  (`we_q`/`wdata_q`, captured at the same `posedge clk` the FSM already runs on), not a
  same-edge combinational decode. A first cut gated on `clk && sel` decoded straight
  from `cpu_state`/`rd`/`reg_wdata` — a real race, since those are flops changing at
  the exact edge that opens the latch, with no designed margin and no clean STA story
  for an ANDed clock. The shipped scheme instead opens each latch on
  `!clk && we_q[gi]`: `we_q`/`wdata_q` are flop outputs, stable a full period before the
  falling edge that reads them, so the window is never raced by the signals that gate
  it — the ordinary registered-enable construction OpenSTA's own latch/time-borrowing
  checks are built to verify, with no dedicated clock-gate cell needed since the
  enable never leaves the data path into a clock tree.
- Kept inline behind an `ifdef` rather than split into a second file: nano's
  mutation/probe scripts and every `nano/formal/*.sby` file assume `nano.v` is the one
  RTL source they read, and a second file buys nothing the `ifdef` doesn't already
  give.
- **Read-after-write checked, not assumed**: a retiring write's own latch is still
  closed at the moment of retire, but nothing in nano's real execution reads a
  register's *value* until `execute_instr`, several cycles later either way. RVFI's own
  same-edge `rd_wdata` capture (verification-only) now reads `reg_wdata` directly under
  `NANO_LATCH_RF` rather than the not-yet-reopened array, ifdef-gated so the default
  build's text is untouched — `nano-latch-test`/`nano-latch-startup-test` are the check
  this has to pass, since a stale `rd_wdata` is exactly what makes the monitor's
  independent reference disagree with the core's own report. Both variants retire
  identically on both simulator legs.
- **Default build proven unchanged**: every added line sits behind an `ifndef`/`ifdef`,
  so `iverilog -E` on `nano/nano.v` with no macros reproduces origin/main's file
  verbatim (modulo blank-line padding), and `synth;dfflibmap;abc` against sky130hd
  lands on the identical 61,411.3984 µm² before and after.
- `nano-latch-test`/`nano-latch-startup-test` join `make test`, at about nano-test's own
  cost (~20s locally), introducing no new comparison logic.
- `nano-tt-area-selfhosted.yml` gains a `regfile` choice input (`flops` default,
  `latches`), injecting `VERILOG_DEFINES` into the merged LibreLane config the same way
  `synth_strategy`/`disallow_congestion` already do.
- **Formal statement, now a real decision gate**: elaborating `nano/nano.v` under
  `NANO_LATCH_RF` through `ill_e.sby`'s and `dmemcheck.sby`'s own scripts produces
  fifteen clean `$dlatch` cells through `check`/`flatten`, but both backends nano's
  checks depend on refuse them — `write_btor`/`write_smt2` each error `Unsupported cell
  type $dlatch ... please run clk2fflogic before write_btor/write_smt2`. `NANO_LATCH_RF`
  cannot become the default until `make -C nano/formal check` reads it: `clk2fflogic`
  added to every `.sby` script and `checks.cfg`, and F/G re-derived.
- ADR-0189 records the design, the timing argument for the registered scheme, the
  equivalence evidence, and local structural findings, citing ADR-0184's work order.

## Measurement — DECISION NEEDED, routed numbers not yet obtained

Local (non-TT-flow) cell census against the flip-flop build, from a
`synth;dfflibmap;techmap<dlxtp>;abc` run (an ad hoc stand-in for the real flow's
`SYNTH_LATCH_MAP`, since plain `dfflibmap` leaves `$_DLATCH_P_` unmapped):

| cell | flops | latches | Δ |
|---|---|---|---|
| `sky130_fd_sc_hd__mux4_2` (write-side hold/load select) | 268 | 0 | −268 |
| `sky130_fd_sc_hd__edfxtp_1` (gated flop) | 792 | 280 | −512 |
| `sky130_fd_sc_hd__dfxtp_1` (plain flop; new `we_q`/`wdata_q`) | 185 | 224 | +39 |
| `sky130_fd_sc_hd__dlxtp_1` (new) | 0 | 480 | +480 |
| local area (µm²) | 61,411.40 | 50,870.04 | −17.17% |

The write-side mux ADR-0184 traced 408 of 941 hold buffers through has no equivalent
structure in the latch build. **The read-address fan-out ADR-0184 found dominant is
untouched by this change**: every `regs[rs1[3:0]]`/`regs[rs2[3:0]]` read site is
byte-identical text under both builds, so that share of routing demand is expected to
carry over largely unchanged.

**The routed 4×2/2×2 numbers ADR-0184's work order needs before mul/div's turn are not
measured.** The self-hosted runner pool (`little-cpu-runners`) has had nothing
`in_progress` for most of this PR's working session across the whole repo, not just
this branch (`gh run list --status in_progress`/`--status queued`), including several
other PRs' CI jobs queued for 20+ minutes at a time. Two dispatches on this branch were
cancelled: the first after sitting over two hours with the pool not moving, the second
deliberately, because it started against the pre-fix racy RTL before the registered-
enable scheme landed. Follow-up, once the pool is moving:

```
gh workflow run nano-tt-area-selfhosted.yml --ref thejefflarson/jef-1035-measure-a-latch-array-register-file-for-nanocpu-in-tiny \
  -f tiles=4x2 -f stop_after_synthesis=false -f disallow_congestion=true -f "synth_strategy=AREA 0" -f regfile=latches
```
then `tiles=2x2`, per ADR-0184's own work order.

## Testing

- `make nano-test`, `make nano-latch-test`, `make nano-startup-test`,
  `make nano-latch-startup-test` — all green, both simulator legs, identical retire
  counts across variants, under `RISCV_FORMAL` (which is what would have caught the
  RVFI `rd_wdata` staleness this PR's review caught before that).
- `make test` — full suite green (75/75 `.S`/`.c` programs, all repo-scanning
  `*-test` targets, `probe-gates`).
- Equivalence: `iverilog -E` diff (default define) and a `synth;dfflibmap;abc` area
  match (61,411.3984 µm² both) against origin/main's `nano/nano.v`.
- Formal: hand-run `ill_e.sby`/`dmemcheck.sby` scripts under `NANO_LATCH_RF` against
  the corrected RTL, both backends error identically on `$dlatch`, confirming the
  formal statement above still holds.


thejefflarson:
Runner pool came back up after this PR was opened. Dispatched the latch 4×2 measurement: https://github.com/thejefflarson/little-cpu/actions/runs/34923698630 (queued as of this comment). Follow with `tiles=2x2` then `tiles=3x4` per ADR-0188's DECISION NEEDED section once this one finishes.

thejefflarson:
Review: the code, tests and equivalence evidence are good, and the area and routing measurement stands. **Don't merge this as the latch register file yet, though.** The write scheme isn't safe in silicon.

**The hazard.** Each latch is transparent while `clk && sel` (`always_latch if (clk && sel) regs[gi] = din`). `sel` decodes `cpu_state` and `rd`, and `din` is `reg_wdata`. All of those come from flops on the same rising edge the latch opens on. So at every rising edge, the latch is transparent while its enable and data are still switching:
- **The write cycle's own data isn't safe.** At the edge that ends it, `sel` stays high for a clock-to-Q plus decode delay. Meanwhile `reg_wdata` can already be moving to the next cycle's value, and the latch can capture it. That's a race decided by gate delays.
- **Other registers can be corrupted.** When `cpu_state` and `rd` change bit by bit at an edge, the `== reg_write && rd == gi` decode can glitch high for a register that isn't being written, while `clk` is high. That writes whatever `din` is into it.
- **The flow can't stop it.** `clk && sel` is an ungated AND on the clock. STA can't hold a combinational clock-enable free of glitches.

Zero-delay simulation, cxxrtl or iverilog, can't show any of this, which is why every test here passes.

**The fix.** Open the latch only while the clock is **low**, with an enable and data that settle during the high phase and don't move until the next rising edge closes the latch. Two ways to get there:
- Register the write enable (decoded `sel`) and `reg_wdata` into flops at the rising edge, then let latch `gi` open on `!clk && we_q[gi]`.
- Drive each row from a real latch-based clock-gate cell (`sky130_fd_sc_hd__dlclkp`) with a registered enable.

Either way it's a few dozen flops, small against the 480 latches, so this run's area and routing numbers stay indicative. The ADR should give the timing argument. The flow's hold and clock-gating checks on the latch enables should be read and quoted once it's routed.

**Also:**
- **ADR-0188 cites the wrong ADR.** It names 'ADR-0179's fallback order'. The work order now lives in ADR-0184 (#350): latch register file, then shared-register mul/div, then one read port.
- **The formal gap is a cost of adopting latches.** It isn't optional: the latch build can't ship as the default until nano's formal checks read it (`clk2fflogic` plus a re-derivation of F and G). The ADR should say so in its decision, not only in its findings.

The 4×2 run (34923698630) is still worth reading when it finishes. The routing result doesn't depend on the write scheme's timing.

thejefflarson:
Fixed: the latch write scheme now registers the per-bit write-enable and reg_wdata at posedge clk, then opens each latch on `!clk && we_q[gi]` -- both a full period settled before the window that reads them opens, rather than the original `clk && sel\` racing the same edge that computed `sel`. RVFI's own same-edge rd_wdata capture now reads reg_wdata directly under NANO_LATCH_RF (ifdef-gated, default build untouched) since a retiring write's own latch is still closed at that instant. nano-latch-test/nano-latch-startup-test and the full make test suite are green on the corrected RTL; the formal $dlatch gap is unchanged and is now a stated gate in the ADR's Decision, not just a note. ADR-0188 also now cites ADR-0184's work order (latch RF -> shared-register mul/div -> one read port) in place of ADR-0179's superseded fallback. Pushed as 126937b.

Still owed: the routed 4x2/2x2 numbers -- the self-hosted runner pool has had nothing in_progress repo-wide for most of this session; two dispatches were cancelled (one for a two-hour pool stall, one deliberately for predating this fix) and a clean dispatch is queued as the immediate follow-up once the pool moves.

thejefflarson:
4×2 latches measured on the corrected (registered-enable) RTL: run [34925911447](https://github.com/thejefflarson/little-cpu/actions/runs/34925911447), LibreLane 3.0.14, `AREA 0`, `disallow_congestion=true`.

| | flops (baseline) | latches |
|---|---|---|
| Chip area | 81,879.78 µm² | **70,070.95 µm²** (−14.4%) |
| Placement utilization | 63.638% | **54.980%** |
| GRT total demand | 101.05% | **70.55%** |
| GRT per-layer | — | met1 80.77%, met2 81.43%, met3 62.09%, met4 36.37% |
| Wirelength | 874,050 µm | **653,002 µm** |
| Hold buffers (post-CTS) | 941 | **395** |
| Result | GRT-0116 congestion | GRT-0116 congestion (closer, not closed) |
| Wall time | — | 3107s (~51.8 min) |
| Peak memory | — | 3,221,188,608 B (~3.00 GiB, the pod's own ceiling) |

Real and substantial, but not enough alone: every layer still carries nonzero overflow (317 total) despite aggregate usage dropping under 100% — the localized-hotspot congestion the untouched read-address fan-out (ADR-0184's other finding) predicts.

On the hold/clock-gating ask specifically: OpenSTA's mid-PnR timing (`nom_tt_025C_1v80`, post-CTS, before the routing that never finished) names each latch by its real mechanism — `Endpoint: _nnnnn_ (positive level-sensitive latch clocked by clk')` — and analyzes it cleanly: hold WNS 0 / TNS 0.0 after those 395 buffers (`repair_timing -hold` found 885 violating endpoints first), setup WNS −0.274 ns / TNS −2.042 ns at that same pre-final-routing snapshot (the flow's later setup-repair steps never ran, since global routing failed first).

A 2×2 dispatch ([34929665602](https://github.com/thejefflarson/little-cpu/actions/runs/34929665602)) is in flight — the pair ADR-0184's own work order asks for before mul/div's turn (step 2) begins.

Pushed as 27b269e.

PR #386, opened 2026-09-19
Stage A2: the kill bubble is real and charged to its own column

thejefflarson:
## Summary

Stacks on Stage A1 (#383). A1's redirect discard mechanism (`rtl/fetchctrl.v`'s two-cycle `flush`) was already correct and formally proven; what this ticket adds is a **name and an accounting column** for the cycles it costs, which A1 charged to the generic `buffer_empty` stall reason indistinguishable from a cold start or a steal.

- `rtl/fetchctrl.v` gains `redirect_recovering`, cleared off `buffer_empty` itself (not `q_valid` directly — `q_valid` can still read true for one cycle after a redirect fires, before the flush lands on the next edge).
- `rtl/decoder.v` computes `kill = buffer_empty && redirect_recovering` locally, which makes `kill ⇒ !issuing` provable from decoder's own structure with no assumption, in every harness. `formal/decoder-kill-probe.py` is the forced-red prerequisite (mirrors `decoder-zkt-probe.py`'s shape).
- `formal/pcloop.sv`'s Property 3 ("a killed word never issues") is now load-bearing rather than trivially true.
- No control signal changed — `kill` only reclassifies cycles for `test/cxxrtl.cc` / `test/stall_report.py`, whose identity becomes `issue + kill + the eight reasons = cycles`.
- `soc/fetch_ahead/` is retired (its patches target pre-A1 `rtl/decoder.v` lines; the real `kill` signal replaces what it existed to reproduce).

**A real bug found and fixed during verification**, not by a red check: the first draft's `redirect_recovering` clear condition (off `q_valid`) undercounted the discard window by ~2 cycles per redirect. Measured on Dhrystone: **1.16 cycles/redirect before the fix, 3.00 after** — matching a hand-derived trace and the ticket's own expected ballpark. Caught by re-deriving the number the ticket asks for (this is the figure Stage A3's predictor is measured against), not by any formal or unit-test failure — nothing checks the exact duration.

**Also found and fixed, unrelated to `kill` itself** (pre-existing gaps in A1's own branch, blocking `make test` entirely):
- `test/MUTATION_COVERAGE` had no ruling for `rtl/fetchctrl.v` — ruled `unpaired components_pcloop`.
- Two `test/probe_gates.sh` fixtures (`check-memcheck-depth.py`, `formal/remeasure-fg.py`) were still built against F=6/G=6, the value before A1's own remeasure to 8/8.
- `fencei-wait-and-store-port`'s `asm selfmod.S FAIL 2` pairing no longer fires — measured on a clean checkout of A1's own tip (11a5ec7) with the mutation applied: `PASS retires=47`. A1's queue flush is a third, independent safeguard this pairing predates; `bench decoder_tb`/`bench imem_tb` still catch the mutation. Removed per `docs/manifests/mutation-detectors.md`'s own instruction to measure, not assert. `text-port-drops-load`'s full five-detector pairing is unaffected.

Full details, all numbers, and the mechanism trace are in `docs/adr/0198-stage-a2-the-kill-bubble-is-real-and-charged-to-its-own-column.md`.

## Gates (all run to completion)

| Gate | Result |
|---|---|
| `make -C formal remeasure-fg` | F = 8, G = 8, both reproduce exactly |
| `make -C formal all` | every target PASS — 86/86 generated checks, dmemcheck/imemcheck(+cover), all 6 component proofs including `components_pcloop` (run after the fix) |
| `make test` (incl. `probe-gates`) | **green through the entire run** — fails only at `comment-density-test`, a pre-existing inherited failure from A1 on 4 files this PR never touches (`Makefile`, `formal/imemcheck.sv`, `rtl/fetcher.v`, `rtl/fetchqueue.v`), confirmed via a clean checkout of A1's own tip before any change here |
| `make lint` | clean, both passes |
| `make elaborate-strict` | clean |
| `make mutation-check` | 11 mutations, 25 pairings, all caught by exactly their detectors |
| `make cosim-suite` | matches `test/COSIM_EXPECTED_FAIL` exactly |
| `make dual-smoke` | passes (32/16 retires, two harts vs one) |
| `make fit` | 4697 `ICESTORM_LC` (+17 vs A1's 4680, inside churn band) |
| `make dhrystone` | 1001 cycles/Dhrystone, 0.568 DMIPS/MHz — identical to A1's figure to the cycle |
| `make soc-timing` | still fails to place, for A1's own named reason — **expected, not owned by this PR** |
| `make ecp5-timing` | **places, 43.33 MHz** — up5k's shortfall is `ICESTORM_LC`-specific; ECP5's block RAM headroom is unaffected. First ECP5 figure taken on the Stage A line |

## DECISION NEEDED (flagged for the architect)

`formal/traps.sv` and `test/decoder_tb.v` both needed a minimal, additive touch (new `redirect_recovering` input, `kill` wire) because `rtl/decoder.v` gained a required port every instantiation must supply — this is unavoidable given the new port, but the ticket asked me to stay out of `formal/traps.sv` since Stage A4 (#) is rewiring it concurrently on its own branch off A1. My changes are small and purely additive (see diff), but will need reconciling at A4's merge.

Closes JEF-1065


thejefflarson:
Folded into #383 via merge commit (Stage A1 + A2 + A4 integration). Closing; the branch stays (Stage A3 is built on top of it).

